PR #44, opened 2018-10-20
TLS/SSL implementations for all of our currently supported platforms. Added TLS to channel bootstrap.

JonathanHenson:


source/channel_bootstrap.c
```diff
@@ -108,6 +185,116 @@ struct client_connection_args {
...
+        &connection_args->channel_data.tls_options,
+        tls_slot);
+
+    if (!tls_handler) {
+        aws_mem_release(connection_args->bootstrap->allocator, (void *)tls_slot);
+        return AWS_OP_ERR;
+    }
+
+    aws_channel_slot_insert_end(channel, tls_slot);
+    if (aws_channel_slot_set_handler(tls_slot, tls_handler)) {
```
justinboswell: This inconsistency in freeing things makes me uneasy, because ownership is managed *somewhere else*, except when it's not for some errors. Any reason not to clean up here if possible? Otherwise, this function isn't particularly atomic, which feels like a good source for state bugs or leaks later.

justinboswell: Also, this would be clearer if it were != AWS_OP_SUCCESS

source/channel_bootstrap.c
```diff
@@ -108,6 +185,116 @@ struct client_connection_args {
...
+            return AWS_OP_ERR;
+        }
+
+        aws_channel_slot_insert_right(tls_slot, alpn_slot);
+        if (aws_channel_slot_set_handler(alpn_slot, alpn_handler)) {
+            return AWS_OP_ERR;
+        }
+    }
+
+    if (aws_tls_client_handler_start_negotiation(tls_handler)) {
```
justinboswell: Would be clearer if != AWS_OP_SUCCESS

source/channel_bootstrap.c
```diff
@@ -434,6 +692,112 @@ struct server_channel_data {
...
+    if (connection_args->user_on_error) {
+        connection_args->user_on_error(handler, slot, err, message, connection_args->tls_user_data);
+    }
+}
+
+static inline int s_setup_server_tls(struct server_connection_args *connection_args, struct aws_channel *channel) {
+    struct aws_channel_slot *tls_slot = NULL;
+    struct aws_channel_handler *tls_handler = NULL;
+
+    /* as far as cleanup goes here, since we're adding things to a channel, if a slot is ever successfully
```
justinboswell: Same comment here as above, I worry about the asymmetrical/non-atomic nature of these functions sometimes not cleaning up their broken state/resources, and there's no indication what the distinction is between when this function manages resources vs they're left to be effectively garbage collected.

source/channel_bootstrap.c
```diff
@@ -92,6 +162,13 @@ void aws_client_bootstrap_clean_up(struct aws_client_bootstrap *bootstrap) {
 struct client_channel_data {
     struct aws_channel channel;
     struct aws_socket *socket;
+    struct aws_tls_connection_options tls_options;
+    aws_channel_on_protocol_negotiated_fn *on_protocol_negotiated;
+    aws_tls_on_data_read_fn *user_on_data_read;
+    aws_tls_on_negotiation_result_fn *user_on_negotiation_result;
+    aws_tls_on_error_fn *user_on_error;
+    void *tls_user_data;
+    bool use_tls;
```
justinboswell: Is there any reason why this can't be divined from the presence of something in tls_options? Just feels slightly redundant

include/aws/io/tls_channel_handler.h
```diff
@@ -0,0 +1,328 @@
...
+typedef void(aws_tls_on_negotiation_result_fn)(
+    struct aws_channel_handler *handler,
+    struct aws_channel_slot *slot,
+    int err_code,
+    void *user_data);
+
+/**
+ * Only used if the TLS handler is the last handler in the channel. This allows you to read any data that
+ * was read and decrypted by the handler. If you have application protocol channel handlers, this function
+ * is not necessary and certainly not recommended.
```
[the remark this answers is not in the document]
JonathanHenson: you just want to do a channel that does TLS for the communication. There need not be a full blown application protocol on the other side.

E.g. an echo server

include/aws/io/tls_channel_handler.h
```diff
@@ -0,0 +1,328 @@
...
+     * Apple Only!
+     *
+     * On Apple OS you can also use a pkcs#12 file for your certificate
+     * and private key. This is the path to that file.
+     */
+    const char *pkcs12_path;
+    /**
+     * Password for the pkcs12 file in pkcs12_path.
+     */
+    const char *pkcs12_password;
```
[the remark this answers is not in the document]
JonathanHenson: we can and probably will support pkcs12 on windows as well, it just isn't necessary at the moment.

include/aws/io/tls_channel_handler.h
```diff
@@ -0,0 +1,328 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+/******************************** tls options init stuff ***********************/
+AWS_IO_API void aws_tls_ctx_options_init_default_client(struct aws_tls_ctx_options *options);
+AWS_IO_API void aws_tls_ctx_options_init_client_mtls(
+    struct aws_tls_ctx_options *options,
+    const char *cert_path,
+    const char *pkey_path);
+AWS_IO_API void aws_tls_ctx_options_init_client_mtls_pkcs12(
```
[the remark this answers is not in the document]
JonathanHenson: see above

include/aws/io/tls_channel_handler.h
```diff
@@ -0,0 +1,328 @@
...
+    const char *server_name);
+
+AWS_IO_API void aws_tls_connection_options_set_alpn_list(
+    struct aws_tls_connection_options *conn_options,
+    const char *alpn_list);
+AWS_IO_API void aws_tls_connection_options_set_verify_peer(
+    struct aws_tls_connection_options *conn_options,
+    bool verify_peer);
+
+/********************************* stuff that actually does work *********************************/
```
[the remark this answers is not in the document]
JonathanHenson: sure

include/aws/io/tls_channel_handler.h
```diff
@@ -0,0 +1,328 @@
...
+/**
+ * Creates a new client ctx. This ctx can be used for the lifetime of the application assuming you want the same
+ * options for every outgoing connection. Options will be copied.
+ */
+AWS_IO_API struct aws_tls_ctx *aws_tls_client_ctx_new(struct aws_allocator *alloc, struct aws_tls_ctx_options *options);
+
+/**
+ * Destroys the output from aws_tls_server_ctx_new and aws_tls_client_ctx_new.
+ */
+AWS_IO_API void aws_tls_ctx_destroy(struct aws_tls_ctx *ctx);
```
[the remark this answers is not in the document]
JonathanHenson: handled by vtable.

include/aws/io/channel_bootstrap.h
```diff
@@ -167,6 +185,28 @@ AWS_IO_API int aws_client_bootstrap_new_socket_channel(
...
+ * and `connection_options` for tls configuration.
+ * If AWS_SOCKET_LOCAL is used, host_name should be the name of the socket or named pipe, and port is ignored.
+ * If `host_name` is a dns address, it will be resolved prior to attempting a connection.
+ * `setup_callback` will be invoked once the channel is ready for use and TLS has been
+ * negotiated, or if an error is encountered. `shutdown_callback` will be invoked once the channel has shutdown.
+ * Immediately after the `shutdown_callback` returns, the channel is cleaned up automatically. All callbacks are invoked
+ * in the thread of the event-loop that the new channel is assigned to.
+ *
+ * `connection_options` is copied.
+ */
```
justinboswell: Add a comment that socket type must be STREAM, we don't do DTLS

source/channel_bootstrap.c
```diff
@@ -391,6 +621,22 @@ static inline int s_new_client_channel(
...
+    struct aws_client_bootstrap *bootstrap,
+    const char *host_name,
+    uint16_t port,
+    const struct aws_socket_options *options,
+    const struct aws_tls_connection_options *connection_options,
+    aws_client_bootstrap_on_channel_setup_fn *setup_callback,
+    aws_client_bootstrap_on_channel_shutdown_fn *shutdown_callback,
+    void *user_data) {
+    assert(connection_options);
+    assert(bootstrap->tls_ctx);
```
justinboswell: Add an assert to ensure that socket type is STREAM, we don't support DTLS

source/darwin/tls_channel_handler.c
```diff
@@ -0,0 +1,856 @@
...
+    secure_transport_handler->handler.alloc = allocator;
+    secure_transport_handler->handler.impl = secure_transport_handler;
+    secure_transport_handler->handler.vtable = &s_handler_vtable;
+    secure_transport_handler->wrapped_allocator = secure_transport_ctx->wrapped_allocator;
+    secure_transport_handler->protocol = NULL;
+    secure_transport_handler->ctx =
+        SSLCreateContext(secure_transport_handler->wrapped_allocator, protocol_side, kSSLStreamType);
+
+    if (!secure_transport_handler->ctx) {
+        /* TODO raise error here. */
```
justinboswell: Probably actually raise the error :)

source/pki_utils.c
```diff
@@ -1,298 +1,816 @@
...
+            return aws_raise_error(AWS_IO_MAX_FDS_EXCEEDED);
+        case ENOMEM:
+            return aws_raise_error(AWS_ERROR_OOM);
+        default:
+            return aws_raise_error(AWS_IO_SYS_CALL_FAILURE);
+    }
+}
+
+int aws_byte_buf_init_from_file(struct aws_byte_buf *out_buf, struct aws_allocator *alloc, const char *filename) {
+/* yeah yeah, I know and I don't care. */
```
justinboswell: Can probably remove this for public consumption

source/pki_utils.c
```diff
@@ -1,298 +1,816 @@
...
+            CFRelease(import_output);
+        }
+
+        if (status == errSecSuccess) {
+            CFTypeRef certs[] = {identity_output};
+            *identity = CFArrayCreate(cf_alloc, (const void **)certs, 1L, &kCFTypeArrayCallBacks);
+            return AWS_OP_SUCCESS;
+        }
+    } else {
+        CFRelease(import_keychain);
```
justinboswell: Doesn't the import_output still need to be cleaned up here?

source/pki_utils.c
```diff
@@ -1,298 +1,816 @@
...
+        password_ref = CFStringCreateWithBytes(cf_alloc, password->buffer, password->len, kCFStringEncodingUTF8, false);
+    }
+
+    CFDictionaryAddValue(dictionary, kSecImportExportPassphrase, password_ref);
+
+    OSStatus status = SecPKCS12Import(pkcs12_data, dictionary, &items);
+    CFRelease(pkcs12_data);
+
+    if (password_ref) {
+        CFRelease(password_ref);
```
justinboswell: Should we secure zero the memory first?

source/pki_utils.c
```diff
@@ -1,298 +1,816 @@
...
+    }
+
+    CFRelease(dictionary);
+
+    if (status == errSecSuccess) {
+        CFTypeRef item = (CFTypeRef)CFArrayGetValueAtIndex(items, 0);
```
justinboswell: assert(items)?

source/pki_utils.c
```diff
@@ -1,298 +1,816 @@
...
+    CFRelease(identity);
+}
+
+void aws_release_certificates(CFArrayRef certs) {
+    CFRelease(certs);
+}
+
+#endif
+
+#ifdef _WIN32
```
justinboswell: This should go in a windows/pki_utils.c, right?

source/pki_utils.c
```diff
@@ -1,298 +1,816 @@
...
+        return aws_raise_error(AWS_IO_FILE_INVALID_PATH);
+    }
+
+    /* according to the docs, this has to be the System Store by itself to open the cert store
+       successfully. There are some libraries such as libcurl that don't do this copy with just the
+       System Store. I haven't been able to make that work. I suspect mutual TLS just doesn't work
+       in curl on windows? However, this does work, so this is what we're doing.
+       As far as the 128 byte length, according to the msdn documentation here:
+       https://docs.microsoft.com/en-us/windows/desktop/SecCrypto/system-store-locations
+       this value should be more than enough. We'll add an assert just in case. */
```
justinboswell: I had a really hard time parsing this. I think what you're saying is that "the store_val value has to match the computed store_path and be less than 128 bytes in length, according to the docs: $url"

source/pki_utils.c
```diff
@@ -1,298 +1,816 @@
...
+        aws_array_list_get_at_ptr(&certificates, (void **)&byte_buf_ptr, i);
+
+        CERT_BLOB cert_blob;
+        CERT_CONTEXT *cert_context = NULL;
+
+        cert_blob.pbData = byte_buf_ptr->buffer;
+        cert_blob.cbData = (DWORD)byte_buf_ptr->len;
+
+        DWORD content_type = 0;
+        BOOL query_res = CryptQueryObject(
```
justinboswell: I'm sure it's in the docs, but is this just validating that this is a valid cert blob? Comment perhaps?

source/pki_utils.c
```diff
@@ -1,298 +1,816 @@
...
+                        current_cert_len += current_buf_ptr->len;
+                    }
+                    ++i;
+                }
+                break;
+        }
+    }
+
+end_of_loop:
+    aws_array_list_clean_up(&split_buffers);
```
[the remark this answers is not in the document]
JonathanHenson: they're just cursors into the memory (offsets and pointers)

source/pki_utils.c
```diff
@@ -1,298 +1,816 @@
...
+    SecKeychainCopyDefault(&import_keychain);
+
+    OSStatus status = SecItemImport(
+        aggregate_certificate_data, NULL, &format, &item_type, 0, &import_params, import_keychain, &import_output);
+
+    CFRelease(aggregate_certificate_data);
+    aws_byte_buf_clean_up(&aggregate_buffer);
+
+    if (status != errSecSuccess && status != errSecDuplicateItem) {
+        return AWS_OP_ERR;
```
graebm: not raising error code

source/pki_utils.c
```diff
@@ -1,298 +1,816 @@
...
+        CFRelease(certificate_ref);
+        CFRelease(import_keychain);
+        if (import_output && cleanup_import_output) {
+            CFRelease(import_output);
+        }
+
+        if (status == errSecSuccess) {
+            CFTypeRef certs[] = {identity_output};
+            *identity = CFArrayCreate(cf_alloc, (const void **)certs, 1L, &kCFTypeArrayCallBacks);
+            return AWS_OP_SUCCESS;
```
graebm: no error code is raised if status != errSecSuccess

source/s2n/tls_channel_handler.c
```diff
@@ -0,0 +1,738 @@
...
+
+    if (downstream_size <= current_window_size) {
+        size_t likely_records_count = (downstream_size - current_window_size) % MAX_RECORD_SIZE;
+        size_t offset_size = likely_records_count * (EST_TLS_RECORD_OVERHEAD);
+        size_t window_update_size = (downstream_size - current_window_size) + offset_size;
+        aws_channel_slot_increment_read_window(slot, window_update_size);
+    }
+
+    if (s2n_handler->negotiation_finished) {
+        /* we have messages in a queue and they need to be run after the socket has popped (even if it didn't have data
```
justinboswell: Same as the other place where this comment is copied from, this comment could be more clear

source/s2n/tls_channel_handler.c
```diff
@@ -0,0 +1,738 @@
...
+    .process_read_message = s_s2n_handler_process_read_message,
+    .process_write_message = s_s2n_handler_process_write_message,
+    .shutdown = s_s2n_handler_shutdown,
+    .increment_read_window = s_s2n_handler_increment_read_window,
+    .initial_window_size = s_s2n_handler_initial_window_size,
+};
+
+static int s_parse_protocol_preferences(
+    const char *alpn_list_str,
+    const char protocol_output[4][128],
```
justinboswell: 4 is probably enough, but is there any reason this is hardcoded?

source/s2n/tls_channel_handler.c
```diff
@@ -0,0 +1,738 @@
...
+        case AWS_IO_TLSv1:
+            s2n_config_set_cipher_preferences(s2n_ctx->s2n_config, "CloudFront-TLS-1-0-2016");
+            break;
+        case AWS_IO_TLSv1_1:
+            s2n_config_set_cipher_preferences(s2n_ctx->s2n_config, "CloudFront-TLS-1-1-2016");
+            break;
+        case AWS_IO_TLSv1_2:
+            s2n_config_set_cipher_preferences(s2n_ctx->s2n_config, "CloudFront-TLS-1-2-2018");
+            break;
+        case AWS_IO_TLSv1_3:
```
justinboswell: This is a failure on darwin, should it be a failure here? Is falling back to a version of TLS that the user isn't expecting an ok thing to do here?

tests/alpn_handler_test.c
```diff
@@ -0,0 +1,316 @@
...
+
+    struct alpn_test_on_negotiation_args on_negotiation_args = {
+        .new_slot = NULL, .protocol = {0}, .new_handler = NULL, .allocator = allocator};
+
+    struct aws_channel_handler *handler =
+        aws_tls_alpn_handler_new(allocator, s_alpn_tls_successful_negotiation, &on_negotiation_args);
+    ASSERT_NOT_NULL(handler);
+    ASSERT_SUCCESS(aws_channel_slot_set_handler(slot, handler));
+
+    struct aws_tls_negotiated_protocol_message protocol_message = {.protocol = aws_byte_buf_from_c_str("h2")};
```
justinboswell: Should we add a separate test for a list of protocols?

tests/tls_handler_test.c
```diff
@@ -0,0 +1,699 @@
...
+
+    struct aws_event_loop_group el_group;
+    ASSERT_SUCCESS(aws_event_loop_group_default_init(&el_group, allocator, 0));
+
+    struct aws_mutex mutex = AWS_MUTEX_INIT;
+    struct aws_condition_variable condition_variable = AWS_CONDITION_VARIABLE_INIT;
+
+    struct aws_tls_ctx_options client_ctx_options;
+    aws_tls_ctx_options_init_default_client(&client_ctx_options);
+    aws_tls_ctx_options_override_default_trust_store(&client_ctx_options, "/etc/ssl/certs", NULL);
```
justinboswell: What's this going to do on windows?

source/windows/iocp/iocp_event_loop.c
```diff
@@ -118,10 +124,23 @@ void aws_overlapped_reset(struct aws_overlapped *overlapped) {
...
+    .connect_to_io_completion_port = s_connect_to_io_completion_port,
+    .is_on_callers_thread = s_is_event_thread,
+    .unsubscribe_from_io_events = s_unsubscribe_from_io_events,
+};
+
 struct aws_event_loop *aws_event_loop_new_default(struct aws_allocator *alloc, aws_io_clock_fn *clock) {
     assert(alloc);
     assert(clock);
 
+    /* load the windows dlls and dynamically bind our soul to Satan's dark angels. */
```
justinboswell: I like that this was an addition on its own. That said, maybe not for public release :)

source/channel_bootstrap.c
```diff
@@ -36,6 +37,55 @@
...
+
+        struct aws_task task = {
+            .fn = s_handle_tl_cleanup_task,
+            .arg = &tl_shutdown,
+        };
+
+        aws_event_loop_schedule_task_now(el, &task);
+        aws_condition_variable_wait_pred(&condition_variable, &mutex, s_tl_cleanup_predicate, &tl_shutdown);
+    }
+    aws_mutex_unlock(&mutex);
```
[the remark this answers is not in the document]
JonathanHenson: I don't think this is accurate.
1st, by definition with condition variable wait will unlock the mutex when it waits. Since it's in a different thread, this guarantees the task lock isn't acquired until we try to wait.
2nd, two scenarios, if predicate prevents the wait by returning true, the mutex was never unlocked, otherwise, the condition var wait will reaquire the lock before returning, hence the needed unlock. 

I did move the unlock in the task above the notify though.

source/channel_bootstrap.c
JonathanHenson: so the readability part.... done

On the first comment, yeah, that's good feedback. We should be consistent on error means I didn't take ownership of your memory thing. But.... in this case.... that's not the case, because it's the window update that failed and the handler was set. Let's circle back on this after this PR.

source/channel_bootstrap.c
JonathanHenson: yeah, agreed, we need to clean those semantics up, in a different PR

source/pki_utils.c
```diff
@@ -1,298 +1,816 @@
...
+int aws_read_and_decode_pem_file_to_buffer_list(
+    struct aws_allocator *alloc,
+    const char *filename,
+    struct aws_array_list *cert_chain_or_key) {
+    struct aws_byte_buf raw_file_buffer;
+
+    if (aws_byte_buf_init_from_file(&raw_file_buffer, alloc, filename)) {
+        return AWS_OP_ERR;
+    }
+    assert(raw_file_buffer.buffer);
```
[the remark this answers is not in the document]
JonathanHenson: only way buffer is null is if something bad went wrong. allocation size is always at least 1.

source/pki_utils.c
```diff
@@ -1,298 +1,816 @@
...
+        return AWS_OP_ERR;
+    }
+
+    aws_secure_zero(raw_file_buffer.buffer, raw_file_buffer.len);
+    aws_byte_buf_clean_up(&raw_file_buffer);
+
+    return AWS_OP_SUCCESS;
+}
+
+#ifdef __MACH__
```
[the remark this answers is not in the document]
JonathanHenson: I'll circle back on this

source/pki_utils.c
JonathanHenson: unfortunately, it also fills in the cert context object for use in the next API call.

source/s2n/tls_channel_handler.c
```diff
@@ -0,0 +1,738 @@
...
+struct s2n_ctx {
+    struct aws_tls_ctx ctx;
+    struct s2n_config *s2n_config;
+};
+
+void aws_tls_init_static_state(struct aws_allocator *alloc) {
+
+    (void)alloc;
+
+    setenv("S2N_ENABLE_CLIENT_MODE", "1", 1);
```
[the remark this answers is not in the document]
JonathanHenson: all this does is turn on client mode (it was in experimental mode until I wrote the x.509 code last year). This will go away once s2n goes 1.0

source/s2n/tls_channel_handler.c
```diff
@@ -0,0 +1,738 @@
...
+        struct aws_byte_cursor message_cursor = aws_byte_cursor_from_buf(&message->message_data);
+        aws_byte_cursor_advance(&message_cursor, message->copy_mark);
+        aws_byte_cursor_read(&message_cursor, buf->buffer + written, to_write);
+
+        written += to_write;
+
+        message->copy_mark += to_write;
+
+        if (message->copy_mark == message->message_data.len) {
+            /* note: value is the first member of the allocated struct */
```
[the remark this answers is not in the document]
JonathanHenson: left over from my very first iteration on intrusive lists.

source/s2n/tls_channel_handler.c
```diff
@@ -0,0 +1,738 @@
...
+                message->message_tag = AWS_TLS_NEGOTIATED_PROTOCOL_MESSAGE;
+                struct aws_tls_negotiated_protocol_message *protocol_message =
+                    (struct aws_tls_negotiated_protocol_message *)message->message_data.buffer;
+
+                protocol_message->protocol = s2n_handler->protocol;
+                message->message_data.len = sizeof(struct aws_tls_negotiated_protocol_message);
+                if (aws_channel_slot_send_message(s2n_handler->slot, message, AWS_CHANNEL_DIR_READ)) {
+                    aws_channel_release_message_to_pool(s2n_handler->slot->channel, message);
+                    aws_channel_shutdown(s2n_handler->slot->channel, aws_last_error());
+                    return AWS_OP_SUCCESS;
```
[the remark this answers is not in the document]
JonathanHenson: SUCCESS means, "I processed the message it's mine don't touch it" if you return error, it's the adjacent hander's responsibility to release the message tot he pool.

source/s2n/tls_channel_handler.c
```diff
@@ -0,0 +1,738 @@
...
+        aws_linked_list_push_back(&s2n_handler->input_queue, &message->queueing_handle);
+
+        if (!s2n_handler->negotiation_finished) {
+            size_t message_len = message->message_data.len;
+            if (!s_drive_negotiation(handler)) {
+                aws_channel_slot_increment_read_window(slot, message_len);
+            } else {
+                aws_channel_shutdown(s2n_handler->slot->channel, AWS_IO_TLS_ERROR_NEGOTIATION_FAILURE);
+            }
+            return AWS_OP_SUCCESS;
```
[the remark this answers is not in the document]
JonathanHenson: I'll add a generic note to the channel handler api.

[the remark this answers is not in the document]
JonathanHenson: added to the aws_channel_slot_send_message documentation 

* NOTE: if this function returns an error code, it is the caller's responsibility to release message
 * back to the pool. If this function returns AWS_OP_SUCCESS, the recipient of the message has taken
 * ownership of the message. So, for example, don't release a message to the pool and then return an error.
 * If you encounter an error condition in this case, shutdown the channel with the appropriate error code.

source/s2n/tls_channel_handler.c
```diff
@@ -0,0 +1,738 @@
...
+        }
+
+        ssize_t read = s2n_recv(
+            s2n_handler->connection,
+            outgoing_read_message->message_data.buffer,
+            outgoing_read_message->message_data.capacity,
+            &blocked);
+
+        /* weird race where we received an alert from the peer, but s2n doesn't tell us about it.....
+         * if this happens, it's a graceful shutdown, so kick it off here. */
```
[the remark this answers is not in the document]
JonathanHenson: added this comment:
In other words, s2n, upon graceful shutdown, follows the unix EOF idiom. So just shutdown with
         * SUCCESS.

source/s2n/tls_channel_handler.c
JonathanHenson: yeah, save on heap allocations

source/s2n/tls_channel_handler.c
JonathanHenson: yeah, should be an error.

tests/alpn_handler_test.c
JonathanHenson: there's no universe where you'd get a list of protocols after a TLS negotiation.

tests/tls_handler_test.c
JonathanHenson: nothing. It will probably break on non debian linux distributions though. We need to add some path detection stuff to pki utils. Let's move that to a separate review though.

source/windows/secure_channel_tls_handler.c
```diff
@@ -0,0 +1,1499 @@
...
+    PCERT_CONTEXT pcerts;
+    HCERTSTORE cert_store;
+    HCERTSTORE custom_trust_store;
+};
+
+struct secure_channel_handler {
+    struct aws_channel_handler handler;
+    CtxtHandle sec_handle;
+    CredHandle creds;
+    PCCERT_CONTEXT cert_context[1];
```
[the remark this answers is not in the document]
JonathanHenson: The API for SSPI expects an array of PCERT_CONTEXT with only the leaf certificate associated with its private key.

source/windows/secure_channel_tls_handler.c
```diff
@@ -0,0 +1,1499 @@
...
+       the leaf certificate against. */
+    CERT_CHAIN_ENGINE_CONFIG engine_config;
+    AWS_ZERO_STRUCT(engine_config);
+    engine_config.cbSize = sizeof(engine_config);
+    engine_config.hExclusiveRoot = sc_handler->custom_ca_store;
+
+    HCERTCHAINENGINE engine = NULL;
+    if (!CertCreateCertificateChainEngine(&engine_config, &engine)) {
+        CertFreeCertificateContext(peer_certificate);
+        return AWS_OP_ERR;
```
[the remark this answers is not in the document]
JonathanHenson: the caller raises the error.

source/windows/secure_channel_tls_handler.c
```diff
@@ -0,0 +1,1499 @@
...
+            peer_certificate,
+            NULL,
+            peer_certificate->hCertStore,
+            &chain_params,
+            CERT_CHAIN_REVOCATION_CHECK_CHAIN,
+            NULL,
+            &cert_chain_ctx)) {
+        CertFreeCertificateChainEngine(engine);
+        CertFreeCertificateContext(peer_certificate);
+        return AWS_OP_ERR;
```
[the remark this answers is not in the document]
JonathanHenson: see above

source/windows/secure_channel_tls_handler.c
```diff
@@ -0,0 +1,1499 @@
...
+    aws_array_list_init_static(&alpn_buffers, alpn_buffer_array, 4, sizeof(struct aws_byte_cursor));
+    struct aws_byte_buf alpn_str_buf = aws_byte_buf_from_c_str(sc_handler->options.alpn_list);
+    if (aws_byte_buf_split_on_char(&alpn_str_buf, ';', &alpn_buffers)) {
+        return AWS_OP_ERR;
+    }
+
+    size_t protocols_count = aws_array_list_length(&alpn_buffers);
+
+    size_t index = 0;
+    CHECK_ALPN_BUFFER_SIZE(buffer_size, index + sizeof(uint32_t), alpn_buffers)
```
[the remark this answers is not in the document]
JonathanHenson: because the fields are variable length

source/windows/secure_channel_tls_handler.c
```diff
@@ -0,0 +1,1499 @@
...
+    index += sizeof(uint32_t);
+    CHECK_ALPN_BUFFER_SIZE(buffer_size, index + sizeof(uint32_t), alpn_buffers)
+    uint32_t *extension_name = (uint32_t *)&alpn_buffer_data[index];
+    index += sizeof(uint32_t);
+    CHECK_ALPN_BUFFER_SIZE(buffer_size, index + sizeof(uint32_t), alpn_buffers)
+    uint16_t *protocols_byte_length = (uint16_t *)&alpn_buffer_data[index];
+    index += sizeof(uint16_t);
+    CHECK_ALPN_BUFFER_SIZE(buffer_size, index + sizeof(uint16_t), alpn_buffers)
+
+    *extension_length += sizeof(uint32_t) + sizeof(uint16_t);
```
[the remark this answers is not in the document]
JonathanHenson: it's not inclusive :(

source/windows/secure_channel_tls_handler.c
```diff
@@ -0,0 +1,1499 @@
...
+        return AWS_OP_ERR;
+    }
+
+    if (status == SEC_E_INCOMPLETE_MESSAGE) {
+        sc_handler->estimated_incomplete_size = input_buffers[1].cbBuffer;
+        return aws_raise_error(AWS_IO_READ_WOULD_BLOCK);
+    };
+
+    if (status == SEC_I_CONTINUE_NEEDED || status == SEC_E_OK) {
+        for (size_t i = 0; i < output_buffers_desc.cBuffers; ++i) {
```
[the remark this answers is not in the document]
JonathanHenson: It's actually not until it has a complete message to send.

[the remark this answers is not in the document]
JonathanHenson: I added a comment explaining it.

source/windows/secure_channel_tls_handler.c
```diff
@@ -0,0 +1,1499 @@
...
+        /* if a custom CA store was configured, we have to do the verification ourselves. */
+        if (sc_handler->custom_ca_store) {
+            if (s_manually_verify_peer_cert(handler)) {
+                aws_raise_error(AWS_IO_TLS_ERROR_NEGOTIATION_FAILURE);
+                s_invoke_negotiation_error(handler, AWS_IO_TLS_ERROR_NEGOTIATION_FAILURE);
+                return AWS_OP_ERR;
+            }
+        }
+        /* do all the negotiation completion callbacks here. also, invoke it in a task so that we don't have a race
+        in the next section where we handle any left over data recieved from the server. */
```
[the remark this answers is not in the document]
JonathanHenson: deleted

source/windows/secure_channel_tls_handler.c
```diff
@@ -0,0 +1,1499 @@
...
+            aws_channel_release_message_to_pool(slot->channel, message);
+            return AWS_OP_SUCCESS;
+        }
+
+        aws_channel_shutdown(slot->channel, err);
+        return AWS_OP_ERR;
+    }
+
+    if (sc_handler->buffered_read_out_data_buf.len) {
+        if (s_process_pending_output_messages(handler)) {
```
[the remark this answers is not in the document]
JonathanHenson: by this point the error was already raised.

source/windows/secure_channel_tls_handler.c
```diff
@@ -0,0 +1,1499 @@
...
+    }
+
+    return AWS_OP_SUCCESS;
+}
+
+static int s_increment_read_window(struct aws_channel_handler *handler, struct aws_channel_slot *slot, size_t size) {
+    (void)size;
+    struct secure_channel_handler *sc_handler = (struct secure_channel_handler *)handler->impl;
+
+    if (!sc_handler->stream_sizes.cbMaximumMessage) {
```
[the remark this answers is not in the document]
JonathanHenson: branch prediction hints in msvc are a no-op, but I supose clang will eventually be used here so I'll add it.

tests/tls_handler_test.c
justinboswell: 👍

source/pki_utils.c
```diff
@@ -1,298 +1,816 @@
...
+    import_params.version = SEC_KEY_IMPORT_EXPORT_PARAMS_VERSION;
+    import_params.passphrase = CFSTR("");
+
+    SecKeychainRef import_keychain = NULL;
+    SecKeychainCopyDefault(&import_keychain);
+
+    OSStatus status = SecItemImport(
+        aggregate_certificate_data, NULL, &format, &item_type, 0, &import_params, import_keychain, &import_output);
+
+    CFRelease(aggregate_certificate_data);
```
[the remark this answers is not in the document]
JonathanHenson: it's not a copy. I zeroed aggregate buffer instead.

source/pki_utils.c
JonathanHenson: no, it's ref counted and gets picked up in the clean up code.

source/pki_utils.c
JonathanHenson: password lives on the tls_ctx.

source/pki_utils.c
JonathanHenson: we already checked that the operation was successful.

source/darwin/tls_channel_handler.c
```diff
@@ -0,0 +1,856 @@
...
+/* I have no idea if this code is correct, I can't test it until I have a machine with high-sierra on it
+ * but my employer hasn't pushed it out yet so.... sorry about that. */
+#if ALPN_AVAILABLE
+    struct aws_byte_buf alpn_data = aws_byte_buf_from_c_str(alpn_list);
+    struct aws_array_list alpn_list_array;
+    if (aws_array_list_init_dynamic(&alpn_list_array, alloc, 2, sizeof(struct aws_byte_cursor))) {
+        return;
+    }
+
+    if (aws_byte_buf_split_on_char(&alpn_data, ',', &alpn_list_array)) {
```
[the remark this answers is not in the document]
JonathanHenson: good catch.

source/darwin/tls_channel_handler.c
```diff
@@ -0,0 +1,856 @@
...
+            message->message_tag = AWS_TLS_NEGOTIATED_PROTOCOL_MESSAGE;
+            struct aws_tls_negotiated_protocol_message *protocol_message =
+                (struct aws_tls_negotiated_protocol_message *)message->message_data.buffer;
+
+            protocol_message->protocol = aws_byte_buf_from_array(
+                (uint8_t *)CFStringGetCStringPtr(secure_transport_handler->protocol, kCFStringEncodingUTF8),
+                (size_t)CFStringGetLength(secure_transport_handler->protocol));
+
+            message->message_data.len = sizeof(struct aws_tls_negotiated_protocol_message);
+            if (aws_channel_slot_send_message(secure_transport_handler->parent_slot, message, AWS_CHANNEL_DIR_READ)) {
```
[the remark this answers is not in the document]
JonathanHenson: that's not the semantics of AWS_OP_SUCCESS here. We called shutdown on the channel.

PR #48, opened 2018-11-06
replace aws_channel_init() with aws_channel_new()

graebm:
Channel's can't just vanish, we'll need to move to some kind of ref-counting scheme, so in preparation...

- Replace _init() and _clean_up() with _new() and _destroy().
- Move channel struct declaration to .C file, since it doesn't need to be public anymore.
- Remove some asserts 😞 that were checking against aws_channel's internal members (which are now private). I can put in query functions if we thinks these asserts (mostly in unit tests) are still valuable.
- Fix aws_socket_write() so that, if it returns AWS_OP_ERROR, then the write callback is not ALSO invoked with an error code.
    - Remove hacks in socket_channel_handler which were working around the previous behavior.

### DIFF
--- a/source/channel.c
+++ b/source/channel.c
@@ -32,6 +32,31 @@ enum {
     KB_16 = 16 * 1024,
 };
 
+enum aws_channel_state {
+    AWS_CHANNEL_SETTING_UP,
+    AWS_CHANNEL_ACTIVE,
+    AWS_CHANNEL_SHUTTING_DOWN,
+    AWS_CHANNEL_SHUT_DOWN,
+};
+
+struct aws_shutdown_notification_task {
+    struct aws_task task;
+    int error_code;
+    struct aws_channel_slot *slot;
+    bool shutdown_immediately;
+};
+
+struct aws_channel {
+    struct aws_allocator *alloc;
+    struct aws_event_loop *loop;
+    struct aws_channel_slot *first;
+    struct aws_message_pool *msg_pool;
+    enum aws_channel_state channel_state;
+    struct aws_shutdown_notification_task shutdown_notify_task;
+    aws_channel_on_shutdown_completed_fn *on_shutdown_completed;
+    void *shutdown_user_data;
+};
+
 struct channel_setup_args {
     struct aws_allocator *alloc;
     struct aws_channel *channel;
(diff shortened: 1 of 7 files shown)


source/channel.c
```diff
@@ -154,28 +184,18 @@ static void s_cleanup_slot(struct aws_channel_slot *slot) {
...
-
-    if (!current) {
-        channel->channel_state = AWS_CHANNEL_SHUT_DOWN;
-        return;
-    }
-
-    if (!current->handler) {
-        channel->channel_state = AWS_CHANNEL_SHUT_DOWN;
-    }
+void aws_channel_destroy(struct aws_channel *channel) {
```
justinboswell: The documentation says that aws_channel_shutdown() must have been called. Should we not assert on this at the top of this function?

source/posix/socket.c
```diff
@@ -1029,30 +1030,38 @@ static int s_process_write_requests(struct aws_socket *socket, bool spawned_from
...
 
     if (purge) {
-        aws_raise_error(aws_error);
         while (!aws_linked_list_empty(&socket_impl->write_queue)) {
             struct aws_linked_list_node *node = aws_linked_list_pop_front(&socket_impl->write_queue);
             struct write_request *write_request = AWS_CONTAINER_OF(node, struct write_request, node);
 
-            write_request->written_fn(socket, aws_error, 0, write_request->write_user_data);
+            /* If aws_socket_write() invoked this fn, don't invoke its write_request's error callback.
+               The user will learn about the error from aws_socket_write()'s return value */
```
justinboswell: This was hard to parse, perhaps:
If this fn was invoked directly from aws_socket_write(), don't invoke the error callback as the user will be able to rely on the return value from aws_socket_write()

source/posix/socket.c
```diff
@@ -975,7 +975,7 @@ int aws_socket_shutdown_dir(struct aws_socket *socket, enum aws_channel_directio
  * 1st scenario, someone called write and we want to at least try and make a write call if we can or return an
  * error if something bad has happened to the socket.
  * 2nd scenario, the event loop notified us that the socket went writable. */
-static int s_process_write_requests(struct aws_socket *socket, bool spawned_from_event) {
+static int s_process_write_requests(struct aws_socket *socket, struct write_request *spawned_from_request) {
```
justinboswell: Any reason not to use parent_request nomenclature instead? Makes some of the logic easier to follow, at least for me.

source/channel.c
graebm: I put back the old logic which checked this. I didn't 100% grok it before. I added a comment explaining things.

source/posix/socket.c
graebm: sure, I was using an arg name similar to the one it was replacing, but I admit it was awkward

source/posix/socket.c
graebm: copy/pasted

PR #50, opened 2018-11-07
install libcrypto locks for older libcrypto versions, until we can go…

JonathanHenson:
… rewrite libcrypto for s2n to use.

### DIFF
--- a/source/s2n/s2n_tls_channel_handler.c
+++ b/source/s2n/s2n_tls_channel_handler.c
@@ -26,10 +26,43 @@
 #include <stdio.h>
 #include <stdlib.h>
 
+#include <openssl/crypto.h>
+
 #define EST_TLS_RECORD_OVERHEAD 53 /* 5 byte header + 32 + 16 bytes for padding */
 #define KB_1 1024
 #define MAX_RECORD_SIZE (KB_1 * 16)
 #define EST_HANDSHAKE_SIZE (7 * KB_1)
+
+/* this is completely absurd and the reason I hate dependencies, but I'm assuming
+ * you don't want your older versions of openssl's libcrypto crashing on you. */
+#if defined(LIBRESSL_VERSION_NUMBER) && (OPENSSL_VERSION_NUMBER == 0x20000000L)
+#undef OPENSSL_VERSION_NUMBER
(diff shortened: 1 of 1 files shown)


source/s2n/s2n_tls_channel_handler.c
```diff
@@ -57,6 +90,23 @@ void aws_tls_init_static_state(struct aws_allocator *alloc) {
     setenv("S2N_ENABLE_CLIENT_MODE", "1", 1);
     setenv("S2N_DONT_MLOCK", "1", 1);
     s2n_init();
+
+#if OPENSSL_VERSION_LESS_1_1
+    if (!CRYPTO_get_locking_callback()) {
+        s_libcrypto_allocator = alloc;
+        s_libcrypto_locks = aws_mem_acquire(alloc, sizeof(struct aws_mutex) * CRYPTO_num_locks());
+        assert(s_libcrypto_locks);
```
graebm: why assert when you can exit() 🕶

source/s2n/s2n_tls_channel_handler.c
```diff
@@ -67,6 +117,21 @@ void aws_tls_clean_up_thread_local_state(void) {
 
 void aws_tls_clean_up_static_state(void) {
     s2n_cleanup();
+
+#if OPENSSL_VERSION_LESS_1_1
+    if (CRYPTO_get_locking_callback() == s_locking_fn) {
+        CRYPTO_set_locking_callback(NULL);
```
graebm: should we statically store whether or not we were the ones to install the locking callback?

as is, we don't set it if it was already there, but we'll remove it either way

source/s2n/s2n_tls_channel_handler.c
justinboswell: We need that hard crashing assert methinks

graebm: aws_fatal()

source/s2n/s2n_tls_channel_handler.c
JonathanHenson: doesn't this branch check exactly that? The only way we remove it is if the locking callback is the one we would have installed.

PR #54, opened 2018-11-09
Task cancellation fixes

graebm:
Small epoll task scheduling fixes.

Change how channel schedules tasks, to avoid edge-cases and race-conditions when cancelling tasks.

Specifics:
- Epoll
    - Fix bug where all tasks were scheduled as "future" tasks.
    - Cross-thread tasks are moved into scheduler immediately before running tasks. This doesn't fix any bugs, it's just for consistency with kqueue and IOCP.
- Channels
    - Since event-loop can't cancel off-thread tasks reliably, don't schedule off-thread tasks immediately. The channel stores off-thread aws_channel_tasks in a lock-protected-list until it can migrate these tasks to the event-loop on the proper thread. Now cancelling off-thread tasks just means cleaning up that list.
    - Don't access `channel->channel_state` outside the event-loop thread. Check the lock-protected `channel->cross_thread_tasks.is_channel_shut_down` instead.

### DIFF
--- a/source/linux/epoll_event_loop.c
+++ b/source/linux/epoll_event_loop.c
@@ -431,8 +431,8 @@ static void s_process_task_pre_queue(struct aws_event_loop *event_loop) {
 
     aws_mutex_unlock(&epoll_loop->task_pre_queue_mutex);
 
-    while (!aws_linked_list_empty(&epoll_loop->task_pre_queue)) {
-        struct aws_linked_list_node *node = aws_linked_list_pop_front(&epoll_loop->task_pre_queue);
+    while (!aws_linked_list_empty(&task_pre_queue)) {
+        struct aws_linked_list_node *node = aws_linked_list_pop_front(&task_pre_queue);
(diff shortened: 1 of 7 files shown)


source/linux/epoll_event_loop.c
```diff
@@ -394,22 +402,45 @@ static void s_on_tasks_to_schedule(
...
+    /* several tasks could theoretically have been written (though this should never happen), make sure we drain the
+     * eventfd/pipe. */
+    while (read(epoll_loop->read_task_handle.data.fd, &count_ignore, sizeof(count_ignore)) > -1) {
+    }
+
+    aws_linked_list_swap_contents(&epoll_loop->task_pre_queue, &task_pre_queue);
+
+    aws_mutex_unlock(&epoll_loop->task_pre_queue_mutex);
+
+    while (!aws_linked_list_empty(&epoll_loop->task_pre_queue)) {
```
JonathanHenson: I'm pretty sure this is a copy pasta error, and you meant:

while (!aws_linked_list_empty(&task_pre_queue)) {

etc.... I've already fixed it, pushing it up.

PR #59, opened 2018-11-16
Proper CMake dependency management. Now consumers only need to know a…

JonathanHenson:
…bout Aws::aws-c-io.

### DIFF
--- a/cmake/aws-c-io-config.cmake
+++ b/cmake/aws-c-io-config.cmake
@@ -1,0 +1,9 @@
+include(CMakeFindDependencyMacro)
+
+if (UNIX AND NOT APPLE)
+    find_dependency(s2n)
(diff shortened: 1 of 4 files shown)


cmake/aws-c-io-config.cmake
```diff
@@ -0,0 +1,9 @@
+include(CMakeFindDependencyMacro)
+
+if (UNIX AND NOT APPLE)
+    find_dependency(s2n)
+endif()
+
+find_dependency(aws-c-common)
+
+include(${CMAKE_CURRENT_LIST_DIR}/@CMAKE_PROJECT_NAME@-targets.cmake)
```
justinboswell: What do the @'s do? I've never seen that before.

PR #60, opened 2018-11-17
Apparently this has been broken for a little while and code build had…

JonathanHenson:
… a bug that changed my buildspec unexpectedely so we didn't know.

### DIFF
--- a/codebuild/common-linux.sh
+++ b/codebuild/common-linux.sh
@@ -27,7 +27,7 @@ sudo apt-get install libssl-dev
 
 cd ../
 
-mkdir install
+mkdir -p install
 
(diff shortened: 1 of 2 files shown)


codebuild/common-linux.sh
```diff
@@ -27,7 +27,7 @@ sudo apt-get install libssl-dev
 
 cd ../
 
-mkdir install
+mkdir -p install
```
justinboswell: Isn't -p just make all the directories required to satisfy the path argument? Why -p here?

JonathanHenson: In the case of 32bit builds the libcrypto  build already created the directory.

-p basically makes it not error that the directory already exists.

JonathanHenson:
This was fixed in a different PR. Closing.

PR #66, opened 2018-12-07
Made sure socket handler reads before closing so that we catch data b…

JonathanHenson:
…etween the last read and a close event.

### DIFF
--- a/source/socket_channel_handler.c
+++ b/source/socket_channel_handler.c
@@ -165,9 +165,9 @@ static void s_on_readable_notification(struct aws_socket *socket, int error_code
     (void)socket;
 
     struct socket_handler *socket_handler = user_data;
-    if (!error_code) {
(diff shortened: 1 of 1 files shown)


source/socket_channel_handler.c
```diff
@@ -165,9 +165,9 @@ static void s_on_readable_notification(struct aws_socket *socket, int error_code
     (void)socket;
 
     struct socket_handler *socket_handler = user_data;
-    if (!error_code) {
-        s_do_read(socket_handler);
-    } else if (!socket_handler->shutdown_in_progress) {
+    s_do_read(socket_handler);
```
justinboswell: Probably should drop a little comment here to explain why we're reading when we may know that the socket is in an error state

JonathanHenson:
Closing as this change was merge in a different PR.

PR #69, opened 2018-12-14
Don't reschedule s2n read task if it's already scheduled

ColdenCullen:


graebm:
`sequential_tasks` is also used when `aws_tls_client_handler_start_negotiation()` is called from off-thread. It could end up doing the same kind of stomp.

graebm:
Just talked with Jonathan. It's being used for 2 things now:
1) kicking off negotiation
2) doing something after negotiation is complete

So those 2 things can't overlap. We just had an issue where use #2 was happening multiple times and stomping itself.

PR #72, opened 2018-12-14
Fixed race condition where sockets can be deleted by errors before we…

justinboswell:
…'re done bootstrapping

### DIFF
--- a/source/windows/iocp/socket.c
+++ b/source/windows/iocp/socket.c
@@ -716,15 +716,17 @@ void s_socket_connection_completion(
         struct aws_socket *socket = socket_args->socket;
         socket->readable_fn = NULL;
         socket->readable_user_data = NULL;
-        socket_args->socket = NULL;
         socket_impl->connect_args = NULL;
 
         if (!status_code) {
             socket_impl->vtable->connection_success(socket);
         } else {
+            socket_args->socket->state = ERRORED;
             int error = s_determine_socket_error(status_code);
             socket_impl->vtable->connection_error(socket, error);
         }
+
+        socket_args->socket = NULL;
     }
 
     if (operation_data->socket) {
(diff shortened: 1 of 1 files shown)


source/windows/iocp/socket.c
```diff
@@ -716,15 +716,17 @@ void s_socket_connection_completion(
...
         socket->readable_user_data = NULL;
-        socket_args->socket = NULL;
         socket_impl->connect_args = NULL;
 
         if (!status_code) {
             socket_impl->vtable->connection_success(socket);
         } else {
+            socket_args->socket->state = ERRORED;
             int error = s_determine_socket_error(status_code);
             socket_impl->vtable->connection_error(socket, error);
```
graebm: `connection_error()` is already setting the socket state to ERRORED

source/windows/iocp/socket.c
```diff
@@ -716,15 +716,17 @@ void s_socket_connection_completion(
...
         }
+
+        socket_args->socket = NULL;
```
graebm: why did this get moved down here?

source/windows/iocp/socket.c
```diff
@@ -805,29 +811,28 @@ static inline int s_tcp_connect(
...
     /* with IO completion ports, the overlapped callback triggers even if the operation succedded immediately,
        so we can just act like it's pending and the code path is the same.*/
     if (!connect_res) {
         int error_code = WSAGetLastError();
         if (error_code != ERROR_IO_PENDING) {
-            socket_impl->read_io_data->in_use = false;
-            aws_mem_release(socket->allocator, connect_args);
             socket_impl->connect_args = NULL;
-            socket->state = ERRORED;
+            socket_impl->read_io_data->in_use = false;
```
graebm: if the `socket` might be killed from callbacks above, then won't the `socket_impl` also be deleted?

JonathanHenson: yeah, this was done when we believed that the completion port didn't get invoked in an error condition. Since we now know it does, most of this branch can just be deleted.

source/windows/iocp/socket.c
JonathanHenson: this isn't need at all.

JonathanHenson: oh wait, wrong function, nevermind, this is fine.

source/windows/iocp/socket.c
JonathanHenson: good catch.

justinboswell: Fixed.

source/windows/iocp/socket.c
justinboswell: This was moved because of the above change, which we no longer need. Reverted.

PR #90, opened 2019-02-12
Improvements to CI scripts.

JonathanHenson:


codebuild/common-windows.bat
```diff
@@ -20,7 +20,7 @@ goto :EOF
 mkdir %BUILDS_DIR%\%~1-build
 cd %BUILDS_DIR%\%~1-build
 git clone https://github.com/awslabs/%~1.git
-cmake %CMAKE_ARGS% -DCMAKE_BUILD_TYPE="RelWithDebInfo" -DCMAKE_INSTALL_PREFIX=%INSTALL_DIR% -DCMAKE_PREFIX_PATH=%INSTALL_DIR% ..\%~1 || goto error
+cmake %CMAKE_ARGS% -DCMAKE_BUILD_TYPE="RelWithDebInfo" -DCMAKE_INSTALL_PREFIX=%INSTALL_DIR% -DCMAKE_PREFIX_PATH=%INSTALL_DIR% %~1 || goto error
```
justinboswell: Just a note: We should make sure we're consistent about RelWithDebInfo for all of our builds, so they can link with DWARF/PDB data, but then strip it if the final target doesn't want it. Avoids linker errors, and gives the best customer experience.

PR #91, opened 2019-02-13
Fix unowned uv loop hang and incomplete host resolves

ColdenCullen:

### DIFF
--- a/source/host_resolver.c
+++ b/source/host_resolver.c
@@ -729,7 +729,9 @@ int aws_host_resolver_init_default(
 #    include <uv.h>
 
 struct uv_host_resolver {
-    /* This needs to be the first element in the struct so the pointers are safely castable */
+    /* This needs to be the first element in the struct so the pointers are safely castable.
+     * This means this object may be passed to any of the default resolver's vtable functions safely, including destroy.
+     */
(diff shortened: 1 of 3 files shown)


source/host_resolver.c
```diff
@@ -723,3 +723,95 @@ int aws_host_resolver_init_default(
...
+    uv_async_init(uv_host_resolver->uv_loop, &request->async, NULL);
+
+    return default_resolve_host(resolver, host_name, uv_host_resolved, config, request);
+}
+
+static struct aws_host_resolver_vtable s_uv_vtable = {
+    .purge_cache = resolver_purge_cache,
+    .resolve_host = uv_resolve_host,
+    .record_connection_failure = resolver_record_connection_failure,
+    .destroy = resolver_destroy,
```
justinboswell: Is this actually going to work? resolver_destroy doesn't delete the entire resolver, it only deletes the inner default resolver.

ColdenCullen: OH NO good catch

ColdenCullen: ACTUALLY yeah it should work, because the only extra thing on uv_host_resolver is the uv_loop ptr, and that doesn't need any cleanup.

PR #92, opened 2019-02-14
Fb logging

bretambrose:
*Description of changes:*

Initial implementation of default log system.

Contains:
 * User macros for logging
 * Simple, non-thread-safe controls for global logger
 * Logger interface and base structure
 * Pipeline logger that breaks logging into three components: data formatting, data passing, and data output
 * Default implementations for each pipeline component, and an init function for the overall default logger consisting of the default formatter ([LogLevel] [Timestamp] [thread_id] - <User data>), a locking background thread based channel, and a file-based output destination

Open Areas:
 * log subject design and implementation (type, format, filtering, grouping/aggregation, etc...)
 * improve file writer initialization (move to options structure rather than individual parameters)
 * binary support (necessary?)
 * performance, synchronization costs, file write aggregation, etc...


include/aws/io/logging.h
```diff
@@ -0,0 +1,175 @@
...
+
+/**
+ * The base formatted logging macro that all other formatted logging macros resolve to.
+ * Checks for a logger and filters based on log level.
+ *
+ */
+#define AWS_LOGF_RAW(log_level, subject, ...)                                                                   \
+{                                                                                                               \
+    struct aws_logger *logger = aws_logging_get();                                                              \
+    if (logger != NULL && log_level > 0 && (logger->vtable->get_log_level_fn)(logger, subject) >= log_level) {  \
```
justinboswell: If you re-order this so that the logger fetch/assignment is in the if, you can check log_level > 0 first and early out before fetching the logger.

include/aws/io/logging.h
```diff
@@ -0,0 +1,175 @@
...
+        (logger->vtable->log_fn)(logger, log_level, subject, __VA_ARGS__);                                      \
+    }                                                                                                           \
+}
+
+#define AWS_LOGF(log_level, ...) AWS_LOGF_RAW(log_level, AWS_LOG_SUBJECT_NONE, __VA_ARGS__)
```
justinboswell: This is going to be a pain to upgrade from when subject becomes mandatory, unless we create another macro which requires subject, and let this one always go to NONE. Should we just start requiring subject now?

source/log_channel.c
```diff
@@ -0,0 +1,260 @@
...
+    aws_mutex_lock(&impl->sync);
+    (*channel->writer->vtable->write_fn)(channel->writer, log_line);
+    aws_mutex_unlock(&impl->sync);
+
+    /*
+     * send is considered a transfer of ownership.  write is not a transfer of ownership.
+     * So it's always the channel's responsibility to clean up all log lines that enter
+     * it as soon as they are no longer needed.
+     */
+    aws_string_destroy(log_line);
```
justinboswell: I guess as a default this is fine, but we'll probably want to re-use this bad boy, or have 1 per thread if we use this impl, to avoid actually freeing them

source/log_channel.c
```diff
@@ -0,0 +1,260 @@
...
+            }
+
+            (*channel->writer->vtable->write_fn)(channel->writer, log_line);
+
+            /*
+             * send is considered a transfer of ownership.  write is not a transfer of ownership.
+             * So it's always the channel's responsibility to clean up all log lines that enter
+             * it as soon as they are no longer needed.
+             */
+            aws_string_destroy(log_line);
```
justinboswell: These should definitely be pooled and pre-allocated to a decent capacity. Otherwise the churn will be magnificent, though at least the runtime of it is eaten by the logger thread.

source/log_channel.c
```diff
@@ -0,0 +1,260 @@
...
+        goto cleanup_background_thread_init_fail;
+    }
+
+    channel->vtable = &s_background_channel_vtable;
+    channel->allocator = allocator;
+    channel->impl = impl;
+    channel->writer = writer;
+
+    /*
+     * Logging thread should need very little stack, but let's defer this to later
```
justinboswell: Add a SIM ticket to follow up on this, or we'll forget for sure :)

source/log_formatter.c
```diff
@@ -0,0 +1,199 @@
...
+#include <stdarg.h>
+
+/*
+ * Default formatter implementation
+ */
+
+// (max) strlen of "[<LogLevel>]"
+#define LOG_LEVEL_PREFIX_PADDING 7
+
+// (max) strlen of "[<ThreadId>]"
```
justinboswell: this is threadid + everything before it right?

source/log_formatter.c
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    int result = AWS_OP_ERR;
+
+    /*
+     * Calculate how much room we'll need to build the full log line.
+     * You cannot consume a va_list twice, so we have to copy it.
+     */
+    va_list tmp_args;
+    va_copy(tmp_args, args);
```
justinboswell: This seems pretty runtime wasteful, why not just allocate a decent chunk (like 1K) and log into that, allowing vsnprinf() to truncate?

source/log_formatter.c
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    current_index += log_level_length;
+
+    /*
+     * Add the timestamp.  To avoid copies and allocations, do some byte buffer tomfoolery.
+     *
+     * First, make a byte_buf that points to the current position in the output string
+     */
+    struct aws_byte_buf timestamp_buffer;
```
justinboswell: This is devilry, I like it though.

source/log_formatter.c
```diff
@@ -0,0 +1,199 @@
...
+        goto cleanup;
+    }
+
+    result = AWS_OP_SUCCESS;
+
+    *(struct aws_allocator **) (&raw_string->allocator) = formatter->allocator;
+    *(size_t *) (&raw_string->len) = current_index + written_count;
+
+    *formatted_output = raw_string;
+
```
justinboswell: Could just return AWS_OP_SUCCESS here and not have to worry about result at all. Would remove the result check in cleanup:

source/log_writer.c
```diff
@@ -0,0 +1,214 @@
...
+
+int aws_file_log_writer_init(
+        struct aws_log_writer *writer,
+        struct aws_allocator *allocator,
+        const char *file_name) {
+    return s_aws_file_writer_init_internal(writer, allocator, file_name, &s_file_writer_vtable);
+}
+
+int aws_log_writer_cleanup(struct aws_log_writer *writer) {
+    return (*writer->vtable->cleanup_fn)(writer);
```
justinboswell: Probably want to assert(writer->vtable->cleanup_fn != NULL) here

include/aws/io/pipeline_logger.h
```diff
@@ -0,0 +1,88 @@
...
+    struct aws_allocator *allocator;
+    enum aws_log_level level;
+};
+
+/*
+ * Initializes a pipeline logger from components that have already been initialized.  This is not an ownership transfer.
+ * After the pipeline logger is cleaned up, the components will have to manually be cleaned up by the user.
+ */
+AWS_IO_API
+int aws_pipeline_logger_init_from_unowned_components(
```
justinboswell: nitpick: init_from_external()? init_custom()? As explicit as this is, it reads strangely to me

tests/logging/log_writer_test.c
```diff
@@ -0,0 +1,181 @@
...
+#   include <sys/file.h>
+#endif // WIN32
+
+#define TEST_WRITER_MAX_BUFFER_SIZE 4096
+
+/*
+ * We need to add some filesystem support in common.  Hard-code for now.
+ */
+static const char *s_test_file_name =
+#ifdef WIN32
```
justinboswell: mkstemp() or _mktemp_s()? Should be fine for a test.

include/aws/io/logging.h
bretambrose: Good point, in fact, I should turn the log_level >0 check into an initial assert.

include/aws/io/logging.h
bretambrose: That was the original plan, but I can write up some proposals for subject and if we can get agreement, we can slip it in now and avoid the need for multiple macro sets.

source/log_channel.c
bretambrose: Re: this comment as well as the next

My current thought is to use a custom allocator based on a ringlike memory buffer (per calling thread) for the formatted strings.  By making it per thread, the allocate and frees from each pool stay exactly in order so we could get away with a couple of pointer offsets as all the state we need to allocate/free, with allocation and free being lockless against a couple of atomic values, avoiding the need for a much more complex free list/malloc implementation.  As long as the background thread kept up with the strings being thrown at it, allocation and freeing would be almost completely trivial (tls access + a few atomic integer operations).

We'd need to handle the case when the ring buffer didn't have enough room or the request was too big for it. The easiest thing to do would be just go to the heap.  If we make the ring buffer size tunable, then configuration would let you optimize things to where you "almost always" end up using the ring buffer in practice.

So allocating a string would fall back to the heap when the ring buffer can't handle the request, and freeing would check if the pointer is in the ring buffer and use the heap (well the formatter's allocator) if it's not.

source/log_channel.c
bretambrose: In addition to the above comment, one other thing is that at the very least, we need a simple baseline implementation against which we can do performance comparisons for all the improvements we end up making (like the previous comment, for example).  

I may change the naming convention from "default" to "standard" and then add an options struct where we throw in switches for all of the optimizations.

source/log_channel.c
bretambrose: Done

source/log_formatter.c
bretambrose: Will do

source/log_writer.c
bretambrose: What should be the criteria for determining when/if we check vtable pointers for non-null?  always?

tests/logging/log_writer_test.c
bretambrose: will switch

include/aws/io/pipeline_logger.h
```diff
@@ -0,0 +1,88 @@
+
+#ifndef AWS_IO_PIPELINE_LOGGER_H
+#define AWS_IO_PIPELINE_LOGGER_H
```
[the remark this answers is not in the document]
graebm: Related, and super debateable, would it be worth merging the writer/formatter/pipeline/channel stuff into one header like `logging_components.h`? Hard to know how to balance the cognitive load of 4 header files users shouldn't care about unless they're building a logger, vs the load of 1 header file with 4 different "classes" defined in it.

source/log_channel.c
justinboswell: That sounds good, since the string contains an allocator*, it could be re-routed. Sounds good to me.

source/log_writer.c
justinboswell: Usually on the surface level API functions, e.g. functions called by the user, as high up as we can.

include/aws/io/io.h
```diff
@@ -148,6 +148,12 @@ extern "C" {
 AWS_IO_API
 void aws_io_load_error_strings(void);
 
+AWS_IO_API
+int aws_io_translate_and_raise_file_open_error(int error_no);
```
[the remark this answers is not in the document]
justinboswell: :+1:

include/aws/io/log_writer.h
```diff
@@ -0,0 +1,75 @@
...
+ * default implementations that go to stdout, stderr, and a specified file.
+ */
+struct aws_log_writer;
+
+typedef int (*aws_log_writer_write_fn)(struct aws_log_writer *writer, const struct aws_string *output);
+typedef int (*aws_log_writer_cleanup_fn)(struct aws_log_writer *writer);
+
+struct aws_log_writer_vtable {
+    aws_log_writer_write_fn write_fn;
+    aws_log_writer_cleanup_fn cleanup_fn;
```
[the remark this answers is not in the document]
justinboswell: :+1:

include/aws/io/log_formatter.h
```diff
@@ -0,0 +1,75 @@
...
+    void *impl;
+};
+
+/*
+ * Initializes the default log formatter which outputs lines in the format:
+ *
+ *   [<LogLevel>] [<Timestamp>] [<ThreadId>] - <User content>\n
+ */
+AWS_IO_API
+int aws_default_log_formatter_init(struct aws_log_formatter *formatter, struct aws_allocator *allocator, enum aws_date_format date_format);
```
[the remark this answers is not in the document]
justinboswell: :+1:

include/aws/io/io.h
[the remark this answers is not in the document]
bretambrose: We might not want to export it to users but I can see it being potentially useful for downstream libraries, so my feeling would be to make it public somehow.

source/log_channel.c
```diff
@@ -0,0 +1,260 @@
...
+     */
+    struct aws_thread_options thread_options;
+    thread_options.stack_size = 0;
```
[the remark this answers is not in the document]
justinboswell: :+1:

include/aws/io/log_formatter.h
```diff
@@ -0,0 +1,75 @@
...
+ * permissions and limitations under the License.
+ */
+
+
+#include <aws/io/io.h>
+
+#include <aws/common/date_time.h>
+#include <aws/io/logging.h>
+
+#include <stdio.h>
```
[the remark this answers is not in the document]
bretambrose: Needed for va_list actually.

include/aws/io/pipeline_logger.h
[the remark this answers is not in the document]
bretambrose: Per standup discussion, I've folded pipeline_logger.h/.c into logging.h.  Also renamed it.  All you need to do is include logging.h to use the standard file-based logger.

tests/logging/log_writer_test.c
bretambrose: Putting off in favor of long-term platform-specific file support

include/aws/io/io.h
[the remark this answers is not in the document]
bretambrose: Moved to file_utils.h

source/log_formatter.c
bretambrose: That's a good candidate for performance optimization but for now I'd like to not do any clipping yet and see how things fall out when actually measured.

source/log_formatter.c
bretambrose: It's just the individual thread-id component substring, so based on max of uint64_t's string rep.

include/aws/io/logging.h
```diff
@@ -0,0 +1,175 @@
...
+
+/**
+ * Log subject is a to-be-defined key, indendent of level, that will allow for additional filtering options.
+ *
+ * The general idea is to support the logging of related functionality and allow for level control with a
+ * finer-grained approach.
+ *
+ * For example, enable TRACE logging for mqtt-related log statements, but only WARN logging everywhere else.
+ */
+typedef uint64_t aws_log_subject_t;
```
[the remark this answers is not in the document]
bretambrose: Subject has been updated to work similarly to aws_error.  Subject is not currently logged in the default formatter.

include/aws/io/log_channel.h
```diff
@@ -0,0 +1,73 @@
...
+    void *impl;
+};
+
+/*
+ * Simple channel that results in log lines being written in the same thread they were generated in.
+ *
+ * The passed in log writer is not an ownership transfer.  The log channel does not cleanup the writer.
+ */
+AWS_IO_API
+int aws_foreground_log_channel_init(struct aws_log_channel *channel, struct aws_allocator *allocator, struct aws_log_writer *writer);
```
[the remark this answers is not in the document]
bretambrose: Fixed

include/aws/io/logging.h
```diff
@@ -0,0 +1,175 @@
...
+    void *p_impl;
+};
+
+AWS_EXTERN_C_BEGIN
+
+/**
+ * Sets the aws logger used globally across the process.  Not thread-safe.  Must only be called once.
+ */
+AWS_COMMON_API
+void aws_logging_set(struct aws_logger *logger);
```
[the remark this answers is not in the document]
bretambrose: Updated

include/aws/io/pipeline_logger.h
```diff
@@ -0,0 +1,88 @@
...
+    struct aws_allocator *allocator,
+    enum aws_log_level level,
+    const char *file_name);
+
+/*
+ * Initializes a pipeline logger that is built from the default formatter, a background thread-based channel, and
+ * a writer that outputs to stdout.
+ */
+AWS_IO_API
+int aws_pipeline_logger_stdout_writer_init(
```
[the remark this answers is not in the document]
bretambrose: Done

include/aws/io/log_channel.h
```diff
@@ -0,0 +1,73 @@
...
+struct aws_string;
+struct aws_log_writer;
+
+/*
+ * Log channel interface and default implementations
+ *
+ * A log channel is an abstraction for the transfer of formatted log data between a source (formatter)
+ * and a sink (writer).
+ */
+struct aws_log_channel;
```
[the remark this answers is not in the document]
justinboswell: aws_log_pipeline?

source/file_utils.c
```diff
@@ -0,0 +1,90 @@
...
+ * permissions and limitations under the License.
+ */
+
+#include <aws/io/file_utils.h>
+
+#include <errno.h>
+#include <stdio.h>
+
+int aws_byte_buf_init_from_file(struct aws_byte_buf *out_buf, struct aws_allocator *alloc, const char *filename) {
+#ifdef _MSC_VER
```
justinboswell: Could we just do this at file scope? It seems weird to do it inside a function

bretambrose: Sure, will update

bretambrose: As far as channel vs. other names goes, I know everyone would prefer to not have an overlap with the existing aws_channel, but I feel vocabulary-wise, channel is a much more accurate description (that is in sync with its recent usage within programming languages ala Rust/Go) than anything else proposed.  Pipeline feels like a mismatch given that's how we're describing the sequence of components that make up the default logger.

PR #94, opened 2019-02-14
Attempt to fix the pathing compat with cmake.

JonathanHenson:


CMakeLists.txt
```diff
@@ -2,6 +2,15 @@
 cmake_minimum_required(VERSION 3.1)
 project(aws-c-io C)
 
+if (DEFINED CMAKE_PREFIX_PATH)
```
justinboswell: If we must do this, use file(TO_CMAKE_PATH) for this

PR #100, opened 2019-03-05
Added format checking to log calls

bretambrose:

### DIFF
--- a/source/windows/iocp/iocp_event_loop.c
+++ b/source/windows/iocp/iocp_event_loop.c
@@ -178,7 +178,7 @@ struct aws_event_loop *aws_event_loop_new_system(struct aws_allocator *alloc, aw
         return NULL;
     }
 
-    AWS_LOGF_INFO(AWS_LS_IO_EVENT_LOOP, "id=%p: Initializing IO Completion Port", event_loop);
+    AWS_LOGF_INFO(AWS_LS_IO_EVENT_LOOP, "id=%p: Initializing IO Completion Port", (void *)event_loop);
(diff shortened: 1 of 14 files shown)


source/windows/iocp/iocp_event_loop.c
```diff
@@ -200,7 +200,7 @@ struct aws_event_loop *aws_event_loop_new_system(struct aws_allocator *alloc, aw
         AWS_LOGF_FATAL(
             AWS_LS_IO_EVENT_LOOP,
             "id=%p: CreateIOCompletionPort failed with error %d",
-            event_loop,
+            (void *)event_loop,
             (int)GetLastError());
```
ColdenCullen: Instead of this cast, you can use format specifier `%ld` (`%lu` if unsigned).

source/windows/iocp/iocp_event_loop.c
```diff
@@ -414,8 +416,8 @@ static void s_schedule_task_common(struct aws_event_loop *event_loop, struct aws
         AWS_LOGF_TRACE(
             AWS_LS_IO_EVENT_LOOP,
             "id=%p: scheduling task %p in-thread for timestamp %llu",
-            event_loop,
-            task,
+            (void *)event_loop,
+            (void *)task,
             (unsigned long long)run_at_nanos);
```
ColdenCullen: Instead of this cast, you can use format specifier `%zu` for `size_t` and `%zd` for `ssize_t`.

bretambrose: zu and zd aren't supported on sufficiently old versions of msvc

PR #102, opened 2019-03-07
libuv loop 2.0

ColdenCullen:
Improvements include:
* All subscriptions through the AWS Event Loop API are now edge triggered
* libuv v0.10 and v1.x are both supported
* One of `AWS_USE_IO_COMPLETION_PORTS`, `AWS_USE_EPOLL`, or `AWS_USE_KQUEUE` is always defined.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,6 +63,8 @@ if (WIN32)
                 "source/windows/iocp/*.c"
                 )
          list(APPEND AWS_IO_OS_SRC ${AWS_IO_IOCP_SRC})
+
+         set(EVENT_LOOP_DEFINE "IO_COMPLETION_PORTS")
     endif ()
 
     if (MSVC)
@@ -76,8 +78,6 @@ if (WIN32)
     #Also note, you don't get a choice on TLS implementation for Windows.
     set(PLATFORM_LIBS Secur32 Crypt32)
     set(TLS_STACK_DETERMINED ON)
-
-    set(EVENT_LOOP_DEFINE "IO_COMPLETION_PORTS")
 
 elseif (CMAKE_SYSTEM_NAME STREQUAL "Linux")
     file(GLOB AWS_IO_OS_HEADERS
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,6 +77,8 @@ if (WIN32)
     set(PLATFORM_LIBS Secur32 Crypt32)
     set(TLS_STACK_DETERMINED ON)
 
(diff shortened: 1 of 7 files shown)


source/libuv/uv_event_loop.c
```diff
@@ -145,33 +167,94 @@ static uint64_t s_timestamp_to_uv_millis(struct aws_event_loop *event_loop, cons
...
+#if UV_VERSION_MAJOR == 0
+    /* uv won't let us tell it we care about remote hangups, so we have to trick it into thinking it wants it */
+    handle_data->poll.io_watcher.pevents |= POLLRDHUP;
+#else
+    /* Can't use RDHUP trick here, since uv will turn it off in uv__io_stop */
+    handle_data->uv_events |= UV_DISCONNECT;
+#endif
+
+    /* Start listening for events */
+    uv_poll_start(&handle_data->poll, handle_data->uv_events, (uv_poll_cb)0xBAADF00D);
```
justinboswell: Is this value special to libuv, or just obvious? Would be good to make it clearer.

source/libuv/uv_event_loop.c
```diff
@@ -744,17 +834,28 @@ struct aws_event_loop *aws_event_loop_new_libuv(struct aws_allocator *alloc, aws
     struct libuv_loop *impl = NULL;
     struct libuv_owned *owned = NULL;
     uv_loop_t *uv_loop = NULL;
+
+#if UV_VERSION_MAJOR == 0
+    size_t alloc_count = 3;
```
justinboswell: const?

source/libuv/uv_event_loop.c
```diff
@@ -216,7 +216,7 @@ static void s_do_subscribe(struct libuv_loop *impl, struct handle_data *handle_d
     handle_data->uv_events |= UV_DISCONNECT;
 #endif
 
-    /* Start listening for events */
+    /* Start listening for events. The callback should never actually be called, so we use a poison. */
```
justinboswell: Could we not just route it to a callback of our own that asserts fatally? At least then we'd understand the callstack?

CMakeLists.txt
```diff
@@ -77,6 +77,8 @@ if (WIN32)
     set(PLATFORM_LIBS Secur32 Crypt32)
     set(TLS_STACK_DETERMINED ON)
 
+    set(EVENT_LOOP_DEFINE "IO_COMPLETION_PORTS")
```
graebm: You're always setting this whether or not the USE_IO_COMPLETION_PORT options is ON

PR #103, opened 2019-03-07
URI Implementation

JonathanHenson:
There's more we'll come back and add when we need it (which we will). 

This does not include URL encoding, or canonicalization, but it IS everything we need for aws-c-http.

Also, I noticed we weren't using AWS_C_EXTERN_BEGIN etc... everywhere. Went ahead and updated that too.


source/uri.c
```diff
@@ -0,0 +1,380 @@
...
+
+    aws_byte_cursor_advance(str, parser->uri->scheme.len);
+
+    if (str->len < 3 || str->ptr[0] != ':' || str->ptr[1] != '/' || str->ptr[2] != '/') {
+        aws_raise_error(AWS_ERROR_MALFORMED_INPUT_STRING);
+        parser->state = ERROR;
+        return;
+    }
+
+    aws_byte_cursor_advance(str, 3);
```
justinboswell: /* advance past the "://" */

include/aws/io/uri.h
```diff
@@ -0,0 +1,122 @@
...
+/**
+ * key/value pairs for a query string.
+ */
+struct aws_uri_params {
+    struct aws_byte_cursor key;
+    struct aws_byte_cursor value;
+};
+
+/**
+ * Arguments for building a URI instance. All membmers must
```
ColdenCullen: `membmers`

include/aws/io/uri.h
```diff
@@ -0,0 +1,122 @@
...
+ */
+AWS_IO_API int aws_uri_init(struct aws_uri *uri, struct aws_allocator *allocator, struct aws_uri_builder_args *args);
+AWS_IO_API void aws_uri_clean_up(struct aws_uri *uri);
+
+/**
+ * Returns the scheme portion of the uri (e.g. http, https, ftp, ftps, etc...). If the scheme was not present
+ * in the uri, the returned value will be empty. It is the users job to determine the appropriate defaults
+ * if this field is empty, based on protocol, port, etc...
+ */
+AWS_IO_API const struct aws_byte_cursor *aws_uri_scheme(const struct aws_uri *uri);
```
ColdenCullen: Why have these accessors at all?

include/aws/io/uri.h
```diff
@@ -0,0 +1,122 @@
...
+ */
+struct aws_uri {
+    struct aws_allocator *allocator;
+    struct aws_byte_buf uri_str;
+    struct aws_byte_cursor scheme;
+    struct aws_byte_cursor authority;
+    struct aws_byte_cursor path;
+    struct aws_byte_cursor query_string;
+    struct aws_byte_cursor request_uri;
+    struct aws_byte_cursor host_name;
```
ColdenCullen: The order of these fields seems weird. Isn't `host_name` right after scheme?

tests/uri_test.c
```diff
@@ -0,0 +1,410 @@
...
+        "/path/to/resource?test1=value1&test%20space=value%20space&test2=value2&test2=value3");
+    ASSERT_BIN_ARRAYS_EQUALS(
+        expected_request_uri.ptr, expected_request_uri.len, built_uri.request_uri.ptr, built_uri.request_uri.len);
+
+    aws_uri_clean_up(&uri);
+    aws_uri_clean_up(&built_uri);
+    return AWS_OP_SUCCESS;
+}
+
+AWS_TEST_CASE(uri_builder, s_test_uri_builder);
```
justinboswell: What about a test of something that's just scheme and authority, e.g. http://somehost or chrome://settings

source/uri.c
```diff
@@ -0,0 +1,380 @@
...
+        return;
+    }
+
+    /* make sure we didn't just pick up the port by mistake */
+    if ((size_t)(location_of_colon - str->ptr) < str->len && *(location_of_colon + 1) != '/') {
+        parser->state = ON_AUTHORITY;
+        return;
+    }
+
+    parser->uri->scheme.ptr = str->ptr, parser->uri->scheme.len = location_of_colon - str->ptr;
```
ColdenCullen: NONONONONONONONO

Why you use the comma operator? Make this 2 lines please.

source/uri.c
```diff
@@ -0,0 +1,380 @@
...
+
+    aws_byte_cursor_advance(str, parser->uri->scheme.len);
```
ColdenCullen: You could rewrite this as:
```c
const size_t scheme_len = location_of_colon - str->ptr;
parser->uri->scheme = aws_byte_cursor_advance(str, scheme_len); /* or just location_of_colon - str->ptr for second arg */
```

source/uri.c
```diff
@@ -0,0 +1,380 @@
...
+        end = location_of_slash;
+    } else if (location_of_qmark) {
+        parser->state = ON_QUERY_STRING;
+        end = location_of_qmark;
+    }
+
+    parser->uri->authority.ptr = str->ptr;
+    parser->uri->authority.len = end - str->ptr;
+
+    aws_byte_cursor_advance(str, parser->uri->authority.len);
```
ColdenCullen: Same here, `parser->uri->authority = aws_byte_cursor_advance(str, end - str->ptr);`

include/aws/io/uri.h
```diff
@@ -0,0 +1,122 @@
...
+#include <aws/common/byte_buf.h>
+
+#include <aws/io/exports.h>
+
+/**
+ * Data representing a URI. uri_str is always allocated and filled in.
+ * The other portions are merely storing offsets into uri_str.
+ */
+struct aws_uri {
+    struct aws_allocator *allocator;
```
graebm: worth having a `size_t self_size;` member so we could add additional fields in the future (ex: fragment, scheme enum) and be binary compatible?

hey man, binary compatibility was your idea

tests/uri_test.c
```diff
@@ -0,0 +1,410 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+#include <aws/io/uri.h>
+#include <aws/testing/aws_test_harness.h>
+
+static int s_test_uri_full_parse(struct aws_allocator *allocator, void *ctx) {
+    (void)ctx;
+    const char *str_uri =
+        "https://www.test.com:8443/path/to/resource?test1=value1&test%20space=value%20space&test2=value2&test2=value3";
```
ColdenCullen: why not amazon.com :stuck_out_tongue:

include/aws/io/uri.h
```diff
@@ -0,0 +1,122 @@
...
+
+/**
+ * Returns the path portion of the uri. If the original value was empty, this value will be "/".
+ */
+AWS_IO_API const struct aws_byte_cursor *aws_uri_path(const struct aws_uri *uri);
+
+/**
+ * Returns the query string portion of the uri, minus the '?'. If not present, this value will be empty.
+ */
+AWS_IO_API const struct aws_byte_cursor *aws_uri_query_string(const struct aws_uri *uri);
```
graebm: [trivial] `aws_uri_query` ? I mean, they're all strings.

source/uri.c
JonathanHenson: It was an accident left over from a declared initializer removal.... chill

include/aws/io/uri.h
JonathanHenson: Cause I don’t want users hitting the strict directly (long term at least)

include/aws/io/uri.h
JonathanHenson: Host name and port aren’t part of the uri, they’re offsets into authority. I can move them directly after that though.

include/aws/io/uri.h
```diff
@@ -0,0 +1,122 @@
...
+/**
+ * Returns the request portion of the uri (i.e., the thing you send accross the wire).
+ */
+AWS_IO_API const struct aws_byte_cursor *aws_uri_request_uri(const struct aws_uri *uri);
+
+/**
+ * Parses query string and stores the parameters in 'params'. Returns AWS_OP_SUCCESS on success and
+ * AWS_OP_ERR on failure.
+ */
+AWS_IO_API int aws_uri_query_string_params(const struct aws_uri *uri, struct aws_array_list *params);
```
graebm: I would name the 2nd arg`out_params`.

Also, mention to user whether array_list should be initialized or not before passing in, and mention whether or not it needs to be cleaned up after.

Why not just return a const ref to the internal datastructure? That would avoid a lot of complexity if user just wants to read values.

or just have `size_t aws_uri_query_param_count(uri)` and `param* aws_uri_query_param(uri, i)`

include/aws/io/uri.h
```diff
@@ -0,0 +1,122 @@
...
+    struct aws_byte_cursor query_string;
+    struct aws_byte_cursor request_uri;
+    struct aws_byte_cursor host_name;
+    uint16_t port;
+};
+
+/**
+ * key/value pairs for a query string.
+ */
+struct aws_uri_params {
```
graebm: [URI RFC](https://tools.ietf.org/html/rfc3986#section-3.4) says
> query components  are **often** used to carry identifying information in the form of  "key=value" pairs

Seems like query string can be something other than key=value pairs.

C# only exposes Query as a string.

If we keep this struct around, don't end the name with a plural "s"

include/aws/io/uri.h
```diff
@@ -0,0 +1,122 @@
...
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+#include <aws/common/byte_buf.h>
+
+#include <aws/io/exports.h>
```
graebm: We've been having every public header include its libname.h, instead of "exports.h"

include/aws/io/uri.h
```diff
@@ -0,0 +1,122 @@
...
+/**
+ * Returns the scheme portion of the uri (e.g. http, https, ftp, ftps, etc...). If the scheme was not present
+ * in the uri, the returned value will be empty. It is the users job to determine the appropriate defaults
+ * if this field is empty, based on protocol, port, etc...
+ */
+AWS_IO_API const struct aws_byte_cursor *aws_uri_scheme(const struct aws_uri *uri);
+
+/**
+ * Returns the authority portion of the uri (host[:port]). If it was not present, this was a request uri. In that
+ * case, the value will be empty.
```
graebm: stop using term "request uri"

include/aws/io/uri.h
justinboswell: I might be crazy, but I seem to recall seeing parameters that were just keys, like ?a&b=c&d&e=f

source/uri.c
```diff
@@ -0,0 +1,380 @@
...
+        parser->uri->path.ptr = str->ptr;
+        parser->uri->path.len = str->len;
+        parser->state = FINISHED;
+        aws_byte_cursor_advance(str, parser->uri->path.len);
+        return;
+    }
+
+    if (!str->len) {
+        parser->state = ERROR;
+        aws_raise_error(AWS_ERROR_MALFORMED_INPUT_STRING);
```
bretambrose: Isn't this more an internal state machine error rather than input string?

tests/uri_test.c
```diff
@@ -0,0 +1,410 @@
+/*
```
bretambrose: Do we/should we validate/test the character space (encoding) of the input buffer anywhere?

source/uri.c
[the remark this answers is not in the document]
JonathanHenson: yeah, stupid mistake that clang-format made look intentional

source/uri.c
```diff
@@ -0,0 +1,380 @@
...
+
+    while (parser.state < FINISHED) {
+        s_states[parser.state](&parser, &uri_cur);
+    }
+
+    if (parser.state == FINISHED) {
+        return AWS_OP_SUCCESS;
+    }
+
+    aws_byte_buf_clean_up(&uri->uri_str);
```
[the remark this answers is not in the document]
JonathanHenson: good catch.

include/aws/io/uri.h
justinboswell: Actually, if that's true, we should add a test for that too

tests/uri_test.c
JonathanHenson: you found a bug!

include/aws/io/uri.h
```diff
@@ -0,0 +1,122 @@
...
+/**
+ * Arguments for building a URI instance. All membmers must
+ * be initialized before passing them to aws_uri_init().
+ */
+struct aws_uri_builder_args {
+    struct aws_byte_cursor scheme;
+    struct aws_byte_cursor path;
+    struct aws_byte_cursor host_name;
+    uint16_t port;
+    struct aws_array_list query_params;
```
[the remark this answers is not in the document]
JonathanHenson: yeah, I don't think they will likely not be known ahead of time (envision building a generated aws client on top of it).

include/aws/io/uri.h
JonathanHenson: I aghree with brett on this one.

include/aws/io/uri.h
JonathanHenson: I can certainly change the param name and comment updates, but there's not an internal struct for this, it's parsed on the fly.

include/aws/io/uri.h
JonathanHenson: For practical use cases, C# is being worthless on this one. It's quite common to need individual args. For one, the canonicalization process needs them like this so they can be properly sorted.

JonathanHenson: I can add support for JUST A KEY though.

source/uri.c
JonathanHenson: we only get to a bad state if the input string was malformed.

tests/uri_test.c
JonathanHenson: We test that something url encoded still parses properly.

bretambrose: But should we reject input that isn't encoded properly?

include/aws/io/uri.h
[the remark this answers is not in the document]
graebm: I think a generated aws client would use its own class to gather these parameters. Then it would fill out a _builder on the stack in the same function that creates the aws_uri.

But this is an argument over a trivial point, so dropping it

include/aws/io/uri.h
```diff
@@ -0,0 +1,128 @@
...
+/**
+ * Arguments for building a URI instance. All members must
+ * be initialized before passing them to aws_uri_init().
+ */
+struct aws_uri_builder_options {
+    struct aws_byte_cursor scheme;
+    struct aws_byte_cursor path;
+    struct aws_byte_cursor host_name;
+    uint16_t port;
+    struct aws_array_list query_params;
```
graebm: Since query strings don't HAVE to be key-value pairs, why not offer an alternative `aws_byte_cursor query_string` and the user can provide one or the other.

PR #112, opened 2019-03-20
Fixed race condition in host resolver, fixed ref count bug in channel…

JonathanHenson:
… bootstrap, fixed tcp keep alive to be more responsive.

### DIFF
--- a/source/channel_bootstrap.c
+++ b/source/channel_bootstrap.c
@@ -580,7 +580,7 @@ static void s_on_host_resolved(
             s_connection_args_acquire(client_connection_args);
 
             struct aws_host_address *host_address_ptr = NULL;
-            aws_array_list_get_at_ptr(host_addresses, (void *)&host_address_ptr, i);
+            aws_array_list_get_at_ptr(host_addresses, (void **)&host_address_ptr, i);
 
             struct aws_socket_endpoint connection_endpoint;
             connection_endpoint.port = client_connection_args->outgoing_port;
@@ -598,6 +598,12 @@ static void s_on_host_resolved(
 
             struct aws_socket *outgoing_socket =
                 aws_mem_acquire(client_connection_args->bootstrap->allocator, sizeof(struct aws_socket));
+
+            if (!outgoing_socket) {
+                client_connection_args->failed_count++;
+                s_connection_args_release(client_connection_args);
+                break;
+            }
 
             if (aws_socket_init(outgoing_socket, client_connection_args->bootstrap->allocator, &options)) {
                 client_connection_args->failed_count++;
@@ -629,7 +635,9 @@ static void s_on_host_resolved(
     }
 
     AWS_LOGF_ERROR(
-        AWS_LS_IO_CHANNEL_BOOTSTRAP, "id=%p: dns resolution failed.", (void *)client_connection_args->bootstrap);
+        AWS_LS_IO_CHANNEL_BOOTSTRAP,
+        "id=%p: dns resolution failed, or all socket connections to the endpoint failed.",
+        (void *)client_connection_args->bootstrap);
     client_connection_args->setup_callback(
         client_connection_args->bootstrap, aws_last_error(), NULL, client_connection_args->user_data);
     s_connection_args_release(client_connection_args);
--- a/source/channel_bootstrap.c
+++ b/source/channel_bootstrap.c
@@ -131,6 +131,7 @@ struct aws_client_bootstrap *aws_client_bootstrap_new(
         (void *)bootstrap,
         (void *)el_group);
 
+    AWS_ZERO_STRUCT(*bootstrap);
     bootstrap->allocator = allocator;
     bootstrap->event_loop_group = el_group;
     bootstrap->on_protocol_negotiated = NULL;
@@ -221,12 +222,15 @@ struct client_connection_args {
 
 void s_connection_args_acquire(struct client_connection_args *args) {
     assert(args);
-    args->ref_count++;
+    if (args->ref_count++ == 0) {
+        s_client_bootstrap_acquire(args->bootstrap);
+    }
 }
 
 void s_connection_args_release(struct client_connection_args *args) {
     assert(args);
     assert(args->ref_count > 0);
+
     if (--args->ref_count == 0) {
         struct aws_allocator *allocator = args->bootstrap->allocator;
         s_client_bootstrap_release(args->bootstrap);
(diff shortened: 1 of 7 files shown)


source/host_resolver.c
```diff
@@ -732,38 +732,64 @@ static int default_resolve_host(
...
-            AWS_LOGF_TRACE(
-                AWS_LS_IO_DNS,
-                "id=%p: vending address %s for host %s to caller",
-                (void *)resolver,
-                a_record->address->bytes,
-                host_entry->host_name->bytes);
+            struct aws_host_address a_record_cpy;
+            if (!aws_host_address_copy(a_record, &a_record_cpy)) {
+                aws_array_list_push_back(&callback_address_list, &a_record_cpy);
+                AWS_LOGF_TRACE(
```
justinboswell: Is it worthwhile to move this outside the lock and log the copies so we aren't logging inside a lock?

source/posix/socket.c
```diff
@@ -1089,12 +1090,24 @@ int aws_socket_set_options(struct aws_socket *socket, const struct aws_socket_op
 
         if (socket->options.keep_alive_interval_sec && socket->options.keep_alive_timeout_sec) {
             int ival_in_secs = socket->options.keep_alive_interval_sec;
-            setsockopt(socket->io_handle.data.fd, IPPROTO_TCP, TCP_KEEPIDLE, &ival_in_secs, sizeof(ival_in_secs));
+            success =
```
justinboswell: Should we warn on success != 0?

source/windows/iocp/socket.c
```diff
@@ -2203,6 +2203,18 @@ int aws_socket_set_options(struct aws_socket *socket, const struct aws_socket_op
                 &bytes_returned,
                 NULL,
                 NULL);
+
+#ifdef TCP_KEEPCNT
```
justinboswell: When is this not defined on windows?

source/channel_bootstrap.c
```diff
@@ -572,14 +576,11 @@ static void s_on_host_resolved(
...
 
-        /* retain the connection args in case an early connection fails */
-        s_connection_args_acquire(client_connection_args);
         for (size_t i = 0; i < host_addresses_len; ++i) {
-            /* retain args, need 1 ref per socket */
             s_connection_args_acquire(client_connection_args);
 
             struct aws_host_address *host_address_ptr = NULL;
-            aws_array_list_get_at(host_addresses, (void *)&host_address_ptr, i);
+            aws_array_list_get_at_ptr(host_addresses, (void *)&host_address_ptr, i);
```
justinboswell: Shouldn't this be (void**)?

source/channel_bootstrap.c
```diff
@@ -221,12 +222,15 @@ struct client_connection_args {
 
 void s_connection_args_acquire(struct client_connection_args *args) {
     assert(args);
-    args->ref_count++;
+    if (args->ref_count++ == 0) {
```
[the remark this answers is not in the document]
bretambrose: post inc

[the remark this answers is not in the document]
graebm: It's a post-increment. So the if-branch is only taken the first time the ref-count is acquired

source/host_resolver.c
```diff
@@ -732,52 +732,79 @@ static int default_resolve_host(
...
+            res(resolver, host_name, AWS_OP_SUCCESS, &callback_address_list, user_data);
+        } else {
+            res(resolver, host_name, aws_last_error(), NULL, user_data);
+            error_code = AWS_OP_ERR;
+        }
+
+        for (size_t i = 0; i < aws_array_list_length(&callback_address_list); ++i) {
+            struct aws_host_address *address_ptr = NULL;
+            aws_array_list_get_at_ptr(&callback_address_list, (void **)&address_ptr, i);
+            aws_host_address_clean_up(address_ptr);
```
graebm: [long rant] I really dislike using array_list when the size is known. If we used C-arrays we'd get code that's much easier to read and write, plus we'd get type-checking by the compiler.

I was going to say how it's too late to change this because the array_list is passed to a callback that is already part of the public API. But that brings up a good point. These changes SHOULD have broken the public API. The array_list now holds pointers, it used to hold values, but the API didn't change. **And the comments on `aws_on_host_resolved_result_fn` still talk about how the array_list holds things by value**.

So we DID break the API, but a user who took this change would not be alerted of the break by the compiler, so they'd have to chase down bugs at runtime instead, because array_lists are not type-safe, in addition to being harder to use when there was no need to be dynamic anyway.

source/host_resolver.c
JonathanHenson: probably, though there's an allocation in that copy. And this isn't really a hot path (this almost always comes from the caller's thread.

source/posix/socket.c
JonathanHenson: How about I do bit ors on it and warn on the final step that one or more failed? Theres not a whole lot that can be done to fix it, and it's not the end of the world if this fails.

source/windows/iocp/socket.c
JonathanHenson: It's available Windows 10 1703 and later

source/channel_bootstrap.c
JonathanHenson: yes.... not sure how that compiled

source/windows/iocp/socket.c
```diff
@@ -2203,6 +2203,18 @@ int aws_socket_set_options(struct aws_socket *socket, const struct aws_socket_op
...
+            if (socket->options.keep_alive_max_failed_probes) {
+                DWORD max_probes = socket->options.keep_alive_max_failed_probes;
+                success = setsockopt(
```
graebm: should we warn if not successful? nothing is done with this variable

JonathanHenson: i left it there for debugging purposes, I can add a warn statement though.

source/host_resolver.c
JonathanHenson: well we'd also have to pass number of entries, but I guess that's not the end of the world. I can do that I guess.

JonathanHenson: hmmmm, the more I think about it the more I think this isn't the case. Theres nothing guaranteeing how many records we'll return in the future. Could be all of them in the case of multi-answer. It's just an implementation detail at the moment that we know the max size in advance.

JonathanHenson: so the underlying piece that calls getaddrinfo, returns a variable length list because we don't know how many will come back. Could be 1, could be 24

JonathanHenson: If we ever want to change the load balancing behavior in channel bootstrap, we'd have to go back and convert back to array list anyways.

JonathanHenson: and it was documented as by-value, but in reality it was by reference, now it's actually by-value again. this was because we had to copy to avoid the race condition.

graebm: If you pass the array and its length to the callback, there's nothing stopping you from using an array_list in the implementation. You can pass the buffer and length to the callback.

PR #115, opened 2019-03-28
Add some directory-related functionality to file utils

bretambrose:
Moving some rudimentary directory functionality from auth into io.

### DIFF
--- a/include/aws/io/file_utils.h
+++ b/include/aws/io/file_utils.h
@@ -47,25 +47,19 @@ int aws_io_translate_and_raise_file_write_error(int error_no);
  * Returns true iff the character is a directory separator on ANY supported platform.
  */
 AWS_IO_API
-bool aws_is_directory_separator(char value);
+bool aws_is_any_directory_separator(char value);
 
 /**
  * Returns the directory separator used by the local platform
  */
 AWS_IO_API
-char aws_get_local_platform_directory_separator(void);
+char aws_get_platform_directory_separator(void);
 
 /**
  * Returns the current user's home directory.
  */
 AWS_IO_API
 struct aws_string *aws_get_home_directory(struct aws_allocator *allocator);
-
-/**
- * Returns the current user's home directory, if it is stored in the HOME environment variable
- */
-AWS_IO_API
-struct aws_string *aws_get_home_directory_environment_value(struct aws_allocator *allocator);
 
 AWS_EXTERN_C_END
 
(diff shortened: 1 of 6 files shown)


include/aws/io/file_utils.h
```diff
@@ -31,12 +31,42 @@ AWS_IO_API int aws_byte_buf_init_from_file(
...
+ * Returns the current user's home directory.
+ */
+AWS_IO_API
+struct aws_string *aws_get_home_directory(struct aws_allocator *allocator);
+
+/**
+ * Returns the current user's home directory, if it is stored in the HOME environment variable
+ */
+AWS_IO_API
+struct aws_string *aws_get_home_directory_environment_value(struct aws_allocator *allocator);
```
justinboswell: When would this be valuable to use publicly over ```aws_get_home_directory()```? Also, I think it would be closer to our usual naming if it was ```aws_get_home_directory_from_environment()```

include/aws/io/file_utils.h
```diff
@@ -31,12 +31,42 @@ AWS_IO_API int aws_byte_buf_init_from_file(
...
+ * Returns true iff the character is a directory separator on ANY supported platform.
+ */
+AWS_IO_API
+bool aws_is_directory_separator(char value);
+
+/**
+ * Returns the directory separator used by the local platform
+ */
+AWS_IO_API
+char aws_get_local_platform_directory_separator(void);
```
justinboswell: [nitpick] I don't think local is necessary in the name

include/aws/io/file_utils.h
```diff
@@ -31,12 +31,42 @@ AWS_IO_API int aws_byte_buf_init_from_file(
...
+ * Convert a c library error from reading a file into an aws error. Consider merging with above.
+ */
 AWS_IO_API
 int aws_io_translate_and_raise_file_write_error(int error_no);
 
+/**
+ * Returns true iff the character is a directory separator on ANY supported platform.
+ */
+AWS_IO_API
+bool aws_is_directory_separator(char value);
```
justinboswell: Perhaps ```aws_is_any_directory_separator()``` or something? Or have ```aws_is_platform_directory_separator()``` to denote the difference?

include/aws/io/file_utils.h
bretambrose: It's only use is to be shared between the posix and windows versions and I didn't think it warranted a private header.

include/aws/io/file_utils.h
bretambrose: Will remove

include/aws/io/file_utils.h
bretambrose: It only saves two lines, I'll remove and copy-paste

PR #116, opened 2019-03-28
Remove uv.h include hack, instead insert include path BEFORE

ColdenCullen:


CMakeLists.txt
```diff
@@ -200,7 +200,7 @@ if (USE_LIBUV)
     # Allow specific overriding of uv.h location
     if (UV_HEADER_PATH)
         message(STATUS "Using uv.h location: ${UV_HEADER_PATH}")
-        target_compile_definitions(${CMAKE_PROJECT_NAME} PUBLIC "UV_HEADER_PATH=${UV_HEADER_PATH}")
+        target_include_directories(${CMAKE_PROJECT_NAME} BEFORE PUBLIC ${UV_HEADER_PATH})
```
justinboswell: OMG we're idiots

ColdenCullen: Technically this isn't as good of a solution because global include paths could still win, but hey it's better than including `uv.h` from `~/.nvm/vXX.XX/include/node/` and `uv/unix.h` from `usr/include`, which *may* have been happening before.

PR #118, opened 2019-03-31
Better pki path default

JonathanHenson:
We now handle PKI default paths, properly on unix systems, since if a custom libcrypto build is used, it cannot reliably find them on its own.

### DIFF
--- a/include/aws/io/file_utils.h
+++ b/include/aws/io/file_utils.h
@@ -69,4 +69,4 @@ bool aws_path_exists(const char *path);
 
 AWS_EXTERN_C_END
 
-#endif // AWS_IO_FILE_UTILS_H
+#endif /* AWS_IO_FILE_UTILS_H */
--- a/include/aws/io/file_utils.h
+++ b/include/aws/io/file_utils.h
@@ -65,7 +65,7 @@ struct aws_string *aws_get_home_directory(struct aws_allocator *allocator);
  * Returns true if a file or path exists, otherwise, false.
  */
 AWS_IO_API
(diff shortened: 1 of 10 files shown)


include/aws/io/file_utils.h
```diff
@@ -61,6 +61,12 @@ char aws_get_platform_directory_separator(void);
 AWS_IO_API
 struct aws_string *aws_get_home_directory(struct aws_allocator *allocator);
 
+/**
+ * Returns true if a file or path exists, otherwise, false.
+ */
+AWS_IO_API
+bool aws_does_path_exist(const char *path);
```
graebm: `aws_path_exists()`?
would work well with any future `aws_path_XYZ()` functions we might add

source/posix/file_utils.c
```diff
@@ -34,3 +37,8 @@ struct aws_string *aws_get_home_directory(struct aws_allocator *allocator) {
 
     return NULL;
 }
+
+bool aws_does_path_exist(const char *path) {
```
graebm: we're sure we want to be using c-str instead of byte_cursor?
I can see it both ways, just want to make sure we're sure

include/aws/io/file_utils.h
JonathanHenson: yep

source/posix/file_utils.c
JonathanHenson: it would have to be null terminated, so aws_string would need to be used. At that point, I'd prefer const char * and let the user make the allocation decision.

PR #120, opened 2019-04-04
Make channel shutdown safe

JonathanHenson:
…se, but there's times where it has to be), we need to hold a reference to it when we schedule the shutdown task.

### DIFF
--- a/tests/channel_test.c
+++ b/tests/channel_test.c
@@ -502,6 +502,11 @@ static void s_channel_test_shutdown(struct aws_channel *channel, int error_code,
     aws_condition_variable_notify_one(&test_args->condition_variable);
 }
 
+static bool s_channel_test_shutdown_predicate(void *arg) {
+    struct channel_setup_test_args *test_args = (struct channel_setup_test_args *)arg;
+    return test_args->shutdown_completed;
+}
+
 enum tasks_run_id {
     TASK_NOW_OFF_THREAD,
     TASK_NOW_ON_THREAD,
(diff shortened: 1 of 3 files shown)


graebm:
Add a regression test for this

graebm:
A regression test for this could be to have a handler that calls channel_shutdown() from each of its own handler->shutdown() and handler->delete() functions.

JonathanHenson:
That wouldn’t help In this case. Channel tasks are canceled at shutdown. In this case, the channel is already shutdown and would never be canceled.

tests/channel_test.c
```diff
@@ -715,6 +715,48 @@ static int s_test_channel_cancels_pending_tasks(struct aws_allocator *allocator,
...
+    };
+
+    ASSERT_SUCCESS(aws_mutex_lock(&test_args.mutex));
+    channel = aws_channel_new(allocator, event_loop, &callbacks);
+    ASSERT_NOT_NULL(channel);
+    ASSERT_SUCCESS(aws_condition_variable_wait(&test_args.condition_variable, &test_args.mutex));
+    ASSERT_INT_EQUALS(0, test_args.error_code);
+
+    ASSERT_SUCCESS(aws_channel_shutdown(channel, AWS_ERROR_SUCCESS));
+    ASSERT_SUCCESS(aws_condition_variable_wait(&test_args.condition_variable, &test_args.mutex));
```
justinboswell: Shouldn't this wait for test_args.shutdown_completed? Or at least validate it?

PR #122, opened 2019-04-08
Event loop shutdown fixes

JonathanHenson:

### DIFF
--- a/source/bsd/kqueue_event_loop.c
+++ b/source/bsd/kqueue_event_loop.c
@@ -163,6 +163,7 @@ struct aws_event_loop *aws_event_loop_new_system(struct aws_allocator *alloc, aw
     if (!impl) {
         goto clean_up;
     }
+    /* intialize thread id to 0. It will be set when the event loop thread starts. */
     aws_atomic_init_int(&impl->thread_id, (size_t)0);
     clean_up_impl_mem = true;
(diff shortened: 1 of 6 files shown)


source/bsd/kqueue_event_loop.c
```diff
@@ -356,6 +360,7 @@ static int s_run(struct aws_event_loop *event_loop) {
         AWS_LOGF_FATAL(AWS_LS_IO_EVENT_LOOP, "id=%p: thread creation failed.", (void *)event_loop);
         goto clean_up;
     }
+    aws_atomic_store_int(&impl->thread_id, aws_thread_get_id(&impl->thread));
```
graebm: we should set the thread_id from the beginning of `s_event_thread_main()`, not from here

source/bsd/kqueue_event_loop.c
```diff
@@ -294,6 +297,7 @@ static void s_destroy(struct aws_event_loop *event_loop) {
         assert("Failed to destroy event-thread, resources have been leaked." == NULL);
         return;
     }
+    aws_atomic_store_int(&impl->thread_id, (size_t)aws_thread_current_thread_id());
```
justinboswell: I would really like to see a comment at the sites where thread ids are transitioned to explain why

PR #123, opened 2019-04-09
Host resolver is no longer owned by the channel bootstrap. Updates to…

JonathanHenson:
… make failure/shutdown more graceful from the host resolver.

### DIFF
--- a/source/channel_bootstrap.c
+++ b/source/channel_bootstrap.c
@@ -144,7 +144,6 @@ struct aws_client_bootstrap *aws_client_bootstrap_new(
     }
 
     return bootstrap;
-    ;
 }
 
 int aws_client_bootstrap_set_alpn_callback(
@@ -818,7 +817,7 @@ struct aws_server_bootstrap *aws_server_bootstrap_new(
     return bootstrap;
 }
 
-void aws_server_bootstrap_destroy(struct aws_server_bootstrap *bootstrap) {
+void aws_server_bootstrap_release(struct aws_server_bootstrap *bootstrap) {
     /* if destroy is being called, the user intends to not use the bootstrap anymore
      * so we clean up the thread local state while the event loop thread is
      * still alive */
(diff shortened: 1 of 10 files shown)


source/host_resolver.c
```diff
@@ -489,9 +496,13 @@ static void on_host_value_removed(void *value) {
...
     }
 
+    if (!aws_linked_list_empty(&host_entry->pending_resolution_callbacks)) {
+        aws_raise_error(AWS_IO_DNS_HOST_REMOVED_FROM_CACHE);
+    }
+
     while (!aws_linked_list_empty(&host_entry->pending_resolution_callbacks)) {
         struct aws_linked_list_node *resolution_callback_node =
-            aws_linked_list_front(&host_entry->pending_resolution_callbacks);
+            aws_linked_list_pop_front(&host_entry->pending_resolution_callbacks);
```
justinboswell: How did this ever work?!

source/channel_bootstrap.c
```diff
@@ -183,7 +157,7 @@ int aws_client_bootstrap_set_alpn_callback(
     return AWS_OP_SUCCESS;
 }
 
-void aws_client_bootstrap_destroy(struct aws_client_bootstrap *bootstrap) {
+void aws_client_bootstrap_release(struct aws_client_bootstrap *bootstrap) {
```
justinboswell: Does aws_server_bootstrap need this same treatment? I know it doesn't have a resolver right now, but eventually a server might want to resolve peers? Or would we delegate that to connection handlers?

source/host_resolver.c
JonathanHenson: Because the path was never being hit, due to the fact that it was impossible for host resolver to be cleaned up, because channel bootstrap wasn't cleaned up until all pending resolutions were finished.

source/channel_bootstrap.c
JonathanHenson: The only time you'd need to resolve peers (say, to connect to something else), you'd go through a client bootstrap. Can you think of a different use-case?

JonathanHenson: I did change the server to use release instead of destroy

PR #127, opened 2019-04-29
Stream

bretambrose:
* Initial input stream interface and implementations for in-memory and libc-file-io.


source/stream.c
```diff
@@ -0,0 +1,359 @@
...
+/*
+ * This is an ugly function that, in the absence of better guidance, is designed to handle all possible combinations of
+ *  aws_off_t (int32_t, int64_t) x all possible combinations of size_t (uint32_t, uint64_t).  Whether the anomalous
+ * combination of int64_t vs. uint32_t is even possible on any real platform is unknown.  If size_t ever exceeds 64 bits
+ * this function will fail badly.
+ *
+ *  Safety and invariant assumptions are sprinkled via comments.  The overall strategy is to cast up to 64 bits and
+ * perform all arithmetic there, being careful with signed vs. unsigned to prevent bad operations.
+ *
+ *  Assumption #1: aws_off_t resolves to a signed integer 64 bits or smaller
```
justinboswell: AWS_STATIC_ASSERT(sizeof(aws_off_t) >= 8)

source/stream.c
```diff
@@ -0,0 +1,359 @@
...
+ *  Assumption #2: size_t resolves to an unsigned integer 64 bits or smaller
```
justinboswell: AWS_STATIC_ASSERT(sizeof(size_t) <= 8)

source/stream.c
```diff
@@ -0,0 +1,359 @@
...
+            }
+
+            /* safe because negative offsets were turned into an error */
+            final_offset = (uint64_t)checked_offset;
+            break;
+
+        case AWS_SSB_END:
+            /*
+             * -checked_offset -- safe as long checked_offset is not INT64_MIN which was previously checked
+             * (uint64_t)(-checked_offset) -- safe because (-checked_offset) is positive (and < INT64_MAX < UINT64_MAX)
```
justinboswell: Could add asserts to prove this

source/stream.c
```diff
@@ -0,0 +1,359 @@
...
+    return AWS_OP_SUCCESS;
+}
+
+static int s_aws_input_stream_file_read(
+    struct aws_input_stream *stream,
+    struct aws_byte_buf *dest,
+    size_t *amount_read) {
+    struct aws_input_stream_file_impl *impl = stream->impl;
+
+    *amount_read = 0;
```
justinboswell: assert(amount_read) or if (amount_read)?

source/stream.c
```diff
@@ -0,0 +1,359 @@
...
+            break;
+    }
+
+    /* true because we already validated against (impl->original_cursor.len) which is <= SIZE_MAX */
+    assert(final_offset <= SIZE_MAX);
+
+    impl->current_cursor = impl->original_cursor;
+
+    /* (size_t) final_offset - safe via previous assert's reasoning */
+    aws_byte_cursor_advance(&impl->current_cursor, (size_t)final_offset);
```
graebm: not-fun fact: aws_byte_cursor_advance() fails if distance > SIZE_MAX/2

include/aws/io/stream.h
```diff
@@ -0,0 +1,108 @@
...
+struct aws_stream_status {
+    bool is_end_of_stream;
+    bool is_valid;
+};
+
+typedef int(
+    aws_input_stream_seek_fn)(struct aws_input_stream *stream, aws_off_t offset, enum aws_stream_seek_basis basis);
+typedef int(aws_input_stream_read_fn)(struct aws_input_stream *stream, struct aws_byte_buf *dest, size_t *amount_read);
+typedef int(aws_input_stream_get_status_fn)(struct aws_input_stream *stream, struct aws_stream_status *status);
+typedef int(aws_input_stream_get_length_fn)(struct aws_input_stream *stream, size_t *out_length);
```
graebm: I don't feel strongly about this, just making sure we've thought about it:

Is there any compelling reason to make length part of aws_stream_status, instead of a separate query?

source/stream.c
```diff
@@ -0,0 +1,359 @@
...
+    .seek = s_aws_input_stream_byte_cursor_seek,
+    .read = s_aws_input_stream_byte_cursor_read,
+    .get_status = s_aws_input_stream_byte_cursor_get_status,
+    .get_length = s_aws_input_stream_byte_cursor_get_length,
+    .clean_up = s_aws_input_stream_byte_cursor_clean_up};
+
+struct aws_input_stream *aws_input_stream_new_from_cursor(
+    struct aws_allocator *allocator,
+    const struct aws_byte_cursor *cursor) {
+    struct aws_input_stream *input_stream = aws_mem_acquire(allocator, sizeof(struct aws_input_stream));
```
graebm: we could do a single aws_mem_acquire with enough memory for both structs. `aws_mem_acquire_many()` exists for that purpose.

Or achieve the same thing by having the aws_input_stream_byte_cursor_impl contain the aws_input_stream

source/stream.c
```diff
@@ -0,0 +1,359 @@
...
+
+static int s_aws_input_stream_file_seek(
+    struct aws_input_stream *stream,
+    aws_off_t offset,
+    enum aws_stream_seek_basis basis) {
+    struct aws_input_stream_file_impl *impl = stream->impl;
+
+    int whence = (basis == AWS_SSB_BEGIN) ? SEEK_SET : SEEK_END;
+    if (aws_fseek(impl->file, offset, whence)) {
+        return aws_raise_error(aws_io_translate_and_raise_io_error(aws_last_error()));
```
graebm: trivial: aws_fseek is already raising the translated error

source/stream.c
```diff
@@ -0,0 +1,359 @@
...
+    size_t *amount_read) {
+    struct aws_input_stream_file_impl *impl = stream->impl;
+
+    *amount_read = 0;
+
+    size_t max_read = dest->capacity - dest->len;
+    size_t actually_read = fread(dest->buffer + dest->len, 1, max_read, impl->file);
+    if (actually_read == 0) {
+        if (ferror(impl->file)) {
+            return AWS_OP_ERR;
```
graebm: need to aws_raise_error

source/stream.c
```diff
@@ -0,0 +1,359 @@
...
+
+    struct aws_input_stream_file_impl *impl = aws_mem_acquire(allocator, sizeof(struct aws_input_stream_file_impl));
+    if (impl == NULL) {
+        goto on_error;
+    }
+
+    AWS_ZERO_STRUCT(*impl);
+    impl->file = fopen(file_name, "r");
+    if (impl->file == NULL) {
+        goto on_error;
```
graebm: need to aws_raise_error

source/windows/file_utils.c
```diff
@@ -88,3 +89,37 @@ struct aws_string *aws_get_home_directory(struct aws_allocator *allocator) {
...
+        return aws_raise_error(aws_io_translate_and_raise_io_error(errno));
+    }
+
+    return AWS_OP_SUCCESS;
+}
+
+int aws_file_get_length(FILE *file, size_t *length) {
+    int fd = _fileno(file);
+    if (fd == -1) {
+        return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
```
graebm: AWS_IO_FILE_VALIDATION_FAILURE?

include/aws/io/file_utils.h
```diff
@@ -67,6 +61,26 @@ struct aws_string *aws_get_home_directory(struct aws_allocator *allocator);
...
 
+/*
+ * Wrapper for highest-resolution platform-dependent seek implementation.
+ * Maps to:
+ *
+ *   _fseeki64() on windows
+ *   fseeko() on linux
+ */
+AWS_IO_API
+int aws_fseek(FILE *file, aws_off_t offset, int whence);
```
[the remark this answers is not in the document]
bretambrose: They're in sync constant wise but internally we translate.  I'll add the valid values to the comment.

include/aws/io/stream.h
bretambrose: length may not be a valid query so it would also need an indicator variable.  Additionally, our notion of length is a relatively constant property but the other status members are more dynamic.  So I think it's ok to be separate or eventually folded into a separate structure.

source/stream.c
bretambrose: We allocate polymorphic things in a single block elsewhere?  Haven't noticed that.

include/aws/io/file_utils.h
```diff
@@ -67,6 +61,26 @@ struct aws_string *aws_get_home_directory(struct aws_allocator *allocator);
...
+
+/*
+ * Wrapper for os-specific file length query.  We can't use fseek(END, 0)
+ * because support for it is not technically required.
+ *
+ * Unix flavors call fstat, while Windows variants use GetFileSize on a
+ * HANDLE queried from the libc FILE pointer.
+ */
+AWS_IO_API
+int aws_file_get_length(FILE *file, size_t *length);
```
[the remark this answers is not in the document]
bretambrose: Updated to int64_t since file seeking is signed.

source/stream.c
bretambrose: advance-be-gone

source/stream.c
graebm: Colden has used aws_mem_acquire_many() a few times in aws-c-mqtt

I've been making "subclasses" that contain the base type as a member. I personally feel like that's simpler to implement.

PR #133, opened 2019-05-10
Fixed dll symbol export, made sure PKI buffers were always null termi…

JonathanHenson:
…nated.

### DIFF
--- a/source/tls_channel_handler.c
+++ b/source/tls_channel_handler.c
@@ -70,13 +70,40 @@ int aws_tls_ctx_options_init_client_mtls(
     options->allocator = allocator;
     options->max_fragment_size = g_aws_channel_max_fragment_size;
 
-    if (aws_byte_buf_init_copy_from_cursor(&options->certificate, allocator, *cert)) {
-        return AWS_OP_ERR;
+    /* s2n relies on null terminated c_strings, so, we need to make sure we're properly
+     * terminated, but we don't want length to reflect the terminator because
+     * Apple and Windows will fail hard if you use a null terminator. */
+    if (cert->ptr[cert->len - 1] == 0) {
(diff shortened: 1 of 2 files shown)


source/tls_channel_handler.c
```diff
@@ -70,13 +70,40 @@ int aws_tls_ctx_options_init_client_mtls(
     options->allocator = allocator;
     options->max_fragment_size = g_aws_channel_max_fragment_size;
 
-    if (aws_byte_buf_init_copy_from_cursor(&options->certificate, allocator, *cert)) {
-        return AWS_OP_ERR;
+    /* s2n relies on null terminated c_strings, so, we need to make sure we're properly
+     * terminated, but we don't want length to reflect the terminator because
+     * Apple and Windows will fail hard if you use a null terminator. */
+    if (cert->ptr[cert->len - 1] == 0) {
```
justinboswell: Feels like it's worth putting this logic in a function, since we use it multiple times?

source/tls_channel_handler.c
```diff
@@ -93,6 +120,7 @@ int aws_tls_ctx_options_init_client_mtls_from_path(
     options->allocator = allocator;
     options->max_fragment_size = g_aws_channel_max_fragment_size;
 
+    /* this code already handles null termination properly */
```
justinboswell: This comment doesn't mean anything to the next person to read it unless you give more context.

source/tls_channel_handler.c
JonathanHenson: good call, that's much better now.

source/tls_channel_handler.c
JonathanHenson: removed, it was unnecessary anyways.

PR #139, opened 2019-06-03
Simple shared library per-platform implementation

bretambrose:
Bare-bones shared library open/close/find-symbol platform functionality


source/windows/shared_library.c
```diff
@@ -0,0 +1,76 @@
...
+
+static const char *s_null = "<NULL>";
+static const char *s_unknown_error = "<Unknown>";
+
+int aws_shared_library_init(struct aws_shared_library *library, const char *library_path) {
+    AWS_ZERO_STRUCT(*library);
+
+    library->library_handle = LoadLibrary(library_path);
+    if (library->library_handle == NULL) {
+        DWORD ec = GetLastError();
```
justinboswell: You can convert this to a string using FormatMessage()

bretambrose: I was really reluctant to start using that for windows errors.  We don't use it anywhere else in the platform-specific logic and it's doing windows-internal allocation underneath and I didn't want to be the first =/

PR #145, opened 2019-06-11
Enforce that setup_callback is always called exactly once

justinboswell:
Additionally, if setup_callback is called with an error, then shutdown will never be called.

In order to make this work, the happy eyeballs algorithm was moved into the event loop from the dns resolver thread.

### DIFF
--- a/source/channel_bootstrap.c
+++ b/source/channel_bootstrap.c
@@ -524,8 +524,7 @@ static void s_on_client_connection_established(struct aws_socket *socket, int er
                 error_code);
             /* connection_args will be released after setup_callback */
             s_connection_args_setup_callback(connection_args, error_code, NULL);
-        } 
-        else {
+        } else {
             /* this socket will die silently, free up the ref to the connection_args */
             s_connection_args_release(connection_args);
         }
--- a/source/channel_bootstrap.c
+++ b/source/channel_bootstrap.c
@@ -612,10 +612,10 @@ socket_init_failed:
 socket_alloc_failed:
     err_code = aws_last_error();
     AWS_LOGF_ERROR(
-            AWS_LS_IO_CHANNEL_BOOTSTRAP,
-            "id=%p: failed to create socket with error %d",
-            (void *)task_data->args->bootstrap,
-            err_code);
+        AWS_LS_IO_CHANNEL_BOOTSTRAP,
+        "id=%p: failed to create socket with error %d",
+        (void *)task_data->args->bootstrap,
+        err_code);
 task_cancelled:
     task_data->args->failed_count++;
     /* if this is the last attempted connection and it failed, notify the user */
--- a/source/channel_bootstrap.c
+++ b/source/channel_bootstrap.c
@@ -218,23 +218,27 @@ void s_connection_args_release(struct client_connection_args *args) {
     }
 }
 
-void s_connection_args_setup_callback(
+static void s_connection_args_setup_callback(
     struct client_connection_args *args,
     int error_code,
     struct aws_channel *channel) {
     if (!args->setup_called) {
+        AWS_ASSERT((error_code == AWS_OP_SUCCESS) == (channel != NULL));
         aws_client_bootstrap_on_channel_setup_fn *setup_callback = args->setup_callback;
         setup_callback(args->bootstrap, error_code, channel, args->user_data);
         args->setup_called = true;
     }
 }
 
-void s_connection_args_shutdown_callback(
+static void s_connection_args_shutdown_callback(
     struct client_connection_args *args,
     int error_code,
     struct aws_channel *channel) {
     if (!args->setup_called) {
-        s_connection_args_setup_callback(args, error_code, channel);
+        if (!error_code) {
+            error_code = AWS_ERROR_UNKNOWN;
+        }
+        s_connection_args_setup_callback(args, error_code, NULL);
         return;
     }
     aws_client_bootstrap_on_channel_shutdown_fn *shutdown_callback = args->shutdown_callback;
--- a/source/channel_bootstrap.c
+++ b/source/channel_bootstrap.c
@@ -602,7 +602,11 @@ static void s_attempt_connection(struct aws_task *task, void *arg, enum aws_task
 error:
     task_data->args->failed_count++;
     int err_code = aws_last_error();
-    AWS_LOGF_ERROR(AWS_LS_IO_CHANNEL_BOOTSTRAP, "id=%p: failed to create socket with error %d", (void*)task_data->args->bootstrap, err_code);
+    AWS_LOGF_ERROR(
+        AWS_LS_IO_CHANNEL_BOOTSTRAP,
+        "id=%p: failed to create socket with error %d",
+        (void *)task_data->args->bootstrap,
+        err_code);
     if (outgoing_socket) {
         aws_mem_release(allocator, outgoing_socket);
     }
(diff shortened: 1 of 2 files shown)


source/channel_bootstrap.c
```diff
@@ -410,8 +438,7 @@ static void s_on_client_channel_on_setup_completed(struct aws_channel *channel,
         err_code);
 
 error:
```
graebm: trivial: maybe move the LOGF_ERROR statement below the `error:` label so it prints every time

source/channel_bootstrap.c
```diff
@@ -635,8 +648,7 @@ static void s_on_host_resolved(
     if (err_code == 0) {
```
graebm: commenting on bug I found a bunch of lines up:
```
        if (client_connection_args->failed_count < client_connection_args->addresses_count) {
            s_connection_args_release(client_connection_args);
            return;
        }
```

This for loop will starts a socket connection for each address. After the for-loop, unless EVERY socket failed, we release a refcount on the connection args and assume things will resume when the socket connect callbacks fire.

BUT BUT BUT if during the for-loop we fail to allocator or init a socket, we break out of the loop. The logic was probably "well, if this failed then the rest surely will also fail". But that breaks the `failed_count < addresses_count` check later. If the 1st socket failed, then there are no pending socket connects out there

source/channel_bootstrap.c
```diff
@@ -218,29 +218,59 @@ void s_connection_args_release(struct client_connection_args *args) {
...
+    int error_code,
+    struct aws_channel *channel) {
+    if (!args->setup_called) {
+        if (!error_code) {
+            error_code = AWS_ERROR_UNKNOWN;
+        }
+        s_connection_args_setup_callback(args, error_code, NULL);
+    }
+    aws_client_bootstrap_on_channel_shutdown_fn *shutdown_callback = args->shutdown_callback;
+    shutdown_callback(args->bootstrap, error_code, channel, args->user_data);
```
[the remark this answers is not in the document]
justinboswell: Yes, this is required. Otherwise, unit tests start failing. There's a lot of code that assumes that both will always be called.

source/channel_bootstrap.c
```diff
@@ -410,8 +438,7 @@ static void s_on_client_channel_on_setup_completed(struct aws_channel *channel,
         err_code);
 
 error:
-    connection_args->setup_callback(connection_args->bootstrap, err_code, NULL, connection_args->user_data);
-
+    aws_channel_shutdown(channel, err_code);
```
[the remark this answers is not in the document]
justinboswell: That's how it was, and we switched to this... I think the way it's setup is working out more consistently, it ensures that the shutdown is performed.

source/channel_bootstrap.c
graebm: probably solution is incrementing client_connection_args->addresses_count++ at the start of each for-loop, instead of setting it once before going into the for-loop.

source/channel_bootstrap.c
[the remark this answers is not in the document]
justinboswell: Discussed offline, this matches our desired architecture

source/channel_bootstrap.c
justinboswell: This turned out to be much harder to fix than it looks. We will need to re-visit this when on_host_resolved takes place in an event loop.

justinboswell: I ended up actually fixing this, by moving the socket creation into the event loop

source/channel_bootstrap.c
```diff
@@ -534,14 +554,73 @@ static void s_on_client_connection_established(struct aws_socket *socket, int er
...
+};
+
+static void s_attempt_connection(struct aws_task *task, void *arg, enum aws_task_status status) {
+    (void)task;
+    struct connection_task_data *task_data = arg;
+    struct aws_allocator *allocator = task_data->args->bootstrap->allocator;
+
+    if (status != AWS_TASK_STATUS_RUN_READY) {
+        s_connection_args_release(task_data->args);
+        goto cleanup_task;
```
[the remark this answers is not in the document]
justinboswell: Done.

source/channel_bootstrap.c
```diff
@@ -553,91 +632,53 @@ static void s_on_host_resolved(
...
+    /* use this event loop for all outgoing connection attempts (only one will ultimately win). */
+    struct aws_event_loop *connect_loop =
+        aws_event_loop_group_get_next_loop(client_connection_args->bootstrap->event_loop_group);
+    client_connection_args->addresses_count = (uint8_t)host_addresses_len;
+
+    for (size_t i = 0; i < host_addresses_len; ++i) {
+        s_connection_args_acquire(client_connection_args);
+
+        struct connection_task_data *task_data =
+            aws_mem_calloc(client_connection_args->bootstrap->allocator, 1, sizeof(struct connection_task_data));
```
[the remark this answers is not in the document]
justinboswell: Done.

source/channel_bootstrap.c
```diff
@@ -577,16 +578,16 @@ static void s_attempt_connection(struct aws_task *task, void *arg, enum aws_task
 
     if (status != AWS_TASK_STATUS_RUN_READY) {
         s_connection_args_release(task_data->args);
-        goto cleanup_task;
+        goto task_cancelled;
```
graebm: trivial, but after this jump it's going to LOGF something about aws_last_error(), but we don't have a last error set, so it will log a random value.

Either move the LOGFs around, or raise an error here so something reasonable prints out later

source/channel_bootstrap.c
```diff
@@ -653,28 +660,53 @@ static void s_on_host_resolved(
...
+                if (tasks[j]) {
+                    aws_mem_release(allocator, tasks[j]);
+                }
+            }
+            AWS_LOGF_ERROR(
+                AWS_LS_IO_CHANNEL_BOOTSTRAP,
+                "id=%p: failed to allocate connection task data: err=%d",
+                (void *)client_connection_args->bootstrap,
+                aws_last_error());
+            return;
```
graebm: need to call setup_callback(FAILED)

source/channel_bootstrap.c
```diff
@@ -653,28 +660,53 @@ static void s_on_host_resolved(
...
+            task_data->connect_loop = connect_loop;
+        }
 
-        task_data->options = client_connection_args->outgoing_options;
-        task_data->options.domain =
-            host_address_ptr->record_type == AWS_ADDRESS_RECORD_TYPE_AAAA ? AWS_SOCKET_IPV6 : AWS_SOCKET_IPV4;
+        if (failed) {
+            for (size_t j = 0; j <= i; ++j) {
+                if (tasks[j]) {
+                    aws_mem_release(allocator, tasks[j]);
```
graebm: there might also be aws_host_addresses that need to be cleaned up

source/channel_bootstrap.c
```diff
@@ -534,14 +555,78 @@ static void s_on_client_connection_established(struct aws_socket *socket, int er
...
+    struct aws_event_loop *connect_loop;
+};
+
+static void s_attempt_connection(struct aws_task *task, void *arg, enum aws_task_status status) {
+    (void)task;
+    struct connection_task_data *task_data = arg;
+    struct aws_allocator *allocator = task_data->args->bootstrap->allocator;
+
+    if (status != AWS_TASK_STATUS_RUN_READY) {
+        s_connection_args_release(task_data->args);
```
graebm: s_connection_args_release() is getting run again after the goto

source/channel_bootstrap.c
```diff
@@ -218,29 +218,66 @@ void s_connection_args_release(struct client_connection_args *args) {
     }
 }
 
+static void s_connection_args_setup_callback(
+    struct client_connection_args *args,
+    int error_code,
+    struct aws_channel *channel) {
+    if (!args->setup_called) {
```
graebm: worth making this `ASSERT(!setup_called)`?

source/channel_bootstrap.c
justinboswell: Good catch.

source/channel_bootstrap.c
justinboswell: Done.

source/channel_bootstrap.c
```diff
@@ -802,6 +808,8 @@ static inline int s_new_client_channel(
         memcpy(endpoint.address, host_name, strlen(host_name));
         endpoint.port = 0;
 
+        s_connection_args_acquire(client_connection_args);
```
graebm: you moved this too high.
it should go immediately before the call to aws_socket_connect(), and if the connect call fails you should s_connection_args_release() before the goto error;

PR #157, opened 2019-07-24
Server fix

TingDaoK:
*Issue #, if available:*

*Description of changes:*

Some bugs in connection failing. Add ref count to the server_connection_arg.
And two regression test for the bugs.


source/channel_bootstrap.c
```diff
@@ -1166,14 +1188,11 @@ static void s_on_server_channel_on_shutdown(struct aws_channel *channel, int err
 
     channel_data->server_connection_args->shutdown_callback(
         server_bootstrap, error_code, channel, server_shutdown_user_data);
+    s_server_connection_args_release(channel_data->server_connection_args);
```
graebm: I don't *think* there are bugs here, but let's move the release() down as far as we can, *right* before we delete channel_data.

Who knows if there's some critical data in there that the channel needs to destroy itself.

source/channel_bootstrap.c
```diff
@@ -1237,6 +1256,7 @@ void s_on_server_connection_result(
 
 error_cleanup:
     connection_args->incoming_callback(connection_args->bootstrap, aws_last_error(), NULL, connection_args->user_data);
+    s_server_connection_args_release(connection_args);
```
graebm: likewise, let's move the release() to be as late as possible in this function

source/channel_bootstrap.c
```diff
@@ -1377,10 +1397,10 @@ int aws_server_bootstrap_destroy_socket_listener(struct aws_server_bootstrap *bo
         AWS_CONTAINER_OF(listener, struct server_connection_args, listener);
 
     AWS_LOGF_DEBUG(AWS_LS_IO_CHANNEL_BOOTSTRAP, "id=%p: releasing bootstrap reference", (void *)bootstrap);
+
     aws_socket_stop_accept(listener);
     aws_socket_clean_up(listener);
-    aws_mem_release(bootstrap->allocator, server_connection_args);
-    s_server_bootstrap_release(bootstrap);
+    s_server_connection_args_release(server_connection_args);
```
graebm: Again, not sure if this could cause a bug, but seems like a good idea to move `aws_socket_clean_up(listener);` into the release() function?

tests/tls_handler_test.c
```diff
@@ -843,3 +843,320 @@ static int s_tls_client_channel_negotiation_success_fn(struct aws_allocator *all
...
+        endpoint.address,
+        0,
+        &options,
+        &tls_client_conn_options,
+        s_tls_handler_test_client_setup_callback,
+        s_tls_handler_test_client_shutdown_callback,
+        &outgoing_args));
+    /* put this here to verify ownership semantics are correct. This should NOT cause a segfault. If it does, ya
+     * done messed up. */
+    aws_tls_connection_options_clean_up(&tls_client_conn_options);
```
graebm: trivial: no need to copy/paste this weird part of the test

tests/tls_handler_test.c
```diff
@@ -843,3 +843,320 @@ static int s_tls_client_channel_negotiation_success_fn(struct aws_allocator *all
...
+    aws_tls_clean_up_static_state();
+    return AWS_OP_SUCCESS;
+}
+AWS_TEST_CASE(tls_server_multiple_connections, s_tls_server_multiple_connections_fn)
+
+struct shutdown_listener_tester {
+    struct aws_socket *listener;
+    struct aws_server_bootstrap *server_bootstrap;
+    struct tls_test_args *outgoing_args;
+    struct aws_socket *socket;
```
graebm: style suggestions:
- instead of having `outgoing_socket` declared in the test function, and making `tester->socket *` point to it later, just have tester->socket be the actual client socket itself. It would just make the code simpler since there would be less things pointing at each other.
- also, could you rename this `client_socket`? I think that's more clear than `socket` or `outgoing_socket`

tests/tls_handler_test.c
```diff
@@ -843,3 +843,320 @@ static int s_tls_client_channel_negotiation_success_fn(struct aws_allocator *all
...
+        .rw_handler = NULL,
+        .server = true,
+        .tls_negotiated = false,
+        .shutdown_finished = false,
+    };
+
+    struct tls_test_args outgoing_args = {
+        .mutex = &mutex,
+        .allocator = allocator,
+        .condition_variable = &condition_variable,
```
graebm: Man, it would be nice if the tests in this file didn't require sooooo much copy/paste. Like if there were a single s_tester_init() for all this common setup, like the tests in HTTP.

Don't do it in the PR, it would require changing too much unrelated code. But it would be a wise thing to do as a follow-up PR. So it's easier to add more tests in the future.

tests/tls_handler_test.c
```diff
@@ -843,3 +843,320 @@ static int s_tls_client_channel_negotiation_success_fn(struct aws_allocator *all
...
+    aws_event_loop_current_clock_time(socket->event_loop, &run_at_ns);
+    run_at_ns += aws_timestamp_convert(1, AWS_TIMESTAMP_SECS, AWS_TIMESTAMP_NANOS, NULL);
+    struct aws_task *shutdown_listener_task =
+        aws_mem_acquire(tester->outgoing_args->allocator, sizeof(struct aws_task));
+    aws_task_init(shutdown_listener_task, s_shutdown_listener_task, tester, "wait_a_bit");
+    /* wait 1 sec for server side setup the channel then shut down the listener and close the socket */
+    aws_event_loop_schedule_task_future(socket->event_loop, shutdown_listener_task, run_at_ns);
+}
+
+static int s_tls_server_destroy_by_user_when_connection_is_in_processing_fn(
```
graebm: Did you actually confirm that these tests fail when the fixes are removed?

source/channel_bootstrap.c
```diff
@@ -1142,11 +1168,7 @@ static void s_on_server_channel_on_setup_completed(struct aws_channel *channel,
         err_code,
         NULL,
         channel_data->server_connection_args->user_data);
-
-    if (channel_data->server_connection_args->use_tls) {
-        aws_tls_connection_options_clean_up(&channel_data->server_connection_args->tls_options);
-    }
-
+    s_server_connection_args_release(channel_data->server_connection_args);
```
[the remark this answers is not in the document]
TingDaoK: I think I acquire the arg before the err_code is checked

tests/tls_handler_test.c
TingDaoK: Yes! I pull the original code, and it will fail in the new tests

tests/tls_handler_test.c
TingDaoK: ...I also think about a tester_init() function will be much better.

source/channel_bootstrap.c
```diff
@@ -944,6 +947,28 @@ struct server_channel_data {
...
+        s_server_bootstrap_acquire(args->bootstrap);
+    }
+}
+
+void s_server_connection_args_release(struct server_connection_args *args) {
+    AWS_ASSERT(args);
+
+    if (aws_atomic_fetch_sub(&args->ref_count, 1) == 1) {
+        struct aws_allocator *allocator = args->bootstrap->allocator;
+        aws_socket_clean_up(&args->listener);
```
graebm: I was wrong, move socket_clean_up() back to the destroy() function.

PR #160, opened 2019-07-26
Dynamic TLS setup

bretambrose:
Makes it easier to negotiate tls independently of connection establishment.  Useful for tls proxy support


source/tls_channel_handler.c
```diff
@@ -388,6 +388,7 @@ int aws_tls_connection_options_set_alpn_list(
 
 int aws_channel_setup_client_tls(
     struct aws_channel *channel,
```
ColdenCullen: I don't think you actually need this param anymore... Channel can be pulled off of the slot.

bretambrose: Made a new PR with updates

PR #161, opened 2019-07-26
Add Native TLS Cipher APIs

alexw91:
*Issue #, if available:* N/A

*Description of changes:*
 - Adds `aws_tls_cipher_pref ` enum to `aws_tls_ctx ` struct. 
 - Adds `aws_tls_is_cipher_pref_supported()` API  to Windows, OSX, and Linux TLS Implementations
- s2n is only TLS Implementation that supports non-Default Cipher Setting
- Other TLS Implementations detect non-Default Cipher Setting and return an error if configured.


source/s2n/s2n_tls_channel_handler.c
```diff
@@ -872,6 +889,19 @@ static struct aws_tls_ctx *s_tls_ctx_new(
             s2n_config_set_cipher_preferences(s2n_ctx->s2n_config, "default");
     }
 
+    switch (options->cipher_pref) {
+        case AWS_IO_TLS_CIPHER_PREF_SYSTEM_DEFAULT:
+            /* No-Op, if the user configured a minimum_tls_version then a version-specific Cipher Preference was set */
+            break;
+        case AWS_IO_TLS_CIPHER_PREF_KMS_PQ_TLSv1_0_2019_06:
```
justinboswell: Do you think we'll have enough of these that it's worth making a mapping table like s2n has? What about sets like s2n has? I'm not sure we want to maintain a huge switch long term.

alexw91: > Do you think we'll have enough of these that it's worth making a mapping table like s2n has?

Maybe eventually if there are more Cipher Preferences, but probably not soon. 

> What about sets like s2n has?

Not sure if I understand the question. aws-c-io might need to maintain equivalency mappings from s2n's Cipher Preferences to Windows/OSX Cipher Preferences, but I don't think aws-c-io should be creating any brand new Cipher Preference Sets. 

>I'm not sure we want to maintain a huge switch long term.

Agreed. If aws-c-io starts adding more values to this enum, it probably makes sense to refactor some of this into a lookup table like s2n does: https://github.com/awslabs/s2n/blob/3cda894f20be30216e6f7d780f8bb97307a8a651/tls/s2n_cipher_preferences.c#L806

PR #163, opened 2019-07-27
bug fix for fire callbacks correctly

TingDaoK:
*Issue #, if available:*

*Description of changes:*


source/channel_bootstrap.c
```diff
@@ -1031,8 +1056,9 @@ static inline int s_setup_server_tls(struct server_connection_args *connection_a
         return AWS_OP_ERR;
     }
 
-    tls_handler =
-        aws_tls_server_handler_new(connection_args->bootstrap->allocator, &connection_args->tls_options, tls_slot);
+    struct aws_tls_connection_options tls_options = connection_args->tls_options;
```
graebm: add comment: /* Shallow-copy tls_options so we can override the user_data, making it specific to this channel */

source/channel_bootstrap.c
```diff
@@ -1160,41 +1182,41 @@ static void s_on_server_channel_on_setup_completed(struct aws_channel *channel,
...
+    struct aws_server_bootstrap *server_bootstrap = args->bootstrap;
     struct aws_allocator *allocator = server_bootstrap->allocator;
 
-    channel_data->server_connection_args->shutdown_callback(
-        server_bootstrap, error_code, channel, server_shutdown_user_data);
+    if (!channel_data->incoming_called) {
+        error_code = (error_code) ? error_code : AWS_ERROR_UNKNOWN;
+        s_server_incoming_callback(channel_data, error_code, NULL);
+    } else if (channel_data->can_shutdown) {
+        args->shutdown_callback(server_bootstrap, error_code, channel, server_shutdown_user_data);
```
graebm: There's no need for this `can_shutdown` variable, this can just be an `else {` branch.

If the channel exists, but some error occurs, we **don't want to allow** s_server_incoming_callback to be fired until the channel-shutdown callback.

PR #167, opened 2019-07-30
Document proper stream behavior. Defend against bad implementations.

graebm:
I had a check like this in websockets to ensure the buffer wasn't resized.
Thought it would be good if streams always did that check.

### DIFF
--- a/source/stream.c
+++ b/source/stream.c
@@ -31,8 +31,33 @@ int aws_input_stream_seek(struct aws_input_stream *stream, aws_off_t offset, enu
 
 int aws_input_stream_read(struct aws_input_stream *stream, struct aws_byte_buf *dest) {
     AWS_ASSERT(stream && stream->vtable && stream->vtable->read);
+    AWS_ASSERT(dest);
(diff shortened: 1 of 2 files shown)


source/stream.c
```diff
@@ -31,8 +31,33 @@ int aws_input_stream_seek(struct aws_input_stream *stream, aws_off_t offset, enu
...
+
+    /* Prevent implementations from accidentally overwriting existing data in the buffer.
+     * Hand them a "safe" buffer that starts where the existing data ends. */
+    const void *safe_buf_start = dest->buffer + dest->len;
+    const size_t safe_buf_capacity = dest->capacity - dest->len;
+    struct aws_byte_buf safe_buf = aws_byte_buf_from_empty_array(safe_buf_start, safe_buf_capacity);
+
+    int read_result = stream->vtable->read(stream, &safe_buf);
+
+    /* Ensure the implementation did not commit forbidden acts upon the buffer */
```
justinboswell: Comment of the Week (TM)

PR #172, opened 2019-08-06
async channel shutdown

TingDaoK:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/source/channel.c
+++ b/source/channel.c
@@ -328,7 +328,7 @@ static void s_shutdown_task(struct aws_channel_task *task, void *arg, enum aws_t
                     (void *)slot);
 
                 aws_channel_slot_shutdown(slot, AWS_CHANNEL_DIR_READ, error_code, shutdown_immediately);
-                return ;
+                return;
             }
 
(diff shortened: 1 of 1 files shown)


source/channel.c
```diff
@@ -303,21 +303,17 @@ struct channel_shutdown_task_args {
...
 static void s_shutdown_task(struct aws_channel_task *task, void *arg, enum aws_task_status status) {
 
     (void)task;
 
     struct shutdown_task *shutdown_task = arg;
-
+    struct aws_channel *channel = shutdown_task->channel;
+    int error_code = shutdown_task->error_code;
+    bool shutdown_immediately = shutdown_task->shutdown_immediately;
     if (status == AWS_TASK_STATUS_RUN_READY) {
```
graebm: This `if (status == AWS_TASK_STATUS_RUN_READY) ` should be removed. This was a lurking bug in the existing code.

The other possible status for a task is "CANCELLED". This can happen if the event-loop is being torn down. In those cases, we still want to finish shutting down the channels on that event-loop.

PR #177, opened 2019-08-20
Implement io library-init functions

ColdenCullen:


tests/channel_test.c
```diff
@@ -820,8 +820,8 @@ static bool s_shutdown_complete_pred(void *user_data) {
 static int s_test_channel_connect_some_hosts_timeout(struct aws_allocator *allocator, void *ctx) {
     (void)ctx;
 
-    aws_load_error_strings();
-    aws_io_load_error_strings();
+    aws_common_library_init();
+    aws_io_library_init();
```
graebm: I would have aws_io_library_init() automatically call aws_common_library_init() because:
1) Saves boilerplate. You can't use a library without its dependencies anyway.
2) Lets us add new low-lever libraries later without every downstream app needing to copy/paste 1 more library to be initialized

source/io.c
```diff
@@ -158,11 +160,46 @@ static struct aws_error_info_list s_list = {
...
+static bool s_io_library_initialized = false;
 
-void aws_io_load_error_strings(void) {
-    if (!s_error_strings_loaded) {
-        s_error_strings_loaded = true;
+void aws_io_library_init(void) {
+    if (!s_io_library_initialized) {
+        s_io_library_initialized = true;
         aws_register_error_info(&s_list);
+        aws_register_log_subject_info_list(&s_io_log_subject_list);
```
graebm: `aws_tls_init_static_state()` is another bit of startup boilerplate we could move in here

ColdenCullen: Will do!

PR #185, opened 2019-09-12
Tls cleanup

bretambrose:


source/s2n/s2n_tls_channel_handler.c
```diff
@@ -717,6 +719,38 @@ static int s_parse_protocol_preferences(
     return AWS_OP_SUCCESS;
 }
 
+static size_t s_tl_cleanup_key = 0; /* Address of variable serves as key in hash table */
+static struct aws_event_loop_local_object s_tl_cleanup_object;
```
graebm: make a note that this same object is inserted into multiple event loops' hash tables

source/s2n/s2n_tls_channel_handler.c
```diff
@@ -717,6 +719,38 @@ static int s_parse_protocol_preferences(
...
+
+static void s_aws_cleanup_s2n_thread_local_state(void *user_data) {
+    (void)user_data;
+
+    aws_tls_clean_up_thread_local_state();
+}
+
+static int s_s2n_tls_channel_handler_schedule_tl_cleanup(struct aws_channel_slot *slot) {
```
graebm: Add comment explaining what this is all about. The "tl" acronym isn't super obvious in a "tls" file.
/* s2n allocates many thread-local datastructures. We need to clean these up when the even-loop's thread exits. */

source/s2n/s2n_tls_channel_handler.c
```diff
@@ -717,6 +719,38 @@ static int s_parse_protocol_preferences(
...
+    struct aws_channel *channel = slot->channel;
+
+    struct aws_event_loop_local_object existing_marker;
+    AWS_ZERO_STRUCT(existing_marker);
+
+    if (aws_channel_fetch_local_object(channel, &s_tl_cleanup_key, &existing_marker)) {
```
graebm: /* Check whether any other s2n-channel-handler on this event-loop has already scheduled the cleanup task */

source/s2n/s2n_tls_channel_handler.c
```diff
@@ -717,6 +719,38 @@ static int s_parse_protocol_preferences(
...
+        /* Doesn't exist in event loop table: add it and add the at-exit cleanup callback */
+        AWS_ZERO_STRUCT(s_tl_cleanup_object);
+        s_tl_cleanup_object.key = &s_tl_cleanup_key;
+        s_tl_cleanup_object.object = NULL;
+        s_tl_cleanup_object.on_object_removed = NULL;
```
graebm: Since we're inserting the same object into multiple threads, there's a chance that on thread zeroes out memory while another thread is reading it.

You could just set all this at static initialization time
```
static struct aws_event_loop_local_object s_tl_cleanup_object = {
    .key = &s_tl_cleanup_key,
    ...
};
```

source/s2n/s2n_tls_channel_handler.c
```diff
@@ -717,6 +719,38 @@ static int s_parse_protocol_preferences(
...
+
+        if (aws_channel_put_local_object(channel, &s_tl_cleanup_key, &s_tl_cleanup_object)) {
+            return AWS_OP_ERR;
+        }
```
graebm: So the event-loop API supports stop (join the thread) and restart (start up another thread), but it keeps the same hash table. A few issues...
1) We never remove the item from the hash-table, so we won't be able to schedule an at-exit task on the restarted thread.
2) Not sure where in this code we'd re-check to see if we're on a new thread that needs the at-exit task rescheduled. It's kind of like you'd need to give the event loop a pair of thread-start and thread-stop functions, instead of JUST a thread-stop function.
2) What on earth *would* happen to s2n if it was suddenly running on a different thread than before? Would it get confused because it was using thread-local-storage that was cleaned up?

bretambrose: Do the channels persist over stop-run?

bretambrose: Ok, talked to henso a bit about things and looked at the referenced s2n functionality.

For this particular case, (3) is safe because s2n_cleanup currently just wipes RNG state and their RNG generation functions actually reinitialize state if they detect a new thread.

(1) Can be fixed if we set the event loop as the callback's user data.  We could safely remove the object from the table at callback invocation since event loop would still be alive.

(2) I don't have a good answer for.  It seems like we may want to do some internal analysis at the beginning of the thread function for event loop that checks state (do we have existing channels, maybe call a hey-you're-running-on-a-new-thread functions on the channel which could in turn percolate to the handlers and their we'd have the opportunity to reschedule a tl cleanup exit function).

PR #186, opened 2019-09-13
Elg async cleanup2

bretambrose:
Allows for event loop cleanup to be safely invoked and resolved from an event loop thread.

### DIFF
--- a/include/aws/io/event_loop.h
+++ b/include/aws/io/event_loop.h
@@ -115,6 +115,8 @@ struct aws_event_loop_group {
     struct aws_atomic_var current_index;
 };
 
+typedef void(aws_event_loop_group_cleanup_complete_fn)(void *user_data);
+
 AWS_EXTERN_C_BEGIN
 
 #ifdef AWS_USE_IO_COMPLETION_PORTS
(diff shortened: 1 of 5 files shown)


source/event_loop.c
```diff
@@ -94,6 +95,51 @@ void aws_event_loop_group_clean_up(struct aws_event_loop_group *el_group) {
...
+    AWS_FATAL_ASSERT(aws_thread_init(&cleanup_thread, el_group->allocator) == AWS_OP_SUCCESS);
+
+    struct aws_thread_options thread_options;
+    AWS_ZERO_STRUCT(thread_options);
+
+    AWS_FATAL_ASSERT(
+        aws_thread_launch(&cleanup_thread, s_event_loop_destroy_async_thread_fn, data, &thread_options) ==
+        AWS_OP_SUCCESS);
+
+    aws_thread_clean_up(&cleanup_thread);
```
JonathanHenson: We don’t need to join?

JonathanHenson: Nevermind, I’m dumb, you intentionally detached so it would stay async

bretambrose: clean_up = detach

include/aws/io/event_loop.h
```diff
@@ -115,6 +115,8 @@ struct aws_event_loop_group {
     struct aws_atomic_var current_index;
 };
 
+typedef void(aws_event_loop_group_cleanup_complete_fn)(void *user_data);
```
graebm: we usually pass along the pointer to the event-loop-group

bretambrose: In cases where it's not safe (the loop is gone) I think it's up to the user to create some sort of association via user data if they need to associate specific callbacks with specific loops.

PR #190, opened 2019-09-17
Fix TLS test

graebm:
Fix an occasional use-after-free crash in this test.
Also rename this test and add a bunch of comments because it took me a while to try and figure out what exactly it was testing.

Also remove return code from `aws_server_bootstrap_destroy_socket_listener()` because destructors cannot throw


graebm:
Race was between main thread and 2 event-loop threads, 1 for client 1 for server:
- CLIENT THREAD: async call to server_destroy()
- CLIENT THREAD: sync call to socket_close() ....
- SERVER THREAD: ok done
- MAIN THREAD: ok server done, clean up client socket now
- CLIENT THREAD: ... socket_close() still running, but is now in use-after-free town

Fix is to wait for client socket to finish closing before issuing async server_shutdown() call

PR #193, opened 2019-10-02
Fix for 0-valued error code when socket closed remotely

bretambrose:
Make sure a final read on an s2n-enabled connection whose socket has been closed remotely generates a socket closed error code.


JonathanHenson:
A channel shutting down gracefully is not an error. An alert that is not an error code suggests graceful shutdown. This the current behavior is correct. If this manifests with a bad user experience for Mqtt clients, we should ask the iot team to provide proper alert codes.

PR #198, opened 2019-10-28
Wait for alert to send before continuing shutdown.

JonathanHenson:
Make sure the outgoing TLS alert actually gets sent before letting the channel go away.

This isn't necessary on non-windows platforms, since we own the flushing logic on the underlying socket.

### DIFF
--- a/source/windows/secure_channel_tls_handler.c
+++ b/source/windows/secure_channel_tls_handler.c
@@ -1393,6 +1393,27 @@ static size_t s_initial_window_size(struct aws_channel_handler *handler) {
     return EST_HANDSHAKE_SIZE;
 }
 
+struct alert_data {
+    struct aws_allocator *allocator;
+    struct aws_channel_slot *slot;
+    int shutdown_error;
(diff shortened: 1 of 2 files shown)


source/windows/secure_channel_tls_handler.c
```diff
@@ -1465,12 +1486,31 @@ static int s_handler_shutdown(
...
+                alert_data->slot = slot;
+                alert_data->shutdown_error = error_code;
+                alert_data->shutdown_immediately = abort_immediately;
+
+                outgoing_message->on_completion = s_on_alert_sent;
+                outgoing_message->user_data = alert_data;
+
+                memcpy(outgoing_message->message_data.buffer, output_buffer.pvBuffer, output_buffer.cbBuffer);                
+                
+                /* we don't really care if this succeeds or not, it's just sending the TLS alert. */
```
justinboswell: Comment says we don't care, but we are definitely changing the return value, so, maybe we do?

justinboswell (approved):
Question and Ship

PR #200, opened 2019-10-29
fix epoll data race

graebm:
@nchong-at-aws found this. Thanks Nathan!

If the user subscribes from a different thread, it was possible for the event-loop-thread to process an event before `handle->additional_data` was set.

This bug is not present in the kqueue or IOCP event-loops:
- kqueue uses a task to always do the actual subscription on the event-loop-thread.
- IOCP doesn't make use of `handle->additional_data`


source/linux/epoll_event_loop.c
```diff
@@ -407,11 +407,10 @@ static int s_subscribe_to_io_events(
         AWS_LOGF_ERROR(
             AWS_LS_IO_EVENT_LOOP, "id=%p: failed to subscribe to events on fd %d", (void *)event_loop, handle->data.fd);
         aws_mem_release(event_loop->alloc, epoll_event_data);
+        handle->additional_data = NULL;
```
ColdenCullen: I would be EVER-SO-SLIGHTLY more comfortable if you move this above the mem release here, just because there's now an ever-so-slight chance that another thread or could see `handle->additional_data` point at freed memory.

graebm: ok

PR #201, opened 2019-11-11
fix epoll stop data race

nchong-at-aws:
A stop call (aws_event_loop_stop) on an epoll event loop updates the same stop
task regardless of whether the stop task is currently scheduled. This can lead
to null pointer dereferences.

This bug is not present in the kqueue or IOCP event-loops, which use a
mutex-protected status flag.


source/linux/epoll_event_loop.c
```diff
@@ -286,11 +288,19 @@ static void s_stop_task(struct aws_task *task, void *args, enum aws_task_status
          */
         epoll_loop->should_continue = false;
     }
+    aws_atomic_store_ptr(&epoll_loop->stop_task_ptr, NULL);
```
graebm: Should we reset the atomic BEFORE setting `epoll_loop->should_continue`?
At this point the task is out of the scheduler and it's safe for someone else to get it re-scheduled via a call to stop()

justinboswell: :+1:

PR #210, opened 2019-12-01
get event loop for aws_channel

bretambrose:
Adds get accessor to go from channel (hidden impl) to the owning event loop.  Needed in order to cancel tasks from a channel handler.

### DIFF
--- a/include/aws/io/channel.h
+++ b/include/aws/io/channel.h
@@ -204,6 +204,12 @@ AWS_IO_API
 struct aws_channel_slot *aws_channel_slot_new(struct aws_channel *channel);
 
 /**
+ * Fetches the event loop the channel is a part of.
+ */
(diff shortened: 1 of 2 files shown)


JonathanHenson:
There’s already a channel cancel task api. Is this because the reconnect task is not on the channel? If so why not just grab an event loop off of the eventloop group directly?

JonathanHenson:
Yeah I’m now confused why you need this at all. You already have the event loop group

bretambrose:
> 
> 
> There’s already a channel cancel task api. Is this because the reconnect task is not on the channel? If so why not just grab an event loop off of the eventloop group directly?

I did not see cancel task channel/handler/slot api.

bretambrose:
Closing and renaming to match downstream branch name

PR #213, opened 2019-12-04
Fix warning from gcc 8.3 on ARM

graebm:
Customer hit this warning-as-error on Raspberry Pi: https://github.com/aws/aws-iot-device-sdk-js-v2/issues/10

### DIFF
--- a/source/posix/socket.c
+++ b/source/posix/socket.c
@@ -18,6 +18,7 @@
 #include <aws/common/clock.h>
 #include <aws/common/condition_variable.h>
 #include <aws/common/mutex.h>
+#include <aws/common/string.h>
 
 #include <aws/io/event_loop.h>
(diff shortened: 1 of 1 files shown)


source/posix/socket.c
```diff
@@ -533,6 +534,11 @@ int aws_socket_connect(
         AWS_ASSERT(on_connection_result);
     }
 
+    size_t address_strlen;
+    if (aws_secure_strlen(remote_endpoint->address, AWS_ADDRESS_MAX_LEN, &address_strlen)) {
```
JonathanHenson: yaay! somebody finally using this function.

graebm: We should support passing NULL as the out-param. I just wanted to sanity check that the string was long enough

PR #215, opened 2019-12-10
bootstrap-shutdown-cb

graebm:
Invoke callback when aws_client_bootstrap finishes shutting down. This is necessary because we need to keep the aws_host_resolver alive at least as long as the bootstrap.

### DIFF
--- a/include/aws/io/channel_bootstrap.h
+++ b/include/aws/io/channel_bootstrap.h
@@ -174,7 +174,7 @@ AWS_EXTERN_C_BEGIN
  */
 AWS_IO_API struct aws_client_bootstrap *aws_client_bootstrap_new(
     struct aws_allocator *allocator,
-    struct aws_client_bootstrap_options *options);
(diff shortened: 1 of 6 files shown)


include/aws/io/channel_bootstrap.h
```diff
@@ -140,15 +170,11 @@ struct aws_server_bootstrap {
...
- * channels. host_resolver will be used for resolving host names.
- * If host_resolution_config is NULL, the default will be used, host_resolution_config will be copied.
+ * Create the client bootstrap.
  */
 AWS_IO_API struct aws_client_bootstrap *aws_client_bootstrap_new(
     struct aws_allocator *allocator,
-    struct aws_event_loop_group *el_group,
-    struct aws_host_resolver *host_resolver,
-    struct aws_host_resolution_config *host_resolution_config);
+    struct aws_client_bootstrap_options *options);
```
ColdenCullen: can be `const`.

Issue #217, opened 2019-12-19
Non portable use of posix pthread_t 


nchong-at-aws:
posix pthread_t is opaque so comparison with `==` in event loop linux and bsd implementations is not portable.

Related to https://github.com/awslabs/aws-c-common/issues/562

nchong-at-aws:
There is one further non-portable use of thread id not tackled by the PR. In `source/s2n/s2n_tls_channel_handler.c` there is an `(unsigned long)` cast of `aws_thread_current_thread_id()` in `s_id_fn`. This cast is required because of `CRYPTO_set_id_callback`.

PR #218, opened 2019-12-19
Remove non portable use of pthread_t

nchong-at-aws:
*Issue #, if available:* #217 

*Description of changes:*

Use portable thread id comparison functions from proposed c common changes

### DIFF
--- a/source/bsd/kqueue_event_loop.c
+++ b/source/bsd/kqueue_event_loop.c
@@ -67,8 +67,15 @@ enum pipe_fd_index {
 };
 
 struct kqueue_loop {
-    struct aws_thread thread;
-    struct aws_atomic_var thread_id;
+    /* thread_created_on is the handle to the event loop thread. */
+    struct aws_thread thread_created_on;
+    /* thread_joined_to is used by the thread destroying the event loop. */
+    aws_thread_id_t thread_joined_to;
+    /* running_thread_id is NULL if the event loop thread is stopped or points-to the thread_id of the thread running
+     * the event loop (either thread_created_on or thread_joined_to). Atomic because of concurrent writes (e.g.,
+     * run/stop) and reads (e.g., is_event_loop_thread).
+     * An aws_thread_id_t variable itself cannot be atomic because it is an opaque type that is platform-dependent. */
+    struct aws_atomic_var running_thread_id;
     int kq_fd; /* kqueue file descriptor */
 
     /* Pipe for signaling to event-thread that cross_thread_data has changed. */
@@ -163,10 +170,10 @@ struct aws_event_loop *aws_event_loop_new_default(struct aws_allocator *alloc, a
         goto clean_up;
     }
     /* intialize thread id to NULL. It will be set when the event loop thread starts. */
-    aws_atomic_init_ptr(&impl->thread_id, NULL);
+    aws_atomic_init_ptr(&impl->running_thread_id, NULL);
     clean_up_impl_mem = true;
 
-    err = aws_thread_init(&impl->thread, alloc);
+    err = aws_thread_init(&impl->thread_created_on, alloc);
     if (err) {
         goto clean_up;
     }
@@ -267,7 +274,7 @@ clean_up:
         close(impl->kq_fd);
     }
     if (clean_up_thread) {
-        aws_thread_clean_up(&impl->thread);
+        aws_thread_clean_up(&impl->thread_created_on);
     }
     if (clean_up_impl_mem) {
         aws_mem_release(alloc, impl);
@@ -297,8 +304,8 @@ static void s_destroy(struct aws_event_loop *event_loop) {
         return;
     }
     /* setting this so that canceled tasks don't blow up when asking if they're on the event-loop thread. */
-    impl->thread.thread_id = aws_thread_current_thread_id();
-    aws_atomic_store_ptr(&impl->thread_id, &impl->thread.thread_id);
+    impl->thread_joined_to = aws_thread_current_thread_id();
+    aws_atomic_store_ptr(&impl->running_thread_id, &impl->thread_joined_to);
 
     /* Clean up task-related stuff first. It's possible the a cancelled task adds further tasks to this event_loop.
      * Tasks added in this way will be in cross_thread_data.tasks_to_schedule, so we clean that up last */
@@ -338,7 +345,7 @@ static void s_destroy(struct aws_event_loop *event_loop) {
     close(impl->cross_thread_signal_pipe[READ_FD]);
     close(impl->cross_thread_signal_pipe[WRITE_FD]);
     close(impl->kq_fd);
-    aws_thread_clean_up(&impl->thread);
+    aws_thread_clean_up(&impl->thread_created_on);
     aws_mem_release(event_loop->alloc, impl);
     aws_event_loop_clean_up_base(event_loop);
     aws_mem_release(event_loop->alloc, event_loop);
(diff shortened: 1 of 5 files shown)


nchong-at-aws:
This PR depends on https://github.com/awslabs/aws-c-common/pull/563

source/bsd/kqueue_event_loop.c
```diff
@@ -798,7 +799,7 @@ static void s_event_thread_main(void *user_data) {
     struct kqueue_loop *impl = event_loop->impl_data;
 
     /* set thread id to the event-loop's thread. */
-    aws_atomic_store_int(&impl->thread_id, (size_t)aws_thread_current_thread_id());
+    aws_atomic_store_ptr(&impl->thread_id, &impl->thread.thread_id);
```
nchong-at-aws: This could also use `aws_atomic_exchange_ptr` with an expected `NULL` value as an additional check that `run` and `stop; wait` calls are interleaved properly.

source/bsd/kqueue_event_loop.c
```diff
@@ -297,7 +297,8 @@ static void s_destroy(struct aws_event_loop *event_loop) {
         return;
     }
     /* setting this so that canceled tasks don't blow up when asking if they're on the event-loop thread. */
-    aws_atomic_store_int(&impl->thread_id, (size_t)aws_thread_current_thread_id());
+    aws_thread_id current_thread_id = aws_thread_current_thread_id();
+    aws_atomic_store_ptr(&impl->thread_id, &current_thread_id);
```
danielsn: This seems dangerous, storing the pointer to a stack variable somewhere

source/bsd/kqueue_event_loop.c
```diff
@@ -711,8 +712,8 @@ static int s_unsubscribe_from_io_events(struct aws_event_loop *event_loop, struc
 static bool s_is_event_thread(struct aws_event_loop *event_loop) {
     struct kqueue_loop *impl = event_loop->impl_data;
 
-    uint64_t thread_id = aws_atomic_load_int(&impl->thread_id);
-    return aws_thread_current_thread_id() == thread_id;
+    aws_thread_id *thread_id = aws_atomic_load_ptr(&impl->thread_id);
+    return thread_id && aws_thread_thread_id_equal(*thread_id, aws_thread_current_thread_id());
```
danielsn: What is the semantics of a `null` pointer here.  Should this be an assert?

nchong-at-aws: `NULL` is ok. For example:

```c
// in client:
struct aws_event_loop *el = aws_event_loop_new_default(...); //< ptr set to NULL
bool b = aws_event_loop_thread_is_callers_thread(el); //< okay to call this even though event loop is not running so will return false
```

source/bsd/kqueue_event_loop.c
nchong-at-aws: In this case it is safe because the event loop is being destroyed. The pointer will be free'd by `aws_mem_release(event_loop->alloc, impl)` after the task scheduler has been flushed.

The alternatives we discussed:
  - Use an `AWS_THREAD_LOCAL aws_thread_id` variable
  - Reset pointer to NULL after task scheduler has been flushed

graebm: It's not safe due to how the atomic is accessed in `s_is_event_thread()`

In that function, it loads the atomic (a pointer), then accesses the value that it's pointing to. If, in the moment between loading the atomic and following the pointer, the thread is deleted, then we will access invalid memory when we follow the pointer.

nchong-at-aws: Ah yes, I see the problem. This means thread-local storage won't work either since the race is possible there too.

nchong-at-aws: Hi both, the last change no longer takes a pointer to a stack-local variable. Instead, we re-use the variable `event_loop->impl.thread.thread_id`. At this point of the execution, the event loop thread has been joined to the current thread (by `wait_for_stop_completion()`) so we can reuse the var.

Regarding the potential race if one thread is calling `destroy()` concurrently with another calling `is_on_callers_thread()`. I think this is a user error. If thread T is calling `destroy()` on an event loop then all other threads should have given up their right to call event loop functions before T calls destroy. Otherwise, there is the danger that these calls will encounter a free'd event loop object.

What if we add the following text to the API call `aws_event_loop_destroy`?

> If the event loop is shared by multiple threads then destroy must be called by exactly one thread. All other threads must ensure their API calls to the event loop happen-before the call to destroy.

source/linux/epoll_event_loop.c
```diff
@@ -545,7 +546,7 @@ static void s_main_loop(void *args) {
     struct epoll_loop *epoll_loop = event_loop->impl_data;
 
     /* set thread id to the thread of the event loop */
-    aws_atomic_store_int(&epoll_loop->thread_id, (size_t)aws_thread_current_thread_id());
+    aws_atomic_store_ptr(&epoll_loop->thread_id, &epoll_loop->thread.thread_id);
```
danielsn: I'm confused by having something called `x.thread_id` and something else called `x.thread.thread_id`. What's the meaning of the two

JonathanHenson: typically nested structs indicate something that needs a memory barrier.

danielsn: The proposal: rename the fields
```
struct aws_thread thread_created_on;
struct aws_thread_id_t thread_currently_running_on;
```

source/bsd/kqueue_event_loop.c
danielsn: Can we check for this at runtime and throw an error?

source/linux/epoll_event_loop.c
graebm: `x.thread` is a `struct aws_thread`. He's directly accessing its `thread_id` member instead of using `aws_thread_get_id(&epoll_loop->thread)`, which is a *bit* dirty because we don't typically access the members of our structs directly.

`x.thread_id` is an atomic that **points** to `x.thread.thread_id`. We could get rid of the atomic `x.thread_id` if we held a lock when querying `aws_thread_get_id(&epoll_loop->thread)`. But the way the atomics are set up here, it's a sneaky efficiency trick that I'm pretty sure works.

We could have **another** variable to avoid accessing `x.thread.thread_id` directly.
```
struct epoll_loop {
    ...
    struct aws_thread thread;
    aws_thread_id_t thread_id; // copy of thread's thread_id
    struct aws_atomic_var loop_thread_id_ptr; // points to thread_id
    ...
};
```

source/bsd/kqueue_event_loop.c
nchong-at-aws: @danielsn We could set an atomic flag at the top of destroy() and then check against that in all API calls. This wouldn't catch all misuses, but could be useful in debug builds.

Resolving the original request.

source/bsd/kqueue_event_loop.c
nchong-at-aws: Suggestion, but not implemented.

source/linux/epoll_event_loop.c
nchong-at-aws: Variable names should be clearer now.

@JonathanHenson: right, this is perhaps not as clean as it could be. I'm peeking inside a struct member. I think we're okay doing this (from a thread-safety pov) but will double-check. We could do as @graebm suggests and have a separate variable which we initialize with `aws_thread_current_thread_id()`.

JonathanHenson:
Merging as an Admin because our CI jobs for github actions don't work with a forked PR, and our downstream builder checks "master" from s2n which is currently broken for gcc 4.

PR #219, opened 2019-12-31
Dns

bretambrose:
First draft of a proposed public interface for DNS and host resolution that includes full resolver implementations and separation of concerns between resolvers and CRT-desired functionality.


include/aws/io/dns.h
```diff
@@ -0,0 +1,602 @@
...
+ * Vocabulary/Concepts
+ *
+ * aws_dns_resolver - A thin interface around remote DNS queries.  Includes a default implementation that supports
+ * recursive and iterative queries.  Supports DNS over UDP.  (Eventually) Supports DNS over https and tls.  No support
+ * for non-internet-class queries.  Uses a vtable and can be replaced with a custom implementation.
+ *
+ * aws_dns_service_provider_config - Controls what host/port/protocol combos the default aws_dns_resolver uses to
+ * resolve queries.
+ *
+ * aws_host_resolution_service - A heavy-weight host name resolution service that contains CRT-specific and
```
JonathanHenson: why not just call this aws_dns_caching_resolver or something? 

or aws_dns_cache

include/aws/io/dns.h
```diff
@@ -0,0 +1,602 @@
...
+ * A thin interface around raw dns query/response.  A default resolver is included that uses
+ * aws_dns_service_provider_config to control what remote dns services are used.
+ *
+ * Musts/Requirements:
+ *   (1) 1 question allowed per query (aggregation is caller's responsibility)
+ *   (2) Implements retries (users should not layer retry on top)
+ *
+ * Shoulds:
+ *   (1) Cache authority/zones-of-responsibility in a label-based tree (iterative only?)
+ *   (2) Assume query input is a host name and not an IP address.
```
JonathanHenson: would this break existing systems that rely on querying an ip address just returning that ip address?

include/aws/io/dns.h
```diff
@@ -0,0 +1,602 @@
...
+    AWS_DNS_RR_DLV = 32769,
+
+};
+
+struct aws_dns_resource_record {
+    /* Needs to be on the record (rather than the full result/set) to support ANY-based queries */
+    enum aws_dns_resource_record_type type;
+
+    /* time-to-live in seconds */
+    uint32_t ttl;
```
JonathanHenson: should we store time fetched or something so we can evaluate the TTL?

PR #226, opened 2020-01-16
Fix ThreadSanitizer bugs

nchong-at-aws:
Issue #, if available: 220

Description of changes:

- Fix missing lock in channel_message_passing
- Fix mismatched lock/unlock in event_loop_test
- Fix racy reuse of aws_task in channel_test
- Make following variables atomic
   - keep_active in host_resolver
   - ~~last_use in host_resolver~~ this used to be uint64_t, scared to make it size_t without understanding the ramifications
   - shutdown_called in read_write_test_handler
   - shutdown_completed in channel_test
   - rw_slot in socker_handler_test

### DIFF
--- a/source/host_resolver.c
+++ b/source/host_resolver.c
@@ -120,10 +120,6 @@ struct host_entry {
     struct aws_host_resolution_config resolution_config;
     struct aws_linked_list pending_resolution_callbacks;
     int64_t resolve_frequency_ns;
-    /* this member will be a monotonic increasing value and not protected by a memory barrier.
-       Let it tear on 32-bit systems, we don't care, we just want to see a change. This at least assumes cache coherency
-       for the target architecture, which these days is a fairly safe assumption. Where it's not a safe assumption, we
-       probably don't have multiple cores available anyways. */
     struct aws_atomic_var last_use;
     struct aws_atomic_var keep_active;
 };
--- a/source/host_resolver.c
+++ b/source/host_resolver.c
@@ -14,6 +14,7 @@
  */
 #include <aws/io/host_resolver.h>
 
+#include <aws/common/atomics.h>
 #include <aws/common/clock.h>
 #include <aws/common/condition_variable.h>
 #include <aws/common/hash_table.h>
@@ -123,8 +124,8 @@ struct host_entry {
        Let it tear on 32-bit systems, we don't care, we just want to see a change. This at least assumes cache coherency
        for the target architecture, which these days is a fairly safe assumption. Where it's not a safe assumption, we
        probably don't have multiple cores available anyways. */
-    volatile uint64_t last_use;
-    volatile bool keep_active;
+    struct aws_atomic_var last_use;
+    struct aws_atomic_var keep_active;
 };
 
 static int resolver_purge_cache(struct aws_host_resolver *resolver) {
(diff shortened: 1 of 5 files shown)


source/host_resolver.c
```diff
@@ -123,8 +124,8 @@ struct host_entry {
        Let it tear on 32-bit systems, we don't care, we just want to see a change. This at least assumes cache coherency
        for the target architecture, which these days is a fairly safe assumption. Where it's not a safe assumption, we
        probably don't have multiple cores available anyways. */
-    volatile uint64_t last_use;
-    volatile bool keep_active;
+    struct aws_atomic_var last_use;
+    struct aws_atomic_var keep_active;
```
graebm: Comment can be deleted, now that code is good

tests/channel_test.c
```diff
@@ -206,18 +206,17 @@ static void s_wait_a_bit_task(struct aws_task *task, void *arg, enum aws_task_st
...
-    aws_task_init(&task, s_wait_a_bit_task, &cv, "wait_a_bit");
+    aws_task_init(task, s_wait_a_bit_task, &cv, "wait_a_bit");
 
     uint64_t run_at_ns;
     ASSERT_SUCCESS(aws_event_loop_current_clock_time(loop, &run_at_ns));
     run_at_ns += aws_timestamp_convert(1, AWS_TIMESTAMP_SECS, AWS_TIMESTAMP_NANOS, NULL);
-    aws_event_loop_schedule_task_future(loop, &task, run_at_ns);
+    aws_event_loop_schedule_task_future(loop, task, run_at_ns);
 
     return aws_condition_variable_wait(&cv, &mutex);
```
graebm: we talked about this.
intention of helper function is to wait for task-queue to flush
bug is: spurious wake-ups of condition variables could cause the function to exit before task actually executed
simpler implementation would be spin-locking on an atomic bool

nchong-at-aws:
Windows builds have the following warning/error:

```
read_write_test_handler.c(302):
warning C4267: 'return': conversion from 'size_t' to 'int', possible loss of data 
```

The variable `shutdown_error` is accessed by multiple threads so should be atomic. The problem is that the variable is an int (an error code) and `aws_atomic_load_int` returns `size_t`. Possible workarounds:

- Change `struct aws_atomic_var shutdown_error` into `_Atomic int shutdown_error`. Fixes the bug report from TSan. Fixes MSVC warning. But is not C99.
- Add `pragma warning(disable : 4267)`. Fixes the bug report from TSan. Hides MSVC warning.
- Revert change and add `__attribute__((no_sanitize("thread")))` to hide the bug report from TSan. Fixes MSVC warning.

source/host_resolver.c
nchong-at-aws: This is trickier than I initially thought. `aws_atomic_var` gives a `size_t` so there are truncation issues with this change. I've reverted for now (so TSan will give a warning here). Long term, if the resolver only need to be aware of changes then this could be implemented with a 32-bit atomic counter that is incremented to avoid the problem of a 64-bit atomic value on a 32-bit machine. Or, alternatively, protecting `last_use` with a mutex.

nchong-at-aws:
Fixes all but one TSan bug report. Remaining issue is use of host resolver last_use variable (should be atomic but tricky because it is 64-bit).

PR #229, opened 2020-01-22
Fix more tsan bugs

justinboswell:

### DIFF
--- a/tests/channel_test.c
+++ b/tests/channel_test.c
@@ -30,8 +30,9 @@
 struct channel_setup_test_args {
     struct aws_mutex mutex;
     struct aws_condition_variable condition_variable;
-    bool shutdown_completed;
-    int error_code;
+    struct aws_atomic_var setup_completed;
+    struct aws_atomic_var shutdown_completed;
+    int error_code; /* protected by mutex */
     enum aws_task_status task_status;
 };
 
@@ -41,8 +42,32 @@ static void s_channel_setup_test_on_setup_completed(struct aws_channel *channel,
 
     aws_mutex_lock(&setup_test_args->mutex);
     setup_test_args->error_code |= error_code;
-    aws_condition_variable_notify_one(&setup_test_args->condition_variable);
     aws_mutex_unlock(&setup_test_args->mutex);
+    aws_atomic_store_int(&setup_test_args->setup_completed, true);
+    aws_condition_variable_notify_one(&setup_test_args->condition_variable);
+}
+
+static bool s_channel_setup_test_setup_completed_predicate(void *arg) {
+    struct channel_setup_test_args *setup_test_args = (struct channel_setup_test_args *)arg;
+    return aws_atomic_load_int(&setup_test_args->setup_completed);
+}
+
+/* Create a new channel and wait until its setup completes */
+static int s_channel_setup_create_and_wait(
+    struct aws_allocator *allocator,
+    struct aws_event_loop *event_loop,
+    struct aws_channel_creation_callbacks *callbacks,
+    struct channel_setup_test_args *test_args,
+    struct aws_channel **returned_channel) {
+    ASSERT_NULL(*returned_channel);
+    *returned_channel = aws_channel_new(allocator, event_loop, callbacks);
+    ASSERT_NOT_NULL(*returned_channel);
+    ASSERT_SUCCESS(aws_mutex_lock(&test_args->mutex));
+    ASSERT_SUCCESS(aws_condition_variable_wait_pred(
+        &test_args->condition_variable, &test_args->mutex, s_channel_setup_test_setup_completed_predicate, test_args));
+    ASSERT_INT_EQUALS(0, test_args->error_code);
+    ASSERT_SUCCESS(aws_mutex_unlock(&test_args->mutex));
+    return AWS_OP_SUCCESS;
 }
 
 static int s_test_channel_setup(struct aws_allocator *allocator, void *ctx) {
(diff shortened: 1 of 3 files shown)


tests/channel_test.c
```diff
@@ -239,15 +239,16 @@ static int s_test_channel_refcount(struct aws_allocator *allocator, void *ctx) {
     struct aws_channel_creation_callbacks callbacks = {
         .on_setup_completed = s_channel_setup_test_on_setup_completed,
         .setup_user_data = &test_args,
-        .on_shutdown_completed = s_channel_setup_test_on_setup_completed,
-        .shutdown_user_data = &test_args,
+        .on_shutdown_completed = NULL,
```
nchong-at-aws: Looks like a copy-past error. I assumed waiting on shutdown being complete isn't needed in this test so made them NULL.

tests/channel_test.c
```diff
@@ -239,15 +239,16 @@ static int s_test_channel_refcount(struct aws_allocator *allocator, void *ctx) {
...
+        .shutdown_user_data = NULL,
     };
 
-    ASSERT_SUCCESS(aws_mutex_lock(&test_args.mutex));
     struct aws_channel *channel = aws_channel_new(allocator, event_loop, &callbacks);
     ASSERT_NOT_NULL(channel);
 
+    ASSERT_SUCCESS(aws_mutex_lock(&test_args.mutex));
     ASSERT_SUCCESS(aws_condition_variable_wait(&test_args.condition_variable, &test_args.mutex));
+    ASSERT_SUCCESS(aws_mutex_unlock(&test_args.mutex));
```
nchong-at-aws: Should unlock after condition wait.

tests/channel_test.c
```diff
@@ -281,8 +280,10 @@ static int s_test_channel_refcount(struct aws_allocator *allocator, void *ctx) {
 
     /* Release hold 2/2. The handler and channel should be destroyed. */
     aws_channel_release_hold(channel);
+    ASSERT_SUCCESS(aws_mutex_lock(&destroy_mutex));
```
nchong-at-aws: Reduce size of the critical section by moving this lock down.

tests/event_loop_test.c
```diff
@@ -31,7 +32,7 @@ struct task_args {
     enum aws_task_status status;
     struct aws_mutex mutex;
     struct aws_condition_variable condition_variable;
-    bool thread_complete;
+    struct aws_atomic_var thread_complete;
```
nchong-at-aws: Should be atomic since it is written-by the cleanup thread and read-by the main thread.

tests/channel_test.c
```diff
@@ -281,8 +280,10 @@ static int s_test_channel_refcount(struct aws_allocator *allocator, void *ctx) {
...
     ASSERT_SUCCESS(aws_condition_variable_wait(&destroy_condition_variable, &destroy_mutex));
     ASSERT_TRUE(aws_atomic_load_int(&destroy_called));
+    ASSERT_SUCCESS(aws_mutex_unlock(&destroy_mutex));
```
graebm: outside the scope of fixing immediate tsan errors, but:

Should we replace all uses of `aws_condition_variable_wait()` with `aws_condition_variable_wait_pred()`? I would even vote for **completely removing the vanilla wait() call from aws-c-common.** These tests are only using vanilla wait() because it required less code to get working than wait_pred(). But vanilla wait() can spuriously wake up and break the assumptions of the test.

From the [pthread docs](https://linux.die.net/man/3/pthread_cond_wait):
> When using condition variables there is always a Boolean predicate involving shared variables associated with each condition wait that is true if the thread should proceed. Spurious wakeups from the pthread_cond_timedwait() or pthread_cond_wait() functions may occur. Since the return from pthread_cond_timedwait() or pthread_cond_wait() does not imply anything about the value of this predicate, the predicate should be re-evaluated upon such return.

tests/event_loop_test.c
```diff
@@ -1070,7 +1071,7 @@ AWS_TEST_CASE(event_loop_group_setup_and_shutdown, test_event_loop_group_setup_a
 /* mark the thread complete when the async shutdown thread is done */
 static void s_async_shutdown_thread_exit(void *user_data) {
     struct task_args *args = user_data;
-    args->thread_complete = true;
+    aws_atomic_store_int(&args->thread_complete, true);
```
JonathanHenson: the lock is always acquired in the predicate, and unlocked during wait. why does this need to be atomic? You can just lock in the callback.

tests/channel_test.c
graebm: I don't think this mutex/condition variable need to exist anymore. We can just loop until destroy_called is true.

Before our 1st pass of tsan fixes, this test was broken because it didn't acquire the lock before changing the bool and signaling the condvar.

We changed the bool to be atomic. We're  using the mutex when waiting for the value and **not** using wait_pred(). But we're **not** using the mutex when changing the value and signaling the condvar. That means, if we set the bool and signal **before** the main thread waits for it, the main thread will never be woken up (except for spurious wakeups).

So a fully proper solution is either:
1) use an atomic, main thread just checks in in a while loop. No mutex or condvar required.
2) OR use the lock AND the condition variable when setting the bool, use a predicate when waiting for the bool. It's not necessary for the bool to be atomic at this point.

First off, I think this test had been timing dependent and could theoretically fail
they weren't using predicates so the

tests/channel_test.c
JonathanHenson: I don’t necessarily want to remove it, but we should definitely not be using it in tests. This particular test was written prior to us adding the predicate variant.

tests/event_loop_test.c
nchong-at-aws: It's an ordering violation: we need to ensure the cleanup thread exit happens-before the test finishes. A lock by itself won't do. Do you mean use a condition variable? That would work, but this use of an atomic is simpler, I think.

tests/channel_test.c
nchong-at-aws: I see. Yes agreed on the pathological case. I'll fix with (1) as we can remove some code with that option.

tests/event_loop_test.c
nchong-at-aws: The data-race / ordering violation is:
- cleanup thread launched and detached
- main thread exits and frees allocator
- cleanup thread tries to use allocator to free exit_callback_data

tests/channel_test.c
nchong-at-aws: Yes, replace these calls with their `_pred` equivalents in the tests. Also, document this behavior for future users: https://github.com/awslabs/aws-c-common/issues/580

nchong-at-aws:
Travis failure needs investigating. Looks like `channel_refcount_delays_clean_up` isn't terminating. Also, the following for host_resolver: `test_resolver_ttls`, `test_resolver_connect_failure_recording`, `test_resolver_ttl_refreshes_on_resolve` (could all be same root issue).

tests/channel_test.c
nchong-at-aws: Actually, thinking about this further, it is ok to keep the bool atomic and the condition variable if the main thread does a check with `aws_condition_variable_wait_pred`. In this case, the thread executing  destroy atomically updates the bool and signals the condition variable (no lock necessary). The main thread locks the mutex and calls `aws_condition_variable_wait_pred` with a predicate that atomically loads from the bool. The spurious wakeup case is purely due to the use of `aws_condition_variable_wait`.

PR #241, opened 2020-02-25
Github action for address+undef sanitizers

nchong-at-aws:


.github/workflows/clang-asan-ubsan.yml
```diff
@@ -0,0 +1,26 @@
...
+    - name: Install Dependencies
+      run: |
+        sudo apt-get install ninja-build
+      
+    - name: Build and test
+      run: |
+        ./build-deps.sh -DCMAKE_C_COMPILER=clang -DENABLE_SANITIZERS=ON -DSANITIZERS=",address,undefined"
+        cd build
+        cmake -GNinja -DCMAKE_C_COMPILER=clang -DENABLE_SANITIZERS=ON -DSANITIZERS=",address,undefined" -DCMAKE_PREFIX_PATH=`pwd`/deps/install/ ..
+        ninja
```
graebm: trivial: Do we need `ninja` for this? We can just let cmake use whatever build system in on the machine, and issue the build command like so:
`cmake --build .`

.github/workflows/clang-asan-ubsan.yml
```diff
@@ -0,0 +1,26 @@
...
+        cd tests
+        # always return 0 (ok) until we've squashed all bugs
+        ctest --verbose || true
```
graebm: Is it worth committing this if it doesn't do anything?

nchong-at-aws: There's a test that fails asan because it has a leak: test_concurrent_cert_import. We do something similar for the threadsan action (because of the host_resolver failures). Alternatively, we could fix the leak and then add this action (and make it blocking?).

.github/workflows/clang-asan-ubsan.yml
justinboswell: We could also just omit the -GNinja and just use `cmake --build .`

bretambrose (approved):

PR #253, opened 2020-03-17
No longer both importing and depending on s2n

justinboswell:
Requires an update to v0.6.0 of the builder

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,8 +7,8 @@ on:
       - '!master'
 
 env:
-  BUILDER_VERSION: more-fixes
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -7,7 +7,8 @@ on:
       - '!master'
 
 env:
-  BUILDER_VERSION: v0.5.9
+  BUILDER_VERSION: more-fixes
+  BUILDER_SOURCE: channels
```
graebm: is this something that needs to change before merge?

justinboswell: yes

PR #263, opened 2020-04-03
Windows TLS Error Fix

rccarper:
*Description of changes:*
Receiving a security status of SEC_I_CONTEXT_EXPIRED from DecryptMessage on Windows is the same case as a zero length read here in s2n's s_s2n_handler_process_read_message.  (https://github.com/awslabs/aws-c-io/blob/master/source/s2n/s2n_tls_channel_handler.c#L514)  Both should result in a channel shutdown with success, but Windows currently raise an error.  In IoT Land, this causes a TLS Negotiation Error on Windows, and a more correct Unexpected MQTT Hangup on platforms using s2n.  Changing Windows behavior here to match.

Side note: originally, there was talk of adding an additional line to s_determine_sspi_error to check for SEC_I_CONTEXT_EXPIRED, but because this case is not actually an error, and an error does not need to be raised, I'm not currently doing that.  (It's possible I was misunderstanding the intent there, in which case, I can add it there as well.)


source/windows/secure_channel_tls_handler.c
```diff
@@ -1039,9 +1039,22 @@ static int s_do_application_data_decrypt(struct aws_channel_handler *handler) {
                 read_len);
             sc_handler->buffered_read_in_data_buf.len = read_len;
             aws_raise_error(AWS_IO_READ_WOULD_BLOCK);
+        }
+        /* SEC_I_CONTEXT_EXPIRED means that the message sender has shut down the connection.  One such case
+           where this can happen is an unaccepted certificate. */
+        else if (status == SEC_I_CONTEXT_EXPIRED) {
+            AWS_LOGF_TRACE(
+                AWS_LS_IO_TLS,
+                "id=%p: Message sender has shut down the connection. SECURITY_STATUS is %d.",
```
JonathanHenson, line 1048: If you can find a way to get the alert code and log it that would be super helpful, if you can’t at least log an alert was recieved.

PR #264, opened 2020-04-08
Retry strategy interface with exponential backoff implementation. AIM…

JonathanHenson:
…D will be in a later PR.

This is the doc we're basing this on:

https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/


include/aws/io/retry_strategy.h
```diff
@@ -0,0 +1,175 @@
...
+
+struct aws_retry_strategy;
+struct aws_retry_token;
+
+/**
+ * Invoked upon the acquisition, or failure to acquire a retry token. This function will always be invoked if and only
+ * if aws_retry_strategy_acquire_retry_token() returns AWS_OP_SUCCESS. It will never be invoked synchronously from
+ * aws_retry_strategy_acquire_retry_token().
+ */
+typedef void(aws_retry_strategy_on_retry_token_acquired_fn)(
```
graebm, line 29: [clarification] token will always be NULL if error_code is non-zero, and vice-versa, right?
If so mention that in comments

include/aws/io/retry_strategy.h
```diff
@@ -0,0 +1,175 @@
...
+    AWS_EXPONENTIAL_BACKOFF_JITTER_DECORRELATED,
+    /* Uses AWS_EXPONENTIAL_BACKOFF_JITTER_FULL */
+    AWS_EXPONENTIAL_BACKOFF_JITTER_DEFAULT,
+};
+
+struct aws_exponential_backoff_retry_config {
+    /** Event loop group to use for scheduling tasks. */
+    struct aws_event_loop_group *el_group;
+    /** Max retries to allow. This value must be greater than 0 */
+    size_t max_retries;
```
graebm, line 103: if value must be greater than 0, shouldn't it be `max_tries`?

include/aws/io/retry_strategy.h
```diff
@@ -0,0 +1,175 @@
...
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+#include <aws/io/event_loop.h>
```
graebm, line 17: trivial: I think you could get away with fwd declares here

include/aws/io/retry_strategy.h
```diff
@@ -0,0 +1,175 @@
...
+ * Invoked after a successful call to aws_retry_strategy_schedule_retry(). This function will always be invoked if and
+ * only if aws_retry_strategy_schedule_retry() returns AWS_OP_SUCCESS. It will never be invoked synchronously from
+ * aws_retry_strategy_schedule_retry().
+ */
+typedef void(aws_retry_strategy_on_retry_ready_fn)(struct aws_retry_token *token, int error_code, void *user_data);
+
+enum aws_retry_error_type {
+    /** This is a connection level error such as a socket timeout, socket connect error, tls negotiation timeout etc...
+     * Typically these should never be applied for non-idempotent request types. */
+    AWS_RETRY_ERROR_TYPE_TRANSIENT,
```
graebm, line 45: [clarification] Not sure what you mean by "Typically these should never be applied for non-idempotent request types"

Do you mean that TRANSIENT is the only type where you're not sure whether the request/response completed? So it's the only one that's not safe for retrying non-idempotent actions?

include/aws/io/retry_strategy.h
```diff
@@ -0,0 +1,175 @@
+#ifndef AWS_IO_CLIENT_RETRY_STRATEGY_H
+#define AWS_IO_CLIENT_RETRY_STRATEGY_H
```
graebm, line 2: This seems like a better fit for `aws-c-ops` than `aws-c-io`

include/aws/io/retry_strategy.h
```diff
@@ -0,0 +1,175 @@
...
+    /** By default this will be set to use aws_device_random. If you want something else, set it here. */
+    uint64_t (*generate_random)(void);
+};
+
+/**
+ * Defaults to 10 retries, 25ms scaling factor, and full jitter.
+ */
+#define AWS_EXPONENTIAL_BACKOFF_DEFAULT_CONFIG(el_group)                                                               \
+    /* Don't do this with designated initializers, otherwise C++ users can't touch this. */                            \
+    { el_group, 10, 25, AWS_EXPONENTIAL_BACKOFF_JITTER_FULL, }
```
graebm, line 117: debatable/trivial: could we just treat 0 as "defaults pls" for everything in the config?
but that strategy falls apart if 0 is a valid value for something

include/aws/io/retry_strategy.h
```diff
@@ -0,0 +1,175 @@
...
+    struct aws_allocator *allocator;
+    struct aws_retry_strategy_vtable *vtable;
+    struct aws_atomic_var ref_count;
+    void *impl;
+};
+
+struct aws_retry_token {
+    struct aws_allocator *allocator;
+    struct aws_retry_strategy *retry_strategy;
+    void *impl;
```
graebm, line 82: BIG IDEA: what if the token was just a `int` instead of an allocated type?
or just a copy-by-value type like
```
struct aws_retry_token {
    struct aws_retry_strategy *owner;
    int id;
}
```

then we wouldn't need to worry about how they're refcounted/released, or how it's bound in a GC language where someone keeps around a pointer to the RetryToken even after after they've "given it back" to the strategy

any state about token could just be stored within the strategy and looked up by hashtable

include/aws/io/retry_strategy.h
```diff
@@ -0,0 +1,175 @@
...
+ */
+enum aws_exponential_backoff_jitter_mode {
+    AWS_EXPONENTIAL_BACKOFF_JITTER_NONE,
+    AWS_EXPONENTIAL_BACKOFF_JITTER_FULL,
+    AWS_EXPONENTIAL_BACKOFF_JITTER_DECORRELATED,
+    /* Uses AWS_EXPONENTIAL_BACKOFF_JITTER_FULL */
+    AWS_EXPONENTIAL_BACKOFF_JITTER_DEFAULT,
+};
+
+struct aws_exponential_backoff_retry_config {
```
graebm, line 99: trivial: it's slightly more common to name structs like this "_options".
ex: aws_channel_options, aws_tls_options, aws_http_client_connection_options, aws_http_proxy_options, ...

include/aws/io/retry_strategy.h
```diff
@@ -0,0 +1,175 @@
...
+ *
+ * For a great writeup on these options see:
+ * https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
+ */
+enum aws_exponential_backoff_jitter_mode {
+    AWS_EXPONENTIAL_BACKOFF_JITTER_NONE,
+    AWS_EXPONENTIAL_BACKOFF_JITTER_FULL,
+    AWS_EXPONENTIAL_BACKOFF_JITTER_DECORRELATED,
+    /* Uses AWS_EXPONENTIAL_BACKOFF_JITTER_FULL */
+    AWS_EXPONENTIAL_BACKOFF_JITTER_DEFAULT,
```
graebm, line 96: trivial: consider DEFAULT having the value of 0

include/aws/io/retry_strategy.h
JonathanHenson: aws-c-ops doesn't exist, and after thinking more on it, it's a necessary mechanism for any protocol development. Seems to make sense here to me. Http clients can use it, mqtt could technically etc....

include/aws/io/retry_strategy.h
JonathanHenson: yes, you're right, will fix.

include/aws/io/retry_strategy.h
JonathanHenson: will do.

include/aws/io/retry_strategy.h
JonathanHenson: yes.

include/aws/io/retry_strategy.h
JonathanHenson: I"m going to think  a bit more on this one. I like the flexibility of having a void *impl on these for future use and expansion. I'm not fond of the threading implications of having to maintain a map for somethings that could be stateless.

include/aws/io/retry_strategy.h
JonathanHenson: will do.

include/aws/io/retry_strategy.h
JonathanHenson: sure, will do.

include/aws/io/retry_strategy.h
JonathanHenson: what would be the point of a retry strategy that doesn't do any retries though?

include/aws/io/retry_strategy.h
JonathanHenson: we could.

include/aws/io/retry_strategy.h
graebm: This just seems like the first of N misc parts we'll need to build a full SDK on the CRT that aren't big enough (like aws-c-auth) to warrant their own package.

What other parts will we be building in the next few months that **should** go into aws-c-ops vs aws-c-io?

JonathanHenson: I think the line would be things that need more info/aggregation info than lives in aws-c-http, aws-c-io etc... So things like throttling, max conns, monitoring, metrics, etc.... I don't think aws-c-ops should just be the grab bag of misc components, but stuff that properly doesn't fit into the dependency graph as is. In this case the strategy interface doesn't really need more info than aws-c-io already provides. And any protocol stack can consume it as is. Another example would be, should this error result in a connection closure or reuse? The would need metadata that neither aws-c-io or aws-c-http contain.

source/exponential_backoff_retry_strategy.c
```diff
@@ -0,0 +1,324 @@
...
+static void s_exponential_retry_destroy(struct aws_retry_strategy *retry_strategy) {
+    if (retry_strategy) {
+        aws_mem_release(retry_strategy->allocator, retry_strategy);
+    }
+}
+
+static void s_exponential_retry_task(struct aws_task *task, void *arg, enum aws_task_status status) {
+    (void)task;
+
+    int error_code = AWS_ERROR_IO_OPERATION_CANCELLED;
```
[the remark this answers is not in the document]
JonathanHenson: well retries live longer than a channel, a retry is likely on a new channel. The only cancelation we need to worry about here is the event-loop shutting down. I'm certainly not opposed to ref counting it though.

source/exponential_backoff_retry_strategy.c
```diff
@@ -0,0 +1,324 @@
...
+        (void *)retry_strategy,
+        (void *)&backoff_retry_token->base);
+
+    backoff_retry_token->base.allocator = retry_strategy->allocator;
+    backoff_retry_token->base.retry_strategy = retry_strategy;
+    aws_retry_strategy_acquire(retry_strategy);
+    backoff_retry_token->base.impl = backoff_retry_token;
+
+    struct exponential_backoff_strategy *exponential_backoff_strategy = retry_strategy->impl;
+    backoff_retry_token->bound_loop = aws_event_loop_group_get_next_loop(exponential_backoff_strategy->config.el_group);
```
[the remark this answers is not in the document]
JonathanHenson: It was just to avoid constantly jumping threads, though arguably you'd be doing that anyways the second you jumped onto an http connection.

source/exponential_backoff_retry_strategy.c
```diff
@@ -0,0 +1,324 @@
...
+
+    uint64_t random = token->generate_random();
+    return min + random % (diff);
+}
+
+typedef uint64_t(compute_backoff_fn)(struct exponential_backoff_retry_token *token);
+
+static uint64_t s_compute_no_jitter(struct exponential_backoff_retry_token *token) {
+    size_t retry_count = aws_atomic_load_int(&token->current_retry_count);
+    return ((uint64_t)1 << retry_count) * token->backofff_scale_factor;
```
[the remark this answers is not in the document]
JonathanHenson: good catch.

source/exponential_backoff_retry_strategy.c
```diff
@@ -0,0 +1,324 @@
...
+}
+
+static uint64_t s_compute_deccorelated_jitter(struct exponential_backoff_retry_token *token) {
+    size_t last_backoff_val = aws_atomic_load_int(&token->last_backoff);
+
+    if (!last_backoff_val) {
+        return s_compute_full_jitter(token);
+    }
+
+    return s_random_in_range(token->backofff_scale_factor, last_backoff_val * 3, token);
```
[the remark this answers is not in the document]
JonathanHenson: good catch.

source/exponential_backoff_retry_strategy.c
```diff
@@ -0,0 +1,324 @@
...
+                s_exponential_retry_task,
+                backoff_retry_token,
+                "aws_exponential_backoff_retry_task");
+        }
+        AWS_FATAL_ASSERT(!aws_mutex_unlock(&backoff_retry_token->thread_data.mutex) && "Mutex lock release failed");
+    } /**** END CRITICAL SECTION ***********/
+
+    AWS_FATAL_ASSERT(
+        !already_scheduled && "There was already a retry task scheduled. This is a programming error."
+                              " Wait for retry callback before attempting another.");
```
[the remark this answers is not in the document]
JonathanHenson: okay

source/exponential_backoff_retry_strategy.c
```diff
@@ -0,0 +1,324 @@
...
+        error_type);
+    uint64_t schedule_at = 0;
+
+    /* AWS_RETRY_ERROR_TYPE_CLIENT_ERROR does not count against your retry budget since you were responding to an
+     * improperly crafted request. */
+    if (error_type != AWS_RETRY_ERROR_TYPE_CLIENT_ERROR) {
+        size_t retry_count = aws_atomic_load_int(&backoff_retry_token->current_retry_count);
+
+        if (retry_count >= backoff_retry_token->max_retries) {
+            return aws_raise_error(AWS_IO_MAX_RETRIES_EXCEEDED);
```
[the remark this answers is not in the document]
JonathanHenson: yep, will do.

source/retry_strategy.c
```diff
@@ -0,0 +1,66 @@
...
+int aws_retry_strategy_token_record_success(struct aws_retry_token *token) {
+    AWS_PRECONDITION(token);
+    AWS_PRECONDITION(token->retry_strategy);
+    AWS_PRECONDITION(token->retry_strategy->vtable->record_success);
+
+    return token->retry_strategy->vtable->record_success(token);
+}
+
+void aws_retry_strategy_release_retry_token(struct aws_retry_token *token) {
+    AWS_PRECONDITION(token);
```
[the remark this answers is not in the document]
JonathanHenson: will update.

source/exponential_backoff_retry_strategy.c
```diff
@@ -0,0 +1,322 @@
...
+    struct aws_retry_strategy base;
+    struct aws_exponential_backoff_retry_config config;
+};
+
+struct exponential_backoff_retry_token {
+    struct aws_retry_token base;
+    struct aws_atomic_var current_retry_count;
+    struct aws_atomic_var last_backoff;
+    size_t max_retries;
+    uint64_t backofff_scale_factor;
```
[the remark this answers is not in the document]
JonathanHenson: will update

include/aws/io/retry_strategy.h
JonathanHenson: In fact, I can see multiple implementations of these being in a different lib. The interface though, makes most sense to me in aws-c-io. Just my thoughts at the moment.

source/exponential_backoff_retry_strategy.c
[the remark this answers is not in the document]
JonathanHenson: actually, yeah, I think you're right... I'm going to keep the "release()" verb for the v-table though because it may or may not be a "destroy" operation.

source/exponential_backoff_retry_strategy.c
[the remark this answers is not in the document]
JonathanHenson: for the retry more than 63 portion, I think I'll just do a config check in setup.

source/retry_strategy.c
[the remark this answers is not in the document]
JonathanHenson: fixed.

source/exponential_backoff_retry_strategy.c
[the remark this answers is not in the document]
JonathanHenson: The API wouldn't have to change to alter this. It's not a one-way door. I vote for leaving it as is for now.

source/exponential_backoff_retry_strategy.c
```diff
@@ -0,0 +1,346 @@
...
+            "id=%p: Computed backoff value of %" PRIu64 "ns on token %p",
+            (void *)backoff_retry_token->base.retry_strategy,
+            backoff,
+            (void *)token);
+    }
+
+    bool already_scheduled = false;
+
+    { /***** BEGIN CRITICAL SECTION *********/
+        AWS_FATAL_ASSERT(!aws_mutex_lock(&backoff_retry_token->thread_data.mutex) && "Mutex lock acquisition failed");
```
justinboswell, line 244: Same as above, make the failure just a bit more specific

tests/exponential_backoff_retry_test.c
```diff
@@ -0,0 +1,256 @@
...
+
+struct exponential_backoff_test_data {
+    size_t retry_count;
+    size_t client_error_count;
+    int failure_error_code;
+    struct aws_mutex mutex;
+    struct aws_condition_variable cvar;
+};
+
+static void s_too_many_retries_test_on_retry_ready(struct aws_retry_token *token, int error_code, void *user_data) {
```
justinboswell, line 32: Can you just add a comment on each test case to explain what you're testing for briefly, just to be a bit more clear?

include/aws/io/retry_strategy.h
```diff
@@ -0,0 +1,178 @@
...
+
+/**
+ * Invoked upon the acquisition, or failure to acquire a retry token. This function will always be invoked if and only
+ * if aws_retry_strategy_acquire_retry_token() returns AWS_OP_SUCCESS. It will never be invoked synchronously from
+ * aws_retry_strategy_acquire_retry_token(). Token will always be NULL if error_code is non-zero, and vice-versa.
+ */
+typedef void(aws_retry_strategy_on_retry_token_acquired_fn)(
+    struct aws_retry_strategy *retry_strategy,
+    int error_code,
+    struct aws_retry_token *token,
```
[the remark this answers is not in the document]
JonathanHenson: Well we do a sender style here because I don't think we want users dealing with retry strategies once they're working with a token, hence all remaining token functions only take the token pointer.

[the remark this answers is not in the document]
JonathanHenson: the token does internally point to the retry strategy though.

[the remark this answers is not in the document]
graebm: Yeah, we do need the retry_strategy ptr here because `token` will be NULL if you're unable to get the initial token

include/aws/io/retry_strategy.h
graebm: yeah, didn't think of the threading implications

source/exponential_backoff_retry_strategy.c
```diff
@@ -0,0 +1,351 @@
...
+    struct aws_retry_strategy base;
+    struct aws_exponential_backoff_retry_options config;
+};
+
+struct exponential_backoff_retry_token {
+    struct aws_retry_token base;
+    struct aws_atomic_var current_retry_count;
+    struct aws_atomic_var last_backoff;
+    size_t max_retries;
+    uint64_t backofff_scale_factor_ns;
```
graebm, line 38: "fff"

source/exponential_backoff_retry_strategy.c
```diff
@@ -0,0 +1,351 @@
...
+
+    uint64_t random = token->generate_random();
+    return min + random % (diff);
+}
+
+typedef uint64_t(compute_backoff_fn)(struct exponential_backoff_retry_token *token);
+
+static uint64_t s_compute_no_jitter(struct exponential_backoff_retry_token *token) {
+    size_t retry_count = aws_atomic_load_int(&token->current_retry_count);
+    return aws_mul_u64_saturating((uint64_t)1 << retry_count, token->backofff_scale_factor_ns);
```
graebm, line 173: `size_t retry_count = aws_min_size(aws_atomic_load_int(&token->current_retry_count), 63);`

Someone someday is going to remove the check at setup and everyone on the PR will just think "lol 63 what a random value"

include/aws/io/retry_strategy.h
```diff
@@ -0,0 +1,178 @@
...
+ * Options for exponential backoff retry strategy. el_group must be set, any other option, if set to 0 will signify
+ * "use defaults"
+ */
+struct aws_exponential_backoff_retry_options {
+    /** Event loop group to use for scheduling tasks. */
+    struct aws_event_loop_group *el_group;
+    /** Max retries to allow. The default value is 10 */
+    size_t max_retries;
+    /** Scaling factor to add for the backoff. Default is 25ms */
+    uint32_t backoff_scale_factor_ms;
```
graebm, line 114: [just thinking out loud]: features we might want in some far-flung future:
1) unlimited retries
2) capped backoff time

(ex: MQTT's reconnect logic)

source/exponential_backoff_retry_strategy.c
```diff
@@ -0,0 +1,351 @@
...
+                s_exponential_retry_task,
+                backoff_retry_token,
+                "aws_exponential_backoff_retry_task");
+        }
+        AWS_FATAL_ASSERT(
+            !aws_mutex_unlock(&backoff_retry_token->thread_data.mutex) && "Retry token mutex release failed");
+    } /**** END CRITICAL SECTION ***********/
+
+    if (already_scheduled) {
+        return aws_raise_error(AWS_ERROR_INVALID_STATE);
```
graebm, line 266: AWS_LOGF_ERROR("retry token already scheduled")

PR #266, opened 2020-04-10
Adding functionality to get host resolver address count

rccarper:
*Description of changes:*
Adding functionality to get host resolver address count, which is currently needed by the canary when grabbing DNS addresses.


include/aws/io/host_resolver.h
```diff
@@ -26,6 +26,13 @@ typedef enum aws_address_record_type {
     AWS_ADDRESS_RECORD_TYPE_AAAA
 } aws_address_record_type;
 
+typedef enum aws_get_host_address_flags {
+    /* get number of ipv4 addresses. */
+    AWS_GET_HOST_ADDRESS_COUNT_RECORD_TYPE_A = 0x00000001,
```
JonathanHenson, line 31: Why not use the existing enums here?

rccarper: These are meant to be combinable flags, whereas the exiting enums are zero based.  We could also add more flags here in the future, although I don't know what those would be right now.

JonathanHenson: Okay, we also don’t usually typedef enums.

PR #268, opened 2020-04-15
Fixed DNS resolution on android

justinboswell:

### DIFF
--- a/source/posix/host_resolver.c
+++ b/source/posix/host_resolver.c
@@ -41,6 +41,10 @@ int aws_default_dns_resolve(
     const char *hostname_cstr = aws_string_c_str(host_name);
     AWS_LOGF_DEBUG(AWS_LS_IO_DNS, "static: resolving host %s", hostname_cstr);
 
+    /* Android would prefer NO HINTS IF YOU DON'T MIND, SIR */
+#ifdef ANDROID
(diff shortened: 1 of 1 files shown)


source/posix/host_resolver.c
```diff
@@ -41,13 +41,18 @@ int aws_default_dns_resolve(
...
+#ifdef ANDROID
+    int err_code = getaddrinfo(hostname_cstr, NULL, NULL, &result);
+#else
     struct addrinfo hints;
     AWS_ZERO_STRUCT(hints);
     hints.ai_family = AF_UNSPEC;
     hints.ai_socktype = SOCK_STREAM;
     hints.ai_flags = AI_ALL | AI_V4MAPPED;
 
     int err_code = getaddrinfo(hostname_cstr, NULL, &hints, &result);
```
bretambrose, line 54: Do we still getting a union of v4 and v6 in the results?

justinboswell: If you're asking if we support it, yes we do.

bretambrose: To clarify was just wondering it that's what we get back without hints.

justinboswell: Yes. It boils down to what the interfaces available support, and it's a union between them.

PR #272, opened 2020-04-24
implement aws_input_stream_get_remaining_length and aws_input_stream_…

czakian:
…read_n functions

*Issue #, if available:*

*Description of changes:*
- Adds an aws_input_stream_get_remaining_length() API function which returns the length in bytes remaining in an AWS input stream. 
- Adds an aws_input_stream_read_n() API function which reads up to n bytes from an input stream. 
 

czakian:
I have been able to update the code in aws-c-http so that I don't need these functions.

PR #275, opened 2020-05-20
UDP sockets can now use bind/connect in addition to socket/connect

justinboswell:
Depends on: https://github.com/awslabs/aws-crt-builder/pull/43


source/posix/socket.c
```diff
@@ -526,7 +526,8 @@ int aws_socket_connect(
         return aws_raise_error(AWS_IO_EVENT_LOOP_ALREADY_ASSIGNED);
     }
 
-    if (socket->state != INIT) {
+    /* UDP sockets jump straight to CONNECT_READ once bind is called */
+    if (socket->state != INIT && socket->options.type != AWS_SOCKET_DGRAM) {
```
JonathanHenson, line 530: You should still check for valid state when udp is used.

justinboswell: Good call, fixed.

justinboswell:
> What about windows?

Bah, meant to create a draft :( Not done yet!

tests/socket_test.c
```diff
@@ -253,6 +253,9 @@ static int s_test_socket(
 
     struct aws_socket outgoing;
     ASSERT_SUCCESS(aws_socket_init(&outgoing, allocator, options));
+    if (local && local != endpoint) {
```
bretambrose, line 256: value comparison instead?

.github/workflows/ci.yml
```diff
@@ -7,8 +7,8 @@ on:
       - '!master'
 
 env:
-  BUILDER_VERSION: v0.6.0
-  BUILDER_SOURCE: releases
+  BUILDER_VERSION: fixes
```
bretambrose, line 10: Is this final or would this change once fixes gets merged?

tests/socket_test.c
justinboswell: I can add that.

.github/workflows/ci.yml
justinboswell: Once the dependent change is merged, I'll update this to a release.

PR #286, opened 2020-06-26
Update license to remove Copyright year.

singku-china:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.github/workflows/proof-alarm.yml
+++ b/.github/workflows/proof-alarm.yml
@@ -15,7 +15,7 @@ jobs:
 
     - name: Check
       run: |
-        git diff --quiet 9276b7251 source/linux/epoll_event_loop.c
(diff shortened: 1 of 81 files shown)


nchong-at-aws:
Hi Andrew, you can fix the ProofAlarm failure by updating the hash in `.github/workflows/proof-alarm.yml` from `9276b7251` to `aaabab98c`. We have a set of proofs for the epoll event loop, however they're maintained separately from the implementation. The proof alarm warns if the source file changes, which might mean the proofs need to be updated. In this case, the license change is not a concern for the correctness of the proofs so it's safe to update the alarm.

nchong-at-aws:
Ah the merge changed the hash to 84bfb96d so the alarm is still failing. I'll make a PR.

PR #294, opened 2020-07-09
Windows 7 TLS/PKI support

justinboswell:
*Issue #, if available:* https://github.com/awslabs/aws-crt-cpp/issues/144

*Description of changes:*
* Force checks for ALPN availability in every case where ALPN might be attempted, should allow Win7 and Win8 < Win8.1 to use TLS 1.2 if enabled.
* Fixed spurious failure due to duplicate registration of IOCP on named pipes
* Fixes to unit tests resource installation on windows, should now be robust regardless of build configuration

### DIFF
--- a/source/windows/iocp/iocp_event_loop.c
+++ b/source/windows/iocp/iocp_event_loop.c
@@ -498,8 +498,7 @@ static int s_connect_to_io_completion_port(struct aws_event_loop *event_loop, st
         1);                  /* NumberOfConcurrentThreads */
 
     /* iocp_handle should be the event loop's handle if this succeeded */
-    bool iocp_associated =
-        iocp_handle == impl->iocp_handle
+    bool iocp_associated = iocp_handle == impl->iocp_handle;
 
 /* clang-format off */
 #if defined(AWS_SUPPORT_WIN7)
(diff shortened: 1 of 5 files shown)


source/windows/iocp/iocp_event_loop.c
```diff
@@ -496,7 +496,13 @@ static int s_connect_to_io_completion_port(struct aws_event_loop *event_loop, st
...
         1);                  /* NumberOfConcurrentThreads */
 
-    if (!success) {
+    /*
+     * Some handles, notably named pipes, don't like to be associated more than
+     * once with the same I/O completion port. Therefore, it's ok if this call
+     * fails as long as there's already a bound iocp_handle.
+     * This will result in a last error of ERROR_INVALID_PARAMETER.
+     */
+    if (!success && impl->iocp_handle == INVALID_HANDLE_VALUE) {
```
graebm, line 505: Something is weird here.
`impl` is the `iocp_loop`. I don't know how `impl->iocp_handle` could possibly be `INVALID_HANDLE_VALUE`.

Here are the docs for [CreateIoCompletionPort](https://docs.microsoft.com/en-us/windows/win32/fileio/createiocompletionport#return-value). Could we call `GetLastError()` to determine the specific edge-case we're trying to allow here?

justinboswell: It's ```INVALID_HANDLE_VALUE``` when it goes in for some cases (I believe for normal sockets). It comes out with a valid handle to the IOCP. If the IOCP is invalid, then this is a failure.

PR #295, opened 2020-07-10
add initial vsock support

petreeftime:
This adds the ability to create and use vsock sockets, which can serve
as a communication mechanism between multiple VMs or VM and host.

VSOCK are not available everywhere, so this feature is guarded by the
USE_VSOCK flags and test to check vsock functionality requires the
local CID feature, available on Linux 5.6+ with the vsock_loopback
driver.

*Issue #, if available:*

*Description of changes:*


tests/socket_test.c
```diff
@@ -432,6 +436,33 @@ static int s_test_tcp_socket_communication(struct aws_allocator *allocator, void
...
+
+    struct aws_socket_options options;
+    AWS_ZERO_STRUCT(options);
+    options.connect_timeout_ms = 3000;
+    options.type = AWS_SOCKET_STREAM;
+    options.domain = AWS_SOCKET_VSOCK;
+
+    struct aws_socket_endpoint endpoint;
+
+    snprintf(endpoint.address, sizeof(endpoint.address), "%u", cid);
```
alcioa, line 456: What about `endpoint.port`? Also docs should be updated for this `aws_socket_endpoint` structure in order to state that `address` can also be a decimal CID string among IPv4/IPv6 strings.

source/posix/socket.c
```diff
@@ -718,6 +773,13 @@ int aws_socket_bind(struct aws_socket *socket, const struct aws_socket_endpoint
         address.sock_addr_types.un_addr.sun_family = AF_UNIX;
         strncpy(address.sock_addr_types.un_addr.sun_path, local_endpoint->address, AWS_ADDRESS_MAX_LEN);
         sock_size = sizeof(address.sock_addr_types.un_addr);
+#ifdef USE_VSOCK
+    } else if (socket->options.domain == AWS_SOCKET_VSOCK) {
+        pton_err = parse_cid(local_endpoint->address, &address.sock_addr_types.vm_addr.svm_cid);
+        address.sock_addr_types.vm_addr.svm_family = AF_VSOCK;
+        address.sock_addr_types.vm_addr.svm_port = (int)local_endpoint->port;
```
alcioa, line 780: Not sure you need this cast? I reckon svm_port is `unsigned int`.

source/posix/socket.c
```diff
@@ -551,6 +599,13 @@ int aws_socket_connect(
         address.sock_addr_types.un_addr.sun_family = AF_UNIX;
         strncpy(address.sock_addr_types.un_addr.sun_path, remote_endpoint->address, AWS_ADDRESS_MAX_LEN);
         sock_size = sizeof(address.sock_addr_types.un_addr);
+#ifdef USE_VSOCK
+    } else if (socket->options.domain == AWS_SOCKET_VSOCK) {
+        pton_err = parse_cid(remote_endpoint->address, &address.sock_addr_types.vm_addr.svm_cid);
+        address.sock_addr_types.vm_addr.svm_family = AF_VSOCK;
+        address.sock_addr_types.vm_addr.svm_port = (int)remote_endpoint->port;
```
alcioa, line 606: Not sure you need this cast? I reckon svm_port is `unsigned int`.

source/posix/socket.c
```diff
@@ -498,9 +510,45 @@ struct socket_address {
...
+    /* unsigned long long to handle edge cases in convention explicitly */
+    long long cid = strtoll(cid_str, NULL, 10);
+    if (errno != 0) {
+        return 0;
+    }
+
+    /* -1U means any, so it's a valid value, but it needs to be converted to
+     * unsigned int. */
+    if (cid == -1) {
+        *value = UINT_MAX;
```
alcioa, line 538: For keeping consistency with `vm_sockets.h` this should be VMADDR_CID_ANY

source/posix/socket.c
```diff
@@ -498,9 +510,45 @@ struct socket_address {
...
+static int parse_cid(const char *cid_str, unsigned int *value) {
+    if (cid_str == NULL || value == NULL) {
+        errno = EINVAL;
+        return 0;
+    }
+    /* strtoll returns 0 as both error and correct value */
+    errno = 0;
+    /* unsigned long long to handle edge cases in convention explicitly */
+    long long cid = strtoll(cid_str, NULL, 10);
+    if (errno != 0) {
```
alcioa, line 531: Should check for `(cid == 0 && errno)` as per man pages.

tests/socket_test.c
```diff
@@ -432,6 +436,33 @@ static int s_test_tcp_socket_communication(struct aws_allocator *allocator, void
...
+
+    return s_test_socket(allocator, &options, &endpoint);
+#    endif
```
alcioa, line 459: No need for this space between `#` and `endif` even for nested macros.

tests/socket_test.c
```diff
@@ -432,6 +436,33 @@ static int s_test_tcp_socket_communication(struct aws_allocator *allocator, void
 
 AWS_TEST_CASE(tcp_socket_communication, s_test_tcp_socket_communication)
 
+#if defined(USE_VSOCK)
+static int s_test_vsock_loopback_socket_communication(struct aws_allocator *allocator, void *ctx) {
+/* Without vsock loopback it's difficult to test vsock functionality.
+ * Also note that having this defined does not guarantee that it's available
+ * for use and there's no path to figure out dynamically if it can be used. */
+#    if defined(VMADDR_CID_LOCAL)
```
alcioa, line 444: No need for this space between `#` and `if` even for nested macros.

source/posix/socket.c
petreeftime: This should catch all possible conversion errors (invalid / underflow / overflow). Is there something I'm missing here?

tests/socket_test.c
petreeftime: I've used the `format-check.sh` script and this is the expected identation according to it. You can see his pattern used in other files in this project as well, see https://github.com/awslabs/aws-c-io/blob/master/include/aws/io/exports.h.

source/posix/socket.c
petreeftime: Updated.

source/posix/socket.c
petreeftime: Updated.

tests/socket_test.c
petreeftime: Resolved.

source/posix/socket.c
alcioa: Yeah, it's fine like this.

tests/socket_test.c
```diff
@@ -432,6 +436,30 @@ static int s_test_tcp_socket_communication(struct aws_allocator *allocator, void
...
+    AWS_ZERO_STRUCT(options);
+    options.connect_timeout_ms = 3000;
+    options.type = AWS_SOCKET_STREAM;
+    options.domain = AWS_SOCKET_VSOCK;
+
+    struct aws_socket_endpoint endpoint = {.address = "1" /* VMADDR_CID_LOCAL */, .port = 8127};
+
+    return s_test_socket(allocator, &options, &endpoint);
+#    endif
+    return 0;
```
graebm, line 457: trivial: maybe like this, to avoid the compiler seeing 2 return statements one after the other? Pretty sure some linter somewhere would be mad about that:
```
#else
    return 0;
#endif
```

PR #308, opened 2020-08-06
Internal bootstrap changes

bretambrose:
Ref counting updates/conversions for

* aws_event_loop_group
* aws_host_resolver
* aws_client_bootstrap
* aws_server_bootstrap
* aws_tls_ctx

Additionally, the client and server connection args internal structures have also had ref counting added.

Most the changes are fairly straightforward, if not a bit tedious, but the host resolver required major surgery in order to develop a shutdown process that you can apply invariant-based reasoning to with respect to the lifetimes of the host resolver itself relative to its worker threads.  Like event loop group, we want to be able to tell the host resolver to go away and have a guaranteed, fixed point in time shutdown callback where it's guaranteed that every auxiliary thread has completed and no outstanding memory allocations exist either with respect to the host resolver or its resolution threads.

To do this, I had to simplify some aspects of the host resolver:

1. The host resolver can no longer "resurrect" a resolution thread by flipping the old keep alive from false to true.  This property actually could lead to race conditions and interminable lookups.
1. The decision to terminate a resolution thread can only be made by the resolution thread itself (it can be told to via synchronized flag, but nothing happens until the thread sees that flag).
1. At that termination decision point (underneath both the resolver and the entry locks), the resolution thread removes the host_entry from the resolver's map.  This is critical to give us a consistent view of the resolver map: if the entry is reachable from the map, the resolution thread is alive and queries can be submitted and are guaranteed to get a response.
1. Since it is the resolution thread making the decision to remove from the resolver's table, we can no longer use the table/cache removal callback to destroy the entry - instead the entry is freed on shutdown callback from the thread.  
1. Since the cache removal callback cannot be used as a destructor, we have to switch to hash table because you can't iterate a cache.


source/channel_bootstrap.c
```diff
@@ -23,28 +21,36 @@
...
     aws_client_bootstrap_shutdown_complete_fn *on_shutdown_complete = bootstrap->on_shutdown_complete;
     void *user_data = bootstrap->user_data;
+
+    aws_event_loop_group_release(bootstrap->event_loop_group);
+    aws_host_resolver_release(bootstrap->host_resolver);
+
     aws_mem_release(bootstrap->allocator, bootstrap);
 
     if (on_shutdown_complete) {
         on_shutdown_complete(user_data);
```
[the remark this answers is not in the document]
bretambrose: I don't think either of the two cases are an issue, but I'm just skimming.  For memory checking, the assumption is that you use the aws_global_wait_thingy() call and that will not return until every elg and host resolver thread are 100% gone with no crt allocations un-released (sourced to them).  While there may be other non-crt threads (there aren't in any of the tests though), from your perspective, if you've waited then the only CRT objects left are ones you haven't released and their destruction is synchronous because all of the async things are fully gone.

source/event_loop.c
```diff
@@ -9,45 +9,115 @@
...
+
+    aws_thread_current_at_exit(s_event_loop_group_thread_exit, el_group);
+}
+
+static void s_aws_event_loop_group_shutdown_async(struct aws_event_loop_group *el_group) {
+
+    struct aws_thread cleanup_thread;
+    AWS_ZERO_STRUCT(cleanup_thread);
+
+    AWS_FATAL_ASSERT(aws_thread_init(&cleanup_thread, el_group->allocator) == AWS_OP_SUCCESS);
```
graebm, line 54:
```
/* It's likely that the last refcount was released on an event-loop thread,
 * so we would deadlock if we waited here for all the event-loop threads to shut down.
 * Therefore, we spawn a NEW thread and have it wait for all the event-loop threads to shut down */

File: source/event_loop.c [hunk 3]
```diff
@@ -9,45 +9,115 @@
...
+
+    struct aws_thread_options thread_options;
+    AWS_ZERO_STRUCT(thread_options);
+
+    AWS_FATAL_ASSERT(
+        aws_thread_launch(&cleanup_thread, s_event_loop_destroy_async_thread_fn, el_group, &thread_options) ==
```
graebm, line 60: I've got a way to avoid these FATAL_ASSERTS, but it's probably too much tinkering to do right now when we really want this in mainline. Maybe just put in a TODO:

/* TODO: get rid of FATAL_ASSERTS by initializing and launching cleanup thread during event-loop-group initialization. Cleanup thread waits until it's signaled via condition variable. */

tests/channel_test.c
```diff
@@ -645,11 +645,19 @@ static int s_test_channel_connect_some_hosts_timeout(struct aws_allocator *alloc
...
+        .mutex = &mutex,
+        .cv = AWS_CONDITION_VARIABLE_INIT,
+        .error_code = 0,
+        .channel = NULL,
+        .setup = false,
+        .shutdown = false,
+    };
+
+    struct aws_event_loop_group *event_loop_group = aws_event_loop_group_new_default(allocator, 1, NULL);
+
```
graebm, line 660: `ASSERT_NOT_NULL(event_loop_group)`

This applies to pretty much every test?
or maybe it doesn't matter because everything would just explode?

tests/channel_test.c
```diff
@@ -734,12 +742,11 @@ static int s_test_channel_connect_some_hosts_timeout(struct aws_allocator *alloc
     ASSERT_SUCCESS(aws_array_list_push_back(&address_list, &host_address_1));
     ASSERT_SUCCESS(mock_dns_resolver_append_address_list(&mock_dns_resolver, &address_list));
 
-    struct aws_host_resolver resolver;
-    ASSERT_SUCCESS(aws_host_resolver_init_default(&resolver, allocator, 8, &event_loop_group));
+    struct aws_host_resolver *resolver = aws_host_resolver_new_default(allocator, 8, event_loop_group, NULL);
```
graebm, line 745: ASSERT_NOT_NULL(resolver)

source/host_resolver.c
```diff
@@ -88,50 +83,188 @@ int aws_host_resolver_record_connection_failure(struct aws_host_resolver *resolv
...
     struct aws_allocator *allocator;
     struct aws_host_resolver *resolver;
     struct aws_thread resolver_thread;
+    const struct aws_string *host_name;
+    int64_t resolve_frequency_ns;
+    struct aws_host_resolution_config resolution_config;
+
+    /* synchronized data and its lock */
     struct aws_mutex entry_lock;
+    struct aws_condition_variable entry_signal;
```
graebm, line 127: debatable/suggestion (you will end up modifying like every line of the file):
```
struct synced {
    struct aws_mutex lock;
    struct aws_condition_variable entry_signal;
    struct aws_cache *aaaa_records;
    ...
    enum default_resolver_state state;
};
```

source/host_resolver.c
```diff
@@ -88,50 +83,188 @@ int aws_host_resolver_record_connection_failure(struct aws_host_resolver *resolv
...
-    struct aws_condition_variable resolver_thread_semaphore;
-    const struct aws_string *host_name;
-    struct aws_host_resolution_config resolution_config;
     struct aws_linked_list pending_resolution_callbacks;
-    int64_t resolve_frequency_ns;
-    /* this member will be a monotonic increasing value and not protected by a memory barrier.
-       Let it tear on 32-bit systems, we don't care, we just want to see a change. This at least assumes cache coherency
-       for the target architecture, which these days is a fairly safe assumption. Where it's not a safe assumption, we
-       probably don't have multiple cores available anyways. */
-    volatile uint64_t last_use;
```
graebm, line 117: the only good volatile is a dead volatile
🎉☠️🎉

source/host_resolver.c
```diff
@@ -558,13 +691,58 @@ static void resolver_thread_fn(void *arg) {
...
 
-        aws_mutex_lock(&host_entry->semaphore_mutex);
+        aws_mutex_lock(&host_entry->entry_lock);
+
+        ++host_entry->resolves_since_last_request;
 
         /* we don't actually care about spurious wakeups here. */
-        aws_condition_variable_wait_for(
-            &host_entry->resolver_thread_semaphore, &host_entry->semaphore_mutex, host_entry->resolve_frequency_ns);
+        aws_condition_variable_wait_for_pred(
```
graebm, line 699: delete this comment.
It was bragging about how it wasn't using a predicate.
But now we're using one

source/channel_bootstrap.c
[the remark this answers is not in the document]
graebm: Here's the pathalogical case:
- Main thread doing `aws_global_thread_destroy_wait()`
- Nothing is currently running on the event_loop_group.
- Some object is holding the last reference to the event_loop_group.
- User releases reference to that object from a thread which is not the main thread, or an event-loop thread.
- object cleanup code calls `aws_event_loop_group_release(object->elg);` before object goes on to free its own memory.
- maybe this machine only has 1 core, or is really saturated with work
- Somehow the event_loop_group goes through its WHOLE ENTIRE WACKY CLEANUP SEQUENCE INVOLVING NEW THREADS before the `aws_event_loop_group_release()` returns. So the object still hasn't cleaned up its own memory yet.
- Main thread returns from `aws_global_thread_destroy_wait()` and goes on to check for memory leask

source/darwin/secure_transport_tls_channel_handler.c
```diff
@@ -945,6 +945,27 @@ struct aws_channel_handler *aws_tls_server_handler_new(
...
+    if (secure_transport_ctx->ca_cert) {
+        aws_release_certificates(secure_transport_ctx->ca_cert);
+    }
+
+    if (secure_transport_ctx->alpn_list) {
+        aws_string_destroy(secure_transport_ctx->alpn_list);
+    }
+
+    CFRelease(secure_transport_ctx->wrapped_allocator);
+    aws_mem_release(secure_transport_ctx->ctx.alloc, secure_transport_ctx);
```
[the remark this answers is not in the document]
bretambrose: This was just moved, but I think there were error cases where we didn't call the whole thing and theoretically something could leak.

tests/channel_test.c
bretambrose: Yeah, I've never understood asserting on creation in a test.  A crashed test is a failed test.

source/channel_bootstrap.c
```diff
@@ -134,32 +131,43 @@ struct client_connection_args {
...
-    if (--args->ref_count == 0) {
-        struct aws_allocator *allocator = args->bootstrap->allocator;
-        s_client_bootstrap_release(args->bootstrap);
-        if (args->host_name) {
-            aws_string_destroy(args->host_name);
-        }
+    struct aws_allocator *allocator = args->bootstrap->allocator;
+    aws_client_bootstrap_release(args->bootstrap);
+    if (args->host_name) {
+        aws_string_destroy(args->host_name);
```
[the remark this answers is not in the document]
bretambrose: Tests will check memory after (1) global count reaches 0 which means there are no other crt threads running, (2) the current call stack unwinds back to the test, which means all releases have been invoked.  These two together imply everything has been cleaned up, assuming everything is properly released prior to the wait.

source/channel_bootstrap.c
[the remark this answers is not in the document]
bretambrose: Sorta makes sense but I don't think it invalidates it as a test barrier.  Doing the release from a thread that isn't an elg or the main thread that is waiting would break the reasoning though, yeah.  C++ is the only one doing the wait in its shutdown right now.  May just add calls to wait at the end of its test or something.

source/host_resolver.c
bretambrose: Maybe someday

PR #313, opened 2020-08-27
Removed PKI utils that iOS doesn't support

justinboswell:

### DIFF
--- a/source/tls_channel_handler.c
+++ b/source/tls_channel_handler.c
@@ -78,6 +78,8 @@ static int s_load_null_terminated_buffer_from_cursor(
     return AWS_OP_SUCCESS;
 }
 
+#if !defined(AWS_OS_IOS)
+
 int aws_tls_ctx_options_init_client_mtls(
     struct aws_tls_ctx_options *options,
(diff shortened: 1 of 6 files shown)


source/tls_channel_handler.c
```diff
@@ -226,6 +230,8 @@ int aws_tls_ctx_options_init_server_pkcs12(
 
 #endif /* __APPLE__ */
 
+#if !defined(AWS_OS_IOS)
+
 int aws_tls_ctx_options_init_default_server_from_path(
```
graebm, line 235: Turning off all the server features because it's IOS feels a little bit like we're if-defing on the wrong thing.
It's like, when someone turns off the cheat menu because it's a _NDEBUG. And you're like "just because I want it running fast doesn't mean I don't want the cheat menu"

Like, maybe we should have a distinct AWS_SERVER_ENABLED flag that is always off when we're IOS and that would be better for these particular features.

Anyway, this is fine for now. Don't agonize over a perfect solution unless you know exactly what to do right now

justinboswell: I'm in AGONY now.

PR #314, opened 2020-08-28
vsock: don't do DNS lookup on VSOCK endpoints

petreeftime:
vsock are local, inter-vm sockets, so they would not show up in any DNS
query. Adapt code so that only IP-based sockets use that codepath.

### DIFF
--- a/source/channel_bootstrap.c
+++ b/source/channel_bootstrap.c
@@ -436,6 +436,10 @@ static void s_on_client_channel_on_shutdown(struct aws_channel *channel, int err
     s_client_connection_args_release(connection_args);
 }
 
+static bool s_aws_socket_domain_uses_dns(enum aws_socket_domain domain) {
+    return domain == AWS_SOCKET_IPV4 || domain == AWS_SOCKET_IPV6;
+}
+
 static void s_on_client_connection_established(struct aws_socket *socket, int error_code, void *user_data) {
     struct client_connection_args *connection_args = user_data;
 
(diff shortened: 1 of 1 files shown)


source/channel_bootstrap.c
```diff
@@ -782,6 +784,9 @@ int aws_client_bootstrap_new_socket_channel(struct aws_socket_channel_bootstrap_
         AWS_ZERO_STRUCT(endpoint);
         memcpy(endpoint.address, host_name, host_name_len);
         endpoint.port = 0;
+        if (socket_options->domain == AWS_SOCKET_VSOCK) {
+            endpoint.port = port;
+        }
```
alcioa, line 789: Better with an if/else here.

alcioa (approved):

source/channel_bootstrap.c
```diff
@@ -451,7 +451,9 @@ static void s_on_client_connection_established(struct aws_socket *socket, int er
     }
 
     if (error_code || connection_args->connection_chosen) {
-        if (connection_args->outgoing_options.domain != AWS_SOCKET_LOCAL && error_code) {
+        if ((connection_args->outgoing_options.domain == AWS_SOCKET_IPV4 ||
```
bretambrose, line 454: Can we pull the domain check out into a helper function?

petreeftime: I've added an s_aws_socket_domain_uses_dns.

bretambrose (approved):

PR #316, opened 2020-09-10
s2n: enable OCSP stapling if supported

petreeftime:
OCSP stapling is not available in all configurations, for example
BoringSSL doesn't currently support it. To allow BoringSSL to work as
a LibCrypto backend for s2n, aws-c-io should first check if it can be enabled
and then ignore it if can't.

This fix requires https://github.com/awslabs/s2n/pull/2253 to be
able to function correctly, otherwise it will still behave as is it
did previously.

### DIFF
--- a/source/s2n/s2n_tls_channel_handler.c
+++ b/source/s2n/s2n_tls_channel_handler.c
@@ -1021,10 +1021,28 @@ static struct aws_tls_ctx *s_tls_ctx_new(
     }
 
     if (options->verify_peer) {
-        if (s2n_config_set_check_stapled_ocsp_response(s2n_ctx->s2n_config, 1) ||
-            s2n_config_set_status_request_type(s2n_ctx->s2n_config, S2N_STATUS_REQUEST_OCSP)) {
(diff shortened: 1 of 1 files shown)


alcioa (approved):

justinboswell (approved):

justinboswell:
> This fix requires [awslabs/s2n#2253](https://github.com/awslabs/s2n/pull/2253) to be
> able to function correctly, otherwise it will still behave as is it
> did previously.

Has this commit been shipped in a tagged release in s2n? If so, we'll need to update our submodules downstream.

petreeftime:
> > This fix requires [awslabs/s2n#2253](https://github.com/awslabs/s2n/pull/2253) to be
> > able to function correctly, otherwise it will still behave as is it
> > did previously.
> 
> Has this commit been shipped in a tagged release in s2n? If so, we'll need to update our submodules downstream.

No, not yet.

petreeftime:
The s2n commit is now included in the v0.10.15 tag.

PR #320, opened 2020-09-17
Adding Support for Host Listeners

rccarper:
*Description of changes:*
* Adding concept of host listeners to the host resolver.  Right now, it only supports listening for new addresses being resolved.  Ideally, I would like to add the address-removed/expired callback later.
* This PR originally had a way for grabbing all of the currently cached addresses for a host up-front.  That will be in a separate, future PR.


include/aws/io/host_resolver.h
```diff
@@ -211,6 +223,84 @@ AWS_IO_API size_t aws_host_resolver_get_host_address_count(
...
+
+typedef void(aws_host_resolver_listener_shutdown_fn)(void *user_data);
+
+struct aws_host_resolver_listener_options {
+
+    /* Name of the host to listen for notifications from. */
+    const struct aws_string *host_name;
+
+    /* Callback for when an address is resolved for the specified host. */
+    aws_host_resolver_resolved_address_fn *resolved_address_callback;
```
graebm, line 253: Not sure this new functionality belongs on the base class vtable. It seems specific to the `aws_100gb_nic_soaking_s3_wet_dream_host_resolver` which aggressively pre-caches addresses. Like, if I were to implement a power-conscious host resolver for use by IoT, it would only resolve hosts when asked to, and a callback like this that tells you about unsolicited background activity would make no sense.

But seriously, maybe today's the day we rename the "default" host resolver to `aws_host_resolver_s3_performance` or something like that.

include/aws/io/host_resolver.h
```diff
@@ -211,6 +223,84 @@ AWS_IO_API size_t aws_host_resolver_get_host_address_count(
...
+
+    /* Number of a-addresses to immediately retrieve from the cache. If the cache has less, it'll retrieve as
+     * many as possible. If set to g_aws_host_resolver_all_addresses, all available a-addresses will be returned.*/
+    size_t desired_num_a_addresses;
+
+    /* Number of aaaa-addresses to immediately retrieve from the cache. If the cache has less, it'll retrieve as
+     * many as possible. If set to g_aws_host_resolver_all_addresses, all available a-addresses will be returned.*/
+    size_t desired_num_aaaa_addresses;
+
+    aws_host_resolver_get_cached_addresses_callback_fn *get_cached_addresses_callback;
```
graebm, line 291: I'm confused about how you're using all this stuff.
Are you getting cached addresses at the same time you're registering a listener, so that you can keep a mirror of all available addresses?

If so, don't also need to know when the addresses are evicted from the host-resolver?

Also, there's a gap where new addresses might be registered between getting the cached addresses and registering the listener. Or if you register the listener first then get cached you could end up with duplicates.

But really ... just why? Why mirror the host resolver at all? What about the host resolver isn't fast enough for you that you need to duplicate all its data?

include/aws/io/host_resolver.h
rccarper: I was thinking listener functionality would be optionally overridable, which is how I implemented it in the source file, but I'm realizing I should have commented that as such in the header file.  (Not sure that that solves your concerns though.)

include/aws/io/host_resolver.h
rccarper: It's not a speed problem--it's a distribution problem.  The only available mechanism for querying an address from the host resolver is to get the next LRU address.  I think that's probably what you want 98% of the time, so that makes sense.  But for this particular use case, we want to divide traffic such that we have x number of resolved addresses, with y connections assigned to each of those resolved addresses.  Only being able to ask the host resolver for the next LRU address makes that difficult, because we only want to spin up the next set of connections when we have a new address. :(  It's also awkward because we want to share the host resolver to cut down on cost (ie: other people could be affecting what the LRU address is, someone else may have set up the host configuration).

We can totally add an on-evict callback in the same place.  It's not something I anticipate needing right now.  In trying to keep this change smaller, as long as we had a place to put it, I was okay with waiting until the need arose for it.  One reason I'm not using it is because the first caller to resolve the host gets to set the TTL.  If that's not my system for some reason, I don't control that value.  Another reason is because TTL may be a bit artificial/arbitrary for this use case.  We may abandon traffic in the future on certain addresses, but I'm skeptical that TTL would be the driving factor, or at the very least, the only factor.

In short, I don't need an exact mirror, I'm just looking to get x number of resolved addresses so that I can use them in a different way.  If they're available, I want to get them immediately.  If they're not, I want to wait until they show up.

And you're right, get-cache and add-listener calls are not guaranteed to be in sync/atomic.  I was worried that might be overkill, and was thinking that initially checking for duplicates or occasionally having an initial missed address in between those two calls would actually be okay.  But you're right, that's lame--I'll try to fix that.

source/host_resolver.c
```diff
@@ -113,6 +150,32 @@ struct default_host_resolver {
...
+
+    /* String copy of the host name */
+    struct aws_string *host_name;
+
+    /* User-supplied callbacks/user_data */
+    aws_host_listener_resolved_address_fn *resolved_address_callback;
+    aws_host_listener_shutdown_fn *shutdown_callback;
+    void *user_data;
+
+    /* Synchronous data, requires host resolver lock to read/modify*/
```
justinboswell, line 167: Let's put this in its own struct like we usually do, `synced_data` most likely

source/host_resolver.c
```diff
@@ -583,6 +709,112 @@ static bool s_host_entry_finished_pred(void *user_data) {
...
+    struct default_host_resolver *resolver,
+    const struct aws_array_list *new_address_list,
+    struct aws_linked_list *listener_list) {
+    (void)resolver;
+
+    AWS_ASSERT(resolver);
+    AWS_ASSERT(new_address_list);
+    AWS_ASSERT(listener_list);
+
+    struct aws_linked_list_node *listener_node = aws_linked_list_begin(listener_list);
```
justinboswell, line 749: How do we not need to lock around modifying the listener list?

source/host_resolver.c
```diff
@@ -113,6 +150,32 @@ struct default_host_resolver {
...
+    struct aws_string *host_name;
+
+    /* User-supplied callbacks/user_data */
+    aws_host_listener_resolved_address_fn *resolved_address_callback;
+    aws_host_listener_shutdown_fn *shutdown_callback;
+    void *user_data;
+
+    /* Synchronous data, requires host resolver lock to read/modify*/
+    bool owned_by_resolver_thread;
+    bool pending_destroy;
```
justinboswell, line 169: Let's put these at the end of the structure, and bit pack them

source/host_resolver.c
rccarper: The rest of the host resolver doesn't use that paradigm so I was worried it might be confusing to mix that in.  Might be okay though!

source/host_resolver.c
rccarper: Because this listener list is owned by the resolver thread, and only the resolver thread ever modifies it.

source/host_resolver.c
rccarper: Will do!

source/host_resolver.c
```diff
@@ -583,6 +709,112 @@ static bool s_host_entry_finished_pred(void *user_data) {
...
+
+        aws_linked_list_push_back(listener_list, &listener->node);
+
+        listener = s_pop_host_listener_from_entry(resolver, host_name, &listener_entry);
+    }
+}
+
+/* Notify all listeners with resolve address callbacks, and also clean up any that are waiting to be cleaned up. */
+/* Assumes no lock is held. */
+static void s_resolver_thread_update_listeners(
```
TingDaoK, line 739: maybe it will be more clear to comment that the `listener_list` is a local list, so it doesn't need to be protected by a lock?

source/host_resolver.c
justinboswell: We should move toward it over time. For this one, I'd rather break consistency and use the right pattern.

source/host_resolver.c
justinboswell: Ok, can you make that clearer in the comments?

source/host_resolver.c
```diff
@@ -83,6 +85,32 @@ int aws_host_resolver_record_connection_failure(struct aws_host_resolver *resolv
     return resolver->vtable->record_connection_failure(resolver, address);
 }
 
+struct aws_host_listener *aws_host_resolver_add_host_listener(
+    struct aws_host_resolver *resolver,
+    const struct aws_host_listener_options *options) {
+    AWS_ASSERT(resolver);
```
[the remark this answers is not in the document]
rccarper: I was trying to keep this consistent with the existing aws_host_resolver functions. :)

source/host_resolver.c
```diff
@@ -113,6 +150,32 @@ struct default_host_resolver {
...
+
+    /* User-supplied callbacks/user_data */
+    aws_host_listener_resolved_address_fn *resolved_address_callback;
+    aws_host_listener_shutdown_fn *shutdown_callback;
+    void *user_data;
+
+    /* Synchronous data, requires host resolver lock to read/modify*/
+    bool owned_by_resolver_thread;
+    bool pending_destroy;
+    struct aws_linked_list_node node;
```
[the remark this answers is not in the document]
rccarper: It is, I was trying to keep this consistent with the way the rest of the host_resolver code is written, ie, no synced_data.  But I'm going to do the synced_data approach.

include/aws/io/host_resolver.h
```diff
@@ -211,6 +222,51 @@ AWS_IO_API size_t aws_host_resolver_get_host_address_count(
...
+     * struct aws_host_address *host_address = NULL;
+     * aws_array_list_get_at_ptr(new_address_list, (void **)&host_address, address_index);
+     * */
+    const struct aws_array_list *new_address_list,
+
+    /* User data that was specified when adding the listener. */
+    void *user_data);
+
+/* Callback for when the listener has completed its clean up. */
+typedef void(aws_host_listener_shutdown_fn)(void *user_data);
```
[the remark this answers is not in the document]
rccarper: I feel like most systems in the CRT use the word "shutdown" for this sort of thing, but I could be wrong.  I'll double check!

include/aws/io/host_resolver.h
```diff
@@ -211,6 +222,51 @@ AWS_IO_API size_t aws_host_resolver_get_host_address_count(
...
+/* Callback for receiving new host addresses from a listener. Memory for the new address list is only guaranteed to
+ * exist during the callback, and must be copied if the caller needs it to persist after. */
+typedef void(aws_host_listener_resolved_address_fn)(
+    /* Listener that owns this callback. */
+    struct aws_host_listener *listener,
+
+    /* Array list of aws_host_address structures.  To get an item:
+     *
+     * struct aws_host_address *host_address = NULL;
+     * aws_array_list_get_at_ptr(new_address_list, (void **)&host_address, address_index);
```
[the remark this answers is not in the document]
rccarper: I've found getting a value out of an array list can be somewhat confusing.  (I usually crash once!)  So I liked calling that out here, but maybe it's overkill.

source/host_resolver.c
```diff
@@ -113,6 +150,32 @@ struct default_host_resolver {
...
+};
+
+/* Structure for holding all listeners for a particular host name. */
+struct host_listener_entry {
+    struct default_host_resolver *resolver;
+    struct aws_linked_list listeners;
```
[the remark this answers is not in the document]
rccarper: Will do!

source/host_resolver.c
rccarper: Will do!

source/host_resolver.c
```diff
@@ -101,9 +129,18 @@ struct default_host_resolver {
...
+    /* host_name (string) -> host_entry */
     struct aws_hash_table host_entry_table;
 
+    /* Hash table of listener entries per host name. We keep this decoupled from the host entry table to allow for
+     * listeners to be added/removed regardless of whether or not a corresponding host entry exists.
+     *
+     * Any time the listener list in the listener entry becomes empty, we remove the entry from the table.  This
+     * includes when a resolver thread moves all of the available listeners to its local list.
+     */
+    /* host_name (string) -> host_listener_entry */
```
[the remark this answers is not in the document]
rccarper: I think it's okay but adding the asterisk won't hurt. :)

include/aws/io/host_resolver.h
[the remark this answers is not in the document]
justinboswell: `destroy` is the direct API name we would use, but `shutdown_fn` is the right moniker here for a callback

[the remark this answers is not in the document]
TingDaoK: Got it!

source/host_resolver.c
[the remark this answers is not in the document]
TingDaoK: yeah, since a previous bug in aws-c-mqtt is like we stored pointer in array or table, then we forgot it...

justinboswell (approved):

justinboswell (approved):

PR #323, opened 2020-09-29
FindLibCrypto.cmake is easier to consume

graebm:
I'm trying to write a simple CMakeLists.txt file for a standalone app using the aws-crt-cpp libraries, and it was ending up with this same copy/paste voodoo that we put in every CMakeLists.txt file we write:

```
# This is required in order to append /lib/cmake to each element in CMAKE_PREFIX_PATH
set(AWS_MODULE_DIR "/${CMAKE_INSTALL_LIBDIR}/cmake")
string(REPLACE ";" "${AWS_MODULE_DIR};" AWS_MODULE_PATH "${CMAKE_PREFIX_PATH}${AWS_MODULE_DIR}")
# Append that generated list to the module search path
list(APPEND CMAKE_MODULE_PATH ${AWS_MODULE_PATH})
```
I was searching for a better way and found this [stackoverflow post](https://stackoverflow.com/questions/44920389/installing-a-cmake-library-also-ship-find-modules-for-the-dependencies).

This change makes it so that any consumer of `aws-c-io-config.cmake` gets their CMAKE_MODULE_PATH automatically modified so that FindLibCrypto.cmake will be found. . It feels a little hacky, but overall a better solution than what we had before.

If this is successful, we should do the same thing in s2n, and possibly get rid of aws-c-io's duplicate FindLibCrypto.cmake.

### DIFF
--- a/cmake/aws-c-io-config.cmake
+++ b/cmake/aws-c-io-config.cmake
@@ -1,7 +1,7 @@
 include(CMakeFindDependencyMacro)
 
 if (UNIX AND NOT APPLE)
-    set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_CURRENT_LIST_DIR}/modules")
+    list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/modules")
     find_dependency(s2n)
     find_dependency(LibCrypto)
 endif()
--- a/cmake/aws-c-io-config.cmake
+++ b/cmake/aws-c-io-config.cmake
@@ -1,6 +1,7 @@
 include(CMakeFindDependencyMacro)
 
 if (UNIX AND NOT APPLE)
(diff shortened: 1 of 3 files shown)


cmake/aws-c-io-config.cmake
```diff
@@ -1,6 +1,7 @@
 include(CMakeFindDependencyMacro)
 
 if (UNIX AND NOT APPLE)
+    set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_CURRENT_LIST_DIR}/modules")
```
justinboswell, line 4: Should this be `list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/modules")`?

graebm: That looks better, will do

justinboswell (approved):

PR #330, opened 2020-11-16
Added more diagnostics around read() failures

justinboswell:


source/posix/socket.c
```diff
@@ -1690,7 +1690,7 @@ int aws_socket_read(struct aws_socket *socket, struct aws_byte_buf *buffer, size
 
     int error = errno;
 
-    if (error == EAGAIN) {
+    if (error == EAGAIN || error == EWOULDBLOCK) {
```
JonathanHenson, line 1693: I think you might need to do some #ifdef defines for EWOULDBLOCK as I'm not certain it's available in all glibc versions.

JonathanHenson (approved):

PR #331, opened 2020-11-19
Use cal to provide crypto hooks

justinboswell:
Depends on: https://github.com/awslabs/aws-c-cal/pull/56


CMakeLists.txt
```diff
@@ -206,6 +201,7 @@ target_include_directories(${PROJECT_NAME} PUBLIC
         $<INSTALL_INTERFACE:include>)
 
 aws_use_package(aws-c-common)
+aws_use_package(aws-c-cal)
```
graebm, line 204: need to update README.txt so it walks through installing aws-c-cal in addition to aws-c-common

cmake/aws-c-io-config.cmake
```diff
@@ -7,6 +7,7 @@ if (UNIX AND NOT APPLE)
 endif()
 
 find_dependency(aws-c-common)
+find_dependency(aws-c-cal)
```
graebm, line 10: should you remove the `find_dependency(LibCrypto)` a few lines above this?

also, not sure if it matters,  but we don't *need* to have this line anymore:
`list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/modules")`
That was added so downstream projects would find the `FindLibCrypto.cmake` file that we were installing there, but that file no longer exists, so probably the `/modules` folder doesn't exist either on a clean install. Not sure if this matters...

graebm (approved):

PR #337, opened 2020-12-08
Host resolution thread pinning

bretambrose:
* Added resolver thread pinning via listener config
* Updated host resolution logging to make debugging easier
* Relaxed assert constraints due to suspected platform incosistencies


source/host_resolver.c
```diff
@@ -1020,9 +1050,21 @@ static void resolver_thread_fn(void *arg) {
                 host_entry->host_name);
             aws_mutex_unlock(&host_entry->entry_lock);
 
-            AWS_ASSERT(err_code != AWS_ERROR_SUCCESS || aws_array_list_length(&callback_address_list) > 0);
+            size_t callback_list_size = aws_array_list_length(&callback_address_list);
+            if (callback_list_size > 0) {
+                AWS_LOGF_DEBUG(
```
justinboswell, line 1055: This should probably be TRACE?

source/host_resolver.c
```diff
@@ -942,15 +958,29 @@ static void resolver_thread_fn(void *arg) {
...
         }
+
+        if (err_code == AWS_ERROR_SUCCESS) {
+            AWS_LOGF_DEBUG(
+                AWS_LS_IO_DNS,
+                "static, resolving host %s successful, returned %d addresses",
+                aws_string_c_str(host_entry->host_name),
+                (int)aws_array_list_length(&address_list));
+        } else {
+            AWS_LOGF_DEBUG(
```
justinboswell, line 976: Should this be a WARN/INFO?

source/host_resolver.c
```diff
@@ -942,15 +958,29 @@ static void resolver_thread_fn(void *arg) {
...
         /* resolve and then process each record */
         int err_code = AWS_ERROR_SUCCESS;
         if (host_entry->resolution_config.impl(
                 host_entry->allocator, host_entry->host_name, &address_list, host_entry->resolution_config.impl_data)) {
 
             err_code = aws_last_error();
         }
+
+        if (err_code == AWS_ERROR_SUCCESS) {
+            AWS_LOGF_DEBUG(
```
justinboswell, line 970: Feels like TRACE?

source/host_resolver.c
```diff
@@ -1020,9 +1050,21 @@ static void resolver_thread_fn(void *arg) {
...
-            AWS_ASSERT(err_code != AWS_ERROR_SUCCESS || aws_array_list_length(&callback_address_list) > 0);
+            size_t callback_list_size = aws_array_list_length(&callback_address_list);
+            if (callback_list_size > 0) {
+                AWS_LOGF_DEBUG(
+                    AWS_LS_IO_DNS,
+                    "static, invoking resolution callback for host %s with %d addresses",
+                    aws_string_c_str(host_entry->host_name),
+                    (int)callback_list_size);
+            } else {
+                AWS_LOGF_DEBUG(
```
justinboswell, line 1061: This log message doesn't seem accurate to what is actually happening? Isn't this just "there were no callbacks registered"?

include/aws/io/host_resolver.h
```diff
@@ -278,6 +278,9 @@ struct aws_host_listener_options {
 
     /* User data to be passed into each callback. */
     void *user_data;
+
+    /* flags/bitfield instead? */
+    bool pin_host_entry;
```
justinboswell, line 283: Can you document what the pinning accomplishes here?

source/host_resolver.c
bretambrose: Honestly, I don't want any host resolution logging trace.  Given the periodic per-second nature of the default host resolver, finding issues gets real ugly with trace as we've already seen with the crt http client where the customer has to use dropbox/third party service just to give us logs that cover a few minutes of light usage.

source/host_resolver.c
bretambrose: We could change it to warn.

source/host_resolver.c
bretambrose: Same as below

source/host_resolver.c
bretambrose: No, this is there's no addresses available

rccarper (approved):
Approved

Issue #340, opened 2020-12-15
Support custom keychain on macOS


EeroKurimo:
In https://github.com/aws/aws-iot-device-sdk-cpp-v2/issues/163 I suggested adding support for using a custom keychain instead of the default login keychain for use-cases like daemons using the AWS IoT SDK. It was suggested that I'd instead change the default keychain for the account to the private keychain, or just a PKCS#12 file which appeared to not use a keychain.

As commented in the ticket, now when doing the daemon integration, we wound that changing the default keychain of the daemon account appeared to cause some other issues, and using PKCS#12 appears to also rely on the default keychain. To avoid dealing with the default keychain, I'd wish to add support for optionally using a custom private keychain for the IoT TLS connections only.

There's also a related issue https://github.com/awslabs/aws-crt-cpp/issues/204 which implements the changes needed to the C++ wrapper to support these changes.

EeroKurimo:
This was implemented in #341

PR #341, opened 2020-12-15
Custom keychain

EeroKurimo:
*Issue #, if available:* #340 

*Description of changes:*

Introduces functions `aws_tls_ctx_options_init_client_mtls_custom_keychain` and `aws_tls_ctx_options_init_client_mtls_from_path_custom_keychain` that offer a path to the custom keychain. The only difference to the existing `init_client_mtls` functions is that they define `options->keychain_path = keychain_path`.

I also considered if introducing a new argument `keychain_path=NULL` to `aws_tls_ctx_options_init_client_mtls` would've made sense, but as this is a very macOS-specific change, keeping the `aws_tls_ctx_options_init_client_mtls` function identical for all platforms and offering a separate apple-only funtion for custom keychains sounded safer.


include/aws/io/tls_channel_handler.h
```diff
@@ -159,6 +159,15 @@ struct aws_tls_ctx_options {
      * Password for the pkcs12 data in pkcs12.
      */
     struct aws_byte_buf pkcs12_password;
+
+#    if !defined(AWS_OS_IOS)
+    /**
+     * On Apple OS you can also use a custom keychain instead of
+     * the default keychain of the account.
+     */
+    const char *keychain_path;
```
JonathanHenson, line 168: this needs to be aws_string

source/darwin/darwin_pki_utils.c
```diff
@@ -22,7 +22,8 @@ int aws_import_public_and_private_keys_to_identity(
     CFAllocatorRef cf_alloc,
     const struct aws_byte_cursor *public_cert_chain,
     const struct aws_byte_cursor *private_key,
-    CFArrayRef *identity) {
+    CFArrayRef *identity,
+    const char *keychain_path) {
```
JonathanHenson, line 26: this should be a byte_cursor if you don't need to preserve the null terminator or aws_string if you do.

source/darwin/darwin_pki_utils.c
```diff
@@ -45,7 +46,34 @@ int aws_import_public_and_private_keys_to_identity(
...
+        OSStatus keychain_status = SecKeychainOpen(keychain_path, &import_keychain);
+        if (keychain_status != errSecSuccess) {
+            AWS_LOGF_ERROR(
+                AWS_LS_IO_PKI,
+                "static: error opening keychain \"%s\" with OSStatus %d",
+                keychain_path,
+                keychain_status);
+            return AWS_OP_ERR;
+        }
+        keychain_status = SecKeychainUnlock(import_keychain, 0, "", TRUE);
```
JonathanHenson, line 60: can we use c99 standard types here? as in pass the stdint.h's bool (true)

source/tls_channel_handler.c
```diff
@@ -130,8 +130,63 @@ int aws_tls_ctx_options_init_client_mtls_from_path(
...
+    const struct aws_byte_cursor *cert,
+    const struct aws_byte_cursor *pkey,
+    const char *keychain_path) {
+    AWS_ZERO_STRUCT(*options);
+    options->minimum_tls_version = AWS_IO_TLS_VER_SYS_DEFAULTS;
+    options->cipher_pref = AWS_IO_TLS_CIPHER_PREF_SYSTEM_DEFAULT;
+    options->verify_peer = true;
+    options->allocator = allocator;
+    options->max_fragment_size = g_aws_channel_max_fragment_size;
+    options->keychain_path = keychain_path;
```
JonathanHenson, line 146: I mentioned before, but this is a foot gun. copy it to an aws_string inside this function.

source/darwin/darwin_pki_utils.c
EeroKurimo: It's passed on to [SecKeychainOpen](https://developer.apple.com/documentation/security/1396431-seckeychainopen?language=objc) which appears to to expect a null terminated string buffer, so aws_string sounds like the way to go.

source/darwin/darwin_pki_utils.c
EeroKurimo: Sure, I'll change it.

source/tls_channel_handler.c
EeroKurimo: I'll change both init functions to allocate that with `aws_string_new_from_c_str(options->allocator, keychain_path)` and free it with `aws_string_destroy()` in `aws_tls_ctx_options_clean_up()`

include/aws/io/tls_channel_handler.h
```diff
@@ -234,6 +243,36 @@ AWS_IO_API int aws_tls_ctx_options_init_client_mtls(
     const struct aws_byte_cursor *cert,
     const struct aws_byte_cursor *pkey);
 
+#    ifdef __APPLE__
```
bretambrose, line 246: Rather than more constructors -- which in this case mirror existing ones but with an additional parameter -- maybe it would be better to just add a set method for the keychain path ala aws_tls_ctx_options_set_alpn_list?

EeroKurimo: That sounds better than the extra constructors, I'll do that instead.

bretambrose:
Merged in https://github.com/awslabs/aws-c-io/pull/368

PR #342, opened 2020-12-16
Fix for tls data writes when channel overhead is not zero

bretambrose:
Several of the tls channel handlers did incorrect calculations on upstream data writes, but nothing went wrong until the upstream overhead was non zero, which hasn't happened until now (double tls support: to-proxy, tunnel-to-endpoint).  In the double tls use case, this turned into an infinite loop because the message that we requested was not able to fit any actual data in it other than the overhead.  

### DIFF
--- a/source/s2n/s2n_tls_channel_handler.c
+++ b/source/s2n/s2n_tls_channel_handler.c
@@ -207,20 +207,19 @@ static int s_generic_send(struct s2n_handler *handler, struct aws_byte_buf *buf)
 
     size_t processed = 0;
     while (processed < buf->len) {
+        const size_t overhead = aws_channel_slot_upstream_message_overhead(handler->slot);
+        const size_t message_size_hint = buf->len - processed + overhead;
         struct aws_io_message *message = aws_channel_acquire_message_from_pool(
-            handler->slot->channel, AWS_IO_MESSAGE_APPLICATION_DATA, buf->len - processed);
+            handler->slot->channel, AWS_IO_MESSAGE_APPLICATION_DATA, message_size_hint);
 
-        if (!message) {
+        if (!message || message->message_data.capacity <= overhead) {
             errno = ENOMEM;
             return -1;
         }
 
-        const size_t overhead = aws_channel_slot_upstream_message_overhead(handler->slot);
-        const size_t available_msg_write_capacity = buffer_cursor.len - overhead;
-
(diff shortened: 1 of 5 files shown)


justinboswell (approved):

source/s2n/s2n_tls_channel_handler.c
```diff
@@ -207,20 +207,19 @@ static int s_generic_send(struct s2n_handler *handler, struct aws_byte_buf *buf)
 
     size_t processed = 0;
     while (processed < buf->len) {
+        const size_t overhead = aws_channel_slot_upstream_message_overhead(handler->slot);
+        const size_t message_size_hint = buf->len - processed + overhead;
```
graebm, line 211: trivial: (a - b) + c

source/s2n/s2n_tls_channel_handler.c
```diff
@@ -207,20 +207,19 @@ static int s_generic_send(struct s2n_handler *handler, struct aws_byte_buf *buf)
...
-
-        const size_t to_write = message->message_data.capacity > available_msg_write_capacity
-                                    ? available_msg_write_capacity
-                                    : message->message_data.capacity;
+        const size_t available_msg_write_capacity = message->message_data.capacity - overhead;
+        const size_t to_write =
+            available_msg_write_capacity >= buffer_cursor.len ? buffer_cursor.len : available_msg_write_capacity;
 
         struct aws_byte_cursor chunk = aws_byte_cursor_advance(&buffer_cursor, to_write);
         if (aws_byte_buf_append(&message->message_data, &chunk)) {
```
graebm, line 225: You say windows isn't affected by this. Do we understand why?
Wouldn't it be safer to add this code than leave it out?
Are there unit tests that will at least catch it we leave it out, but it turns out to be necessary in some future scenario?

bretambrose: Windows wasn't affected because I was looking in the wrong spot and didn't see overhead usage but when I look in the right spot there it is too, so I'll add that.

There's a definite TBD: add a unit test that sets up two tls channel handlers in a row and makes sure both mocked negotitations fully succeed.

graebm (approved):

JonathanHenson (approved):

source/s2n/s2n_tls_channel_handler.c
bretambrose: Added unit test and fixed windows too.

PR #352, opened 2021-02-03
Thread join

bretambrose:
* managed thread updates for safest possible shutdown

### DIFF
--- a/build-deps.sh
+++ b/build-deps.sh
@@ -88,7 +88,7 @@ if [ $clean ]; then
 fi
 mkdir -p $deps_dir
 
-install_dep aws-c-common
+install_dep aws-c-common ThreadJoin
 install_dep aws-c-cal
 if [[ $OSTYPE != darwin* ]]; then
     install_dep s2n
--- a/build-deps.sh
+++ b/build-deps.sh
@@ -88,7 +88,7 @@ if [ $clean ]; then
 fi
 mkdir -p $deps_dir
 
-install_dep aws-c-common ThreadJoin
+install_dep aws-c-common
 install_dep aws-c-cal
 if [[ $OSTYPE != darwin* ]]; then
     install_dep s2n
--- a/source/bsd/kqueue_event_loop.c
+++ b/source/bsd/kqueue_event_loop.c
@@ -366,10 +366,12 @@ static int s_run(struct aws_event_loop *event_loop) {
      * and it's ok to touch cross_thread_data without locking the mutex */
     impl->cross_thread_data.state = EVENT_THREAD_STATE_RUNNING;
 
+    aws_thread_increment_unjoined_count();
     int err =
         aws_thread_launch(&impl->thread_created_on, s_event_thread_main, (void *)event_loop, &impl->thread_options);
 
     if (err) {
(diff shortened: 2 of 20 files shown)


build-deps.sh
```diff
@@ -88,7 +88,7 @@ if [ $clean ]; then
 fi
 mkdir -p $deps_dir
 
-install_dep aws-c-common
+install_dep aws-c-common ThreadJoin
```
bretambrose, line 91: Will undo once approved

source/bsd/kqueue_event_loop.c
```diff
@@ -366,10 +366,12 @@ static int s_run(struct aws_event_loop *event_loop) {
      * and it's ok to touch cross_thread_data without locking the mutex */
     impl->cross_thread_data.state = EVENT_THREAD_STATE_RUNNING;
 
+    aws_thread_increment_unjoined_count();
```
justinboswell, line 369: Why isn't this managed by the thread wrapper itself? This seems really error prone, to force downstream code to effectively ref-count this, when the `aws_thread` code knows whether or not it created a thread.

Only an unmanaged thread should have to do this. And unmanaged threads should be EXTREMELY rare. In fact, I'm not sure when we'd want one.

tests/default_host_resolver_test.c
```diff
@@ -143,7 +143,7 @@ static int s_test_default_with_ipv6_lookup_fn(struct aws_allocator *allocator, v
     aws_string_destroy((void *)host_name);
     aws_host_resolver_release(resolver);
     aws_event_loop_group_release(el_group);
-    ASSERT_SUCCESS(aws_global_thread_creator_shutdown_wait_for(10));
+    aws_thread_join_all_managed();
```
justinboswell, line 146: This should be accomplished by aws_io_library_clean_up() calling aws_common_library_clean_up(), it should not be a requirement for downstream users to manage. Lifetime should be clear: init -> run application -> clean_up.

source/bsd/kqueue_event_loop.c
bretambrose: event loop threads aren't technically managed (there's a bunch of wrapping logic around their lifetime in the event loop platform impls) but I think it's best to have them counted as if they were.

source/event_loop.c
```diff
@@ -69,6 +67,7 @@ static void s_aws_event_loop_group_shutdown_async(struct aws_event_loop_group *e
 
     struct aws_thread_options thread_options;
     AWS_ZERO_STRUCT(thread_options);
+    thread_options.join_strategy = AWS_TJS_MANAGED;
```
justinboswell, line 70: This says the threads are managed, but then we are manually twiddling the thread counts, which indicates unmanaged threads. We should pick one, this reads like a bug. Alternatively, we'll need to document it thoroughly, because this is not how we would expect threads to be managed.

bretambrose: It's the clean up thread which is managed.  Event loop threads are unmanaged but counted.

justinboswell: Oh my bad, thanks for clearing that up!

justinboswell (approved):

PR #353, opened 2021-02-04
socket write-completion callback always fires asynchronously 

graebm:
Previously, the callback could fire syncronously within a write() call. This was leading to bugs if multiple writes occurred in the same callstack (ex: https://github.com/aws/aws-iot-device-sdk-cpp-v2/issues/194).

### DIFF
--- a/source/posix/socket.c
+++ b/source/posix/socket.c
@@ -163,12 +163,22 @@ struct posix_socket {
     struct aws_linked_list written_queue;
     struct aws_task written_task;
     struct posix_socket_connect_args *connect_args;
-    int internal_refcount;
+    /* Note that only the posix_socket impl part is refcounted.
+     * The public aws_socket can be a stack variable and cleaned up synchronously
+     * (by blocking until the event-loop cleans up the impl part).
+     * In hindsight, aws_socket should have been heap-allocated and refcounted, but alas */
+    struct aws_ref_count internal_refcount;
+    struct aws_allocator *allocator;
     bool written_task_scheduled;
     bool currently_subscribed;
     bool continue_accept;
     bool *close_happened;
 };
+
+static void s_socket_destroy_impl(void *user_data) {
+    struct posix_socket *socket_impl = user_data;
+    aws_mem_release(socket_impl->allocator, socket_impl);
+}
 
 static int s_socket_init(
     struct aws_socket *socket,
@@ -208,7 +218,8 @@ static int s_socket_init(
     aws_linked_list_init(&posix_socket->written_queue);
     posix_socket->currently_subscribed = false;
     posix_socket->continue_accept = false;
-    posix_socket->internal_refcount = 1;
+    aws_ref_count_init(&posix_socket->internal_refcount, posix_socket, s_socket_destroy_impl);
+    posix_socket->allocator = alloc;
     posix_socket->connect_args = NULL;
     posix_socket->close_happened = NULL;
     socket->impl = posix_socket;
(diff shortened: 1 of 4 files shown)


source/posix/socket.c
```diff
@@ -1367,29 +1385,62 @@ int aws_socket_shutdown_dir(struct aws_socket *socket, enum aws_channel_directio
...
+            size_t bytes_written = write_request->original_buffer_len - write_request->cursor_cpy.len;
+            write_request->written_fn(socket, write_request->error_code, bytes_written, write_request->write_user_data);
+            aws_mem_release(allocator, write_request);
+            if (node == stop_after) {
+                break;
+            }
+        } while (!aws_linked_list_empty(&socket_impl->written_queue));
+    }
+
+    if (--socket_impl->internal_refcount == 0) {
```
justinboswell, line 1426: Any reason why this doesn't use aws_refcount?

source/posix/socket.c
```diff
@@ -1367,29 +1385,62 @@ int aws_socket_shutdown_dir(struct aws_socket *socket, enum aws_channel_directio
...
+     * to close the socket based on something they read (SSL validation failed for example).
+     * if clean_up happens when internal_refcount > 0, socket_impl is kept dangling */
+    struct aws_socket *socket = arg;
+    struct aws_allocator *allocator = socket->allocator;
+    struct posix_socket *socket_impl = socket->impl;
+    ++socket_impl->internal_refcount;
+    socket_impl->written_task_scheduled = false;
+
+    /* Notes about weird loop:
+     * 1) Use `stop_after` node so we only process the initial contents of queue.
```
justinboswell, line 1402: Initial contents of queue when this task is run, ignoring any writes queued during delivery

tests/socket_test.c
```diff
@@ -1326,6 +1326,232 @@ static int s_cleanup_in_write_cb_doesnt_explode(struct aws_allocator *allocator,
...
+
+static void s_async_write_task(struct aws_task *task, void *args, enum aws_task_status status) {
+    (void)task;
+    (void)args;
+    (void)status;
+
+    g_async_tester.currently_writing = true;
+
+    struct aws_byte_cursor data = aws_byte_cursor_from_c_str("A");
+    enum async_role *role = aws_mem_acquire(g_async_tester.allocator, sizeof(role));
```
justinboswell, line 1437: I can't figure out how this role is ever freed. What am I missing?

source/posix/socket.c
```diff
@@ -1263,7 +1268,10 @@ static void s_close_task(struct aws_task *task, void *arg, enum aws_task_status
 
 int aws_socket_close(struct aws_socket *socket) {
     struct posix_socket *socket_impl = socket->impl;
-    AWS_LOGF_DEBUG(AWS_LS_IO_SOCKET, "id=%p fd=%d: closing", (void *)socket, socket->io_handle.data.fd);
+    int fd = socket->io_handle.data.fd; /* cached for logging */
+    (void)fd;
```
TingDaoK, line 1272: Trivial: I am not getting the idea of caching... Will it be overwritten later by eventloop thread?🤔

tests/socket_test.c
TingDaoK: https://github.com/awslabs/aws-c-io/pull/353/files#diff-13f9eb7e9121add300ed1f5e22aaf7bc837babd0d535670cf16b4e1628d30a59R1388, it's used as the callback userdata, and cleaned up in the callback.

graebm: It's freed when the write-completion callback runs

source/posix/socket.c
graebm: yeah, there's a branch here where cleanup is performed on another thread while this thread waits. Then this thread logs that it's done, but the log statement was like:
`fd=7 closing from another thread`
and the next log statement was:
`fd=-1 close task completed`

I'll fix the name/comments here to be more clear

source/posix/socket.c
graebm: didn't use because we didn't need atomics, but yeah it makes this cleaner.
done

justinboswell (approved):

PR #354, opened 2021-02-08
Update README to use AWS-LC for libcrypto

andrew-kaufman:
*Issue #, if available:* N/A

*Description of changes:* This updates build instructions to use AWS-LC when building dependencies like s2n and aws-c-cal. This is dependent on the changes in this [PR](https://github.com/awslabs/aws-c-cal/pull/65)

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ This library is licensed under the Apache 2.0 License.
 
 #### Building s2n (Linux Only)
 
-If you are building on Linux, you will need to build s2n before being able to build aws-c-io.  For our CRT's, we build s2n at a specific commit, and recommend doing the same when using it with this library.  That commit hash can be found [here](https://github.com/awslabs/aws-crt-cpp/tree/main/crt).  The commands below will build s2n using OpenSSL 1.1.1.  For using other versions of OpenSSL, there is additional information in the [s2n Usage Guide](https://github.com/awslabs/s2n/blob/main/docs/USAGE-GUIDE.md).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -30,23 +30,14 @@ git checkout <s2n-commit-hash-used-by-aws-crt-cpp>
...
-cd `tar ztf openssl-1.1.1-latest.tar.gz | head -n1 | cut -f1 -d/`
-./config -fPIC no-shared              \
-         no-md2 no-rc5 no-rfc3779 no-sctp no-ssl-trace no-zlib     \
-         no-hw no-mdc2 no-seed no-idea enable-ec_nistp_64_gcc_128 no-camellia\
-         no-bf no-ripemd no-dsa no-ssl2 no-ssl3 no-capieng                  \
-         -DSSL_FORBID_ENULL -DOPENSSL_NO_DTLS1 -DOPENSSL_NO_HEARTBEATS      \
-         --prefix=<absolute-install-path>
-make
-make install
+# Get the latest version of AWS-LC
```
graebm, line 33: These instructions clone aws-lc into `<root>/s2n/libcrypto-build/aws-lc`

Should we have users build aws-lc **before** s2n, and just have them put aws-lc in their root, like we do with otehr aws-c-* dependencies?

graebm: Also, we tell users to checkout "s2n-commit-hash-used-by-aws-crt-cpp"
Should we also tell users to checkout a particular aws-lc hash?

andrew-kaufman: I think that makes sense, and I can update the location where `aws-lc` is cloned into. I was just trying to keep it similar to the current process when using OpenSSL. 

Also, we are discussing using a tagged release when checking out and building `aws-lc`. I can update that once I have a better answer after that meeting.

graebm: The current directions for building s2n are super confusing, I would not feel beholden to them.

A straightahead series of:
clone X, install X, clone Y, install Y, clone Z, install Z, ...
seems best to me

bretambrose:
I believe the README has been updated with step-by-step, repo-by-repo instructions now.

PR #358, opened 2021-02-26
source/windows: Fix some compiler warnings in Win32.

Biswa96:
Those warnings are:
* windows_pki_utils.c: passing argument 11 of 'CryptQueryObject' from incompatible pointer type
* pipe.c: function declaration isn't a prototype


Biswa96:
Ping.

bretambrose (approved):

PR #366, opened 2021-03-15
Pem util

TingDaoK:
*Issue #, if available:*

*Description of changes:*

- Anything outside `-----BEGIN "label"-----` and `-----END "label"-----` should be ignored by OpenSSL. Taking those away for some parser to work with comments. https://stackoverflow.com/questions/19578812/comments-in-a-pem-file


source/pem_utils.c
```diff
@@ -0,0 +1,75 @@
...
+#include <aws/common/string.h>
+#include <aws/io/private/pem_utils.h>
+
+enum aws_pem_util_state {
+    BEGIN,
+    ON_DATA,
+    END,
+};
+
+struct aws_string *aws_clean_up_pem(struct aws_byte_cursor pem, struct aws_allocator *allocator) {
```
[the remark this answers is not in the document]
TingDaoK: Feel like string may be used more commonly. Will change to buffer for less mem allocate.

source/pem_utils.c
```diff
@@ -0,0 +1,84 @@
...
+    }
+    char *clean_pem = NULL;
+    clean_pem = aws_mem_calloc(allocator, pem->len, sizeof(char));
+    if (!clean_pem) {
+        return AWS_OP_ERR;
+    }
+    struct aws_byte_cursor pem_cursor = aws_byte_cursor_from_buf(pem);
+    int state = BEGIN;
+    int clean_pem_index = 0;
+    const char *begin_header = "-----BEGIN";
```
bretambrose, line 26: Can we do this with byte cursors and aws_strings?

TingDaoK: you mean the `begin_header` to compare, or the clean_pem we are writing into? I basically follow the [pki_utils](https://github.com/awslabs/aws-c-io/blob/main/source/pki_utils.c#L43)😂

source/pem_utils.c
```diff
@@ -0,0 +1,84 @@
...
+#include <aws/common/string.h>
+#include <aws/io/private/pem_utils.h>
+
+enum aws_pem_util_state {
+    BEGIN,
+    ON_DATA,
+    END,
+};
+
+int aws_clean_up_pem(struct aws_byte_buf *pem, struct aws_allocator *allocator) {
```
justinboswell, line 14: I think we might want to name this `aws_normalize_pem` or `aws_sanitize_pem`, clean_up is the wrong verb, since clean_up means something else throughout the rest of our code.

source/pem_utils.c
justinboswell: I think he was suggesting using `AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL` and using byte cursor comparisons

justinboswell (approved):

Issue #261, opened 2020-03-31
Support for ECC keys in Windows port


massi-ang:
Looking at the code at:
https://github.com/awslabs/aws-c-io/blob/60a451a6a7269f1f99718950cd34befcf7e8a096/source/windows/windows_pki_utils.c#L333

it seems that Windows port only support RSA private keys in PKCS#1 format.

Add support for ECC keys in SEC 1 encoded format and possibly wrapped in PKCS#8.

JonathanHenson:
We should be doing something like this here.

https://github.com/Azure/azure-c-shared-utility/blob/master/adapters/x509_schannel.c#L97

massi-ang:
Maybe we could also check the first line of the PEM file, as that will tell you the type of key and encoding used.

bretambrose:
Ecc-based cert/key import on Windows should now be working correctly, per https://github.com/awslabs/aws-c-io/pull/371

PR #259, opened 2020-03-31
error handling for failed private key import

massi-ang:
*Issue #, if available:*
#258 
*Description of changes:*
Added handling of result code to print a useful message to the user in case the private key import fails

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #371, opened 2021-04-07
Windows certificate import improvements

bretambrose:
* Check and log all errors.  Previously certain system calls could fail and this would not be noticed.
* Refactor windows cert import to follow newer standards for RAII and cleanup
* Adds support for Windows ECC key import.  Requires aws-c-cal to link to NCrypt.

Owes a tremendous amount to the Azure Windows key import examples found here: https://github.com/Azure/azure-c-shared-utility/blob/master/adapters/x509_schannel.c

Addresses https://github.com/awslabs/aws-c-io/issues/261 and pulls in https://github.com/awslabs/aws-c-io/pull/259


source/windows/windows_pki_utils.c
```diff
@@ -181,31 +198,269 @@ int aws_import_trusted_certificates(
...
+    DWORD public_key_blob_length = public_key_blob->cbData;
+    if (public_key_blob_length == 0) {
+        AWS_LOGF_ERROR(AWS_LS_IO_PKI, "static: invalid zero-length ecc key data");
+        aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
+        goto done;
+    }
+
+    public_key_blob_length--;
+    struct aws_byte_cursor public_blob_cursor = {
+        .ptr = public_key_blob->pbData + 1,
```
justinboswell, line 350: Why is this +1? Could you add a comment?

source/windows/windows_pki_utils.c
```diff
@@ -181,31 +198,269 @@ int aws_import_trusted_certificates(
...
+
+    CRYPT_BIT_BLOB *public_key_blob = &cert_context->pCertInfo->SubjectPublicKeyInfo.PublicKey;
+    DWORD public_key_blob_length = public_key_blob->cbData;
+    if (public_key_blob_length == 0) {
+        AWS_LOGF_ERROR(AWS_LS_IO_PKI, "static: invalid zero-length ecc key data");
+        aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
+        goto done;
+    }
+
+    public_key_blob_length--;
```
justinboswell, line 348: Is this ignoring a null terminator?

source/windows/windows_pki_utils.c
bretambrose: Now with a big ol' comment.  Not completely resolved, but I'm not going to try and manually edit the DER form of a certificate to test a hypothesis that is mostly of curiosity.

source/windows/windows_pki_utils.c
bretambrose: per above

justinboswell (approved):
I asked the question, but I wasn't ready for the answer :( Nonetheless, let it fly!

PR #374, opened 2021-04-15
Changing fopen flags in aws_input_stream_new_from_file to r+b

rccarper:
*Description of changes:*
If a binary stream contains a 0x1A character, currently it will terminate early on Windows due to the missing binary flag.

### DIFF
--- a/tests/stream_test.c
+++ b/tests/stream_test.c
@@ -16,7 +16,7 @@ AWS_STATIC_STRING_FROM_LITERAL(s_simple_test, "SimpleTest");
 
 /* 0x1A represents the Windows end-of-file character. Having this in the test data set allows us to verify that file
  * stream reads on binary files do not terminate early on Windows.*/
-const uint8_t s_simple_binary_test[] = "abcdef\x1Aghijk";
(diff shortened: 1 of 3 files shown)


tests/stream_test.c
```diff
@@ -14,6 +14,10 @@
 
 AWS_STATIC_STRING_FROM_LITERAL(s_simple_test, "SimpleTest");
 
+/* 0x1A represents the Windows end-of-file character. Having this in the test data set allows us to verify that file
+ * stream reads on binary files do not terminate early on Windows.*/
+const uint8_t s_simple_binary_test[] = "abcdef\x1Aghijk";
```
graebm, line 19: Defining this with quotes means you get a null terminator at the end of your byte array, did you want a null terminator? If not, consider defining it like:

const uint8_t s_simple_binary_test[] = {'a', 'b', 'c', 'd', 'e', 'f', 0x1A, 'g', 'h', 'i', 'j', 'k'};

rccarper: That's a good point, it's okay for the test, but I'll switch it to an array to be more correct.

graebm (approved):

PR #379, opened 2021-04-29
Fix for shutdown crash when peer hangs up immediately following first non-TLS write

justinboswell:
Found while assisting on  https://github.com/awslabs/aws-crt-swift/pull/17

TLS negotation completes after connecting to https://example.com:443 via MQTT. When the MQTT CONNECT packet is sent, the server immediately hangs up, causing an eventual empty message in the input queue with a NULL allocator.

This was a double-release due to aliasing the incoming message pointer without notifying the upstream handler (socket, in this case) that the SecureTransport handler had taken ownership in all code paths.

### DIFF
--- a/source/darwin/secure_transport_tls_channel_handler.c
+++ b/source/darwin/secure_transport_tls_channel_handler.c
@@ -657,6 +657,7 @@ static int s_process_read_message(
         if (slot->adj_right) {
             if (aws_channel_slot_send_message(slot, outgoing_read_message, AWS_CHANNEL_DIR_READ)) {
                 aws_mem_release(outgoing_read_message->allocator, outgoing_read_message);
+                aws_channel_shutdown(secure_transport_handler->parent_slot->channel, aws_last_error());
                 /* incoming message was pushed to the input_queue, so this handler owns it now */
                 return AWS_OP_SUCCESS;
             }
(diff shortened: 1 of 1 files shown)


source/darwin/secure_transport_tls_channel_handler.c
```diff
@@ -656,7 +657,8 @@ static int s_process_read_message(
         if (slot->adj_right) {
             if (aws_channel_slot_send_message(slot, outgoing_read_message, AWS_CHANNEL_DIR_READ)) {
                 aws_mem_release(outgoing_read_message->allocator, outgoing_read_message);
-                return AWS_OP_ERR;
+                /* incoming message was pushed to the input_queue, so this handler owns it now */
+                return AWS_OP_SUCCESS;
```
JonathanHenson, line 661: Do we need to explicitly shutdown?

justinboswell: Good question, we probably do, let me check

source/darwin/secure_transport_tls_channel_handler.c
```diff
@@ -612,7 +612,8 @@ static int s_process_read_message(
         struct aws_io_message *outgoing_read_message = aws_channel_acquire_message_from_pool(
             slot->channel, AWS_IO_MESSAGE_APPLICATION_DATA, downstream_window - processed);
         if (!outgoing_read_message) {
-            return AWS_OP_ERR;
+            /* even though this is a failure, this handler has taken ownership of the message */
```
JonathanHenson, line 615: don't you need to shutdown here as well? this assumes an allocation failed.

JonathanHenson (approved):

source/darwin/secure_transport_tls_channel_handler.c
justinboswell: Sure, for now. I'll add it, but this should crash and die before ever returning here.

PR #383, opened 2021-05-06
Another way of addressing the pem buffer issue

bretambrose:
* Since s2n is the only supported tls context that needs zero-terminated strings, limit the fix to it, right before the api calls.


justinboswell (approved):
This is what we've done in some of the CRTs, this would render that code redundant (see aws-crt-java/python for instance, IIRC)

PR #386, opened 2021-05-25
Improved Posix Large File Support (LFS)

graebm:
Mildly breaking API change!!! All uses of `aws_off_t` replaced by `int64_t`

### ISSUE:
Bizarre behavior on 32bit Raspberry Pi OS when aws_input_stream_seek() was called.

### CAUSE:
Conditional logic in header files is bad.

The size of `off_t` on Posix is conditional, based on various defines when headers are included.

It's possible for .c files in different libraries to disagree on the size of `off_t`, which means they disagree on the size of `aws_off_t`, which leads to stack corruption when functions taking `aws_off_t` are called.

### CHANGES:
1) Always use `int64_t` for offsets, instead of `aws_off_t`. Deal with 32 vs 64 bit offsets in .c files we control, not headers included by external code. I will fix up downstream libraries in followup PRs.
2) AwsCFlags now attempts to enable LFS on 32 bit system, see https://github.com/awslabs/aws-c-common/pull/808


source/posix/file_utils.c
```diff
@@ -34,14 +34,17 @@ bool aws_path_exists(const char *path) {
...
+#ifdef AWS_HAVE_POSIX_LARGE_FILE_SUPPORT
+    int result = fseeko(file, offset, whence);
 #else
-        fseek(file, offset, whence);
-#endif
+    /* must use fseek(), which takes offset as a long */
+    if (offset < LONG_MIN || offset > LONG_MAX) {
+        return aws_raise_error(AWS_IO_STREAM_INVALID_SEEK_POSITION);
+    }
+    int result = fseek(file, offset, whence);
```
bretambrose, line 46: Are there any worries about a relative seek going pass the 32 bit boundary or is that something that will result in an error from fseek itself?

bretambrose (approved):

source/posix/file_utils.c
graebm: I'm going to assume that's handled by fseek. We're just sanitizing the inputs

graebm: Confirmed, if I do two `fseek(f, LONG_MAX, SEEK_CUR);` in a row the 2nd call fails

bretambrose (approved):

PR #389, opened 2021-05-28
Adding is-valid function for aws_tls_connection_options

rccarper:
*Description of changes:*
Adding is-valid function for aws_tls_connection_options so that it can be used by our other libraries.


include/aws/io/tls_channel_handler.h
```diff
@@ -474,6 +474,9 @@ AWS_IO_API int aws_tls_connection_options_set_alpn_list(
     struct aws_allocator *allocator,
     const char *alpn_list);
 
+/* Returns true if the given TLS connection options have valid fields, false if not.*/
+AWS_IO_API bool aws_tls_connection_options_is_valid(const struct aws_tls_connection_options *conn_options);
```
graebm, line 478: It shouldn't be possible for an aws_tls_connection_options to be invalid. The only "constructor" we offer in C is `void aws_tls_connection_options_init_from_ctx()`, which takes the ctx as an argument.

In our C code, we generally haven't checked that, if a struct arg was passed to us, that it's in a valid state.

This is only an issue in aws-crt-cpp because `TlsConnectionOptions` offers a default constructor that creates an "invalid" tls_connection_options. Could we keep this check local to aws-crt-cpp? Like, anywhere a `TlsConnectionOptions` is passed in, we check that it's valid? Since that's the only place where "invalid" is really a concept

PR #390, opened 2021-06-01
Fix disabling verify_peer on Windows

bretambrose:
* SCH_CRED_AUTO_CRED_VALIDATION and SCH_CRED_MANUAL_CRED_VALIDATION are separate flags that are technically mutually exclusive but the win32 api won't tell you if you enabled them both and instead just pretend like SCH_CRED_AUTO_CRED_VALIDATION was set.  

End result: on Windows, certs were always verified no matter what.


source/windows/secure_channel_tls_handler.c
```diff
@@ -1946,6 +1946,7 @@ struct aws_tls_ctx *s_ctx_new(
             "static: x.509 validation has been disabled. "
             "If this is not running in a test environment, this is likely a security vulnerability.");
 
+        secure_channel_ctx->credentials.dwFlags &= ~(SCH_CRED_AUTO_CRED_VALIDATION);
```
graebm, line 1949: Your change looks good. But following the flow of this function, I can't really figure out what's going on if we're a server.

Like, I see that if `server && verify_peer && ca_file` then we set `SCH_CRED_MANUAL_CRED_VALIDATION`, but but if we're just `server && verify_peer` then it doesn't look like we set any validation flags at all

Issue #398, opened 2021-06-18
Windows handle leak


kankri:
We have noticed that our application which uses aws-crt-cpp is leaking Windows thread handles in aws-c-io.

The issue can easily be caught with _Application Verifier_ using the following steps. It would be nice if you could integrate [Application Verifier](https://docs.microsoft.com/en-us/windows-hardware/drivers/devtest/application-verifier), available in [Windows Software Development Kit](https://developer.microsoft.com/en-us/windows/downloads/windows-10-sdk/), into your test sets.

I checked out tag v1.12.6 of aws-iot-device-sdk-cpp-v2 and made the following changes to convert `raw_pub_sub` into a DLL and add a runner for it:

```diff
diff --git a/samples/mqtt/raw_pub_sub/CMakeLists.txt b/samples/mqtt/raw_pub_sub/CMakeLists.txt
index e13dd0a..1d001d7 100644
--- a/samples/mqtt/raw_pub_sub/CMakeLists.txt
+++ b/samples/mqtt/raw_pub_sub/CMakeLists.txt
@@ -6,7 +6,7 @@ file(GLOB SRC_FILES
        "*.cpp"
 )
 
-add_executable(${PROJECT_NAME} ${SRC_FILES})
+add_library(${PROJECT_NAME} SHARED ${SRC_FILES})
 
 set_target_properties(${PROJECT_NAME} PROPERTIES
     CXX_STANDARD 14)
diff --git a/samples/mqtt/raw_pub_sub/main.cpp b/samples/mqtt/raw_pub_sub/main.cpp
index 0a39efa..2800dab 100644
--- a/samples/mqtt/raw_pub_sub/main.cpp
+++ b/samples/mqtt/raw_pub_sub/main.cpp
@@ -63,6 +63,7 @@ char *s_getCmdOption(char **begin, char **end, const String &option)
     return 0;
 }
 
+extern "C" __declspec(dllexport)
 int main(int argc, char *argv[])
 {
 
diff --git a/samples/mqtt/runner/CMakeLists.txt b/samples/mqtt/runner/CMakeLists.txt
new file mode 100644
index 0000000..00e2f6f
--- /dev/null
+++ b/samples/mqtt/runner/CMakeLists.txt
@@ -0,0 +1,20 @@
+cmake_minimum_required(VERSION 3.1)
+# note: cxx-17 requires cmake 3.8, cxx-20 requires cmake 3.12
+project(runner CXX)
+
[...]

PR #399, opened 2021-06-18
Fix thread handle leak on Windows

kankri:
* Tentative fix for issue #398

* Removed unused 'resolver_thread' in 'struct host_entry'
   * `host_entry->resolver_thread` was used only in function `create_and_init_host_entry()`, converted it to a local variable.
* Clean up 'resolver_thread' to avoid leaking a thread handle on Windows
   * The thread handle inside `struct aws_thread` was never closed, calling `aws_thread_clean_up()` fixes this.
   * There is a copy of the thread struct in `thread_wrapper` with a unique thread handle which is used to synchronize thread exit and it gets properly closed.
   * I noticed that the semantics of `aws_thread_clean_up()` on Linux are a bit different: it calls `pthread_detach()`. Not sure if this matters.

----

### DIFF
--- a/source/host_resolver.c
+++ b/source/host_resolver.c
@@ -1364,6 +1364,8 @@ static inline int create_and_init_host_entry(
     aws_thread_launch(&resolver_thread, resolver_thread_fn, new_host_entry, &thread_options);
     ++default_host_resolver->pending_host_entry_shutdown_completion_callbacks;
 
+    aws_thread_clean_up(&resolver_thread);
+
     return AWS_OP_SUCCESS;
 
 setup_host_entry_error:
--- a/source/host_resolver.c
+++ b/source/host_resolver.c
@@ -210,7 +210,6 @@ struct host_entry {
     /* immutable post-creation */
     struct aws_allocator *allocator;
     struct aws_host_resolver *resolver;
-    struct aws_thread resolver_thread;
     const struct aws_string *host_name;
     int64_t resolve_frequency_ns;
     struct aws_host_resolution_config resolution_config;
@@ -1346,7 +1345,9 @@ static inline int create_and_init_host_entry(
     new_host_entry->resolution_config = *config;
     aws_condition_variable_init(&new_host_entry->entry_signal);
 
-    if (aws_thread_init(&new_host_entry->resolver_thread, resolver->allocator)) {
+    struct aws_thread resolver_thread;
+    AWS_ZERO_STRUCT(resolver_thread);
+    if (aws_thread_init(&resolver_thread, resolver->allocator)) {
         goto setup_host_entry_error;
     }
 
@@ -1360,7 +1361,7 @@ static inline int create_and_init_host_entry(
     struct aws_thread_options thread_options = *aws_default_thread_options();
     thread_options.join_strategy = AWS_TJS_MANAGED;
 
-    aws_thread_launch(&new_host_entry->resolver_thread, resolver_thread_fn, new_host_entry, &thread_options);
+    aws_thread_launch(&resolver_thread, resolver_thread_fn, new_host_entry, &thread_options);
     ++default_host_resolver->pending_host_entry_shutdown_completion_callbacks;
 
     return AWS_OP_SUCCESS;
@@ -1368,7 +1369,7 @@ static inline int create_and_init_host_entry(
 setup_host_entry_error:
 
     if (thread_init) {
-        aws_thread_clean_up(&new_host_entry->resolver_thread);
+        aws_thread_clean_up(&resolver_thread);
     }
 
     s_clean_up_host_entry(new_host_entry);


source/host_resolver.c
```diff
@@ -1360,15 +1361,17 @@ static inline int create_and_init_host_entry(
     struct aws_thread_options thread_options = *aws_default_thread_options();
     thread_options.join_strategy = AWS_TJS_MANAGED;
 
-    aws_thread_launch(&new_host_entry->resolver_thread, resolver_thread_fn, new_host_entry, &thread_options);
+    aws_thread_launch(&resolver_thread, resolver_thread_fn, new_host_entry, &thread_options);
```
TingDaoK, line 1364: Thank you for your PR.

> You must either call join or detach after creating the thread and before calling clean_up.

Referring to [here](https://github.com/awslabs/aws-c-common/blob/main/include/aws/common/thread.h#L119). I think we cannot just make the thread a local variable and clean it up like that. But seems like we do forget about thread clean up when clean up the host_entry. Not super sure, still investigating.

TingDaoK: Maybe instead of making it a local variable, add `aws_thread_clean_up(&entry->resolver_thread)` to `s_clean_up_host_entry`?

bretambrose:
I think we've been using aws_thread_clean_up in an inconsistent fashion (timing/lifecycle wise) within the codebase and while addressing the core issue (handle leak) that inconsistency should be resolved as well.

We want to factor in the managed thread concept (which was created to allow for threads that weren't "owned" by a particular object but were still guaranteed to be joined somehow, somewhere given a core API call (aws_thread_join_all_managed)).

I also worry about the caveat in https://docs.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-waitforsingleobject ("If this handle is closed while the wait is still pending, the function's behavior is undefined.")  It's not 100% clear if the undefined behavior is limited to a handle that was closed while the Wait was happening or also before the Wait was called.  But just to be cautious, I would assume that a Wait on a closed handle was potentially dangerous too (although it's true we are doing that in places right now).

So here's what I'm currently toying with as a contract:

For managed threads (host resolver threads and event loop group async shutdown threads):
   The managed thread logic will invoke aws_thread_clean_up after the thread has been joined (https://github.com/awslabs/aws-c-common/blob/main/source/windows/thread.c#L56).  This would remove a call to aws_thread_clean_up in https://github.com/awslabs/aws-c-io/blob/main/source/event_loop.c#L76 and cause the host resolver threads to get clean up called on them.  I've done some preliminary testing and it looks positive so far.

For unmanaged threads (everything else):
   aws_thread_clean_up must be called and must only be called after any desired join behavior

The consequences I see is that we're doing things backwards in a number of our tests which will now need to call cleanup at the end of their work rather than in the middle.

source/host_resolver.c
kankri: It seemed to me that when creating a managed thread calling join and clean up will be taken care of by the thread wrapper...

kankri:
>  I had forgotten that we actually allocate multiple Windows handles when launching a thread: one is the handle returned by CreateThread, but the other is a handle opened within the thread's wrapper function so that it can be joined eventually by the managed thread logic once the thread function completes.

Yes, this is what I was referring to by this in my description:
> There is a copy of the thread struct in thread_wrapper with a unique thread handle which is used to synchronize thread exit and it gets properly closed.

Regarding:
> "If this handle is closed while the wait is still pending, the function's behavior is undefined."

Since the copy of the handle in the wrapper is used for safe thread joining as far as I can see, it shouldn't matter if the handle returned to the creator of the thread is closed: since it is never used for anything, it is not used for WaitForSingleThread() either.

I believe my change causes the correct amount of CloseHandle()s and thread handle waiting and joining, but I can't claim I fully understand all the details of managed vs non-managed threads in the SDK. And I was also worried if calling aws_thread_clean_up () in Linux has some different effects than on Windows (where it just closes a handle but doesn't otherwise affect the thread).

So I made this PR just in case the simple fix is OK, but if it isn't so, please go ahead and make an alternative PR.

bretambrose:
> I believe my change causes the correct amount of CloseHandle()s and thread handle waiting and joining, but I can't claim I fully understand all the details of managed vs non-managed threads in the SDK. And I was also worried if calling aws_thread_clean_up () in Linux has some different effects than on Windows (where it just closes a handle but doesn't otherwise affect the thread).
> 
> So I made this PR just in case the simple fix is OK, but if it isn't so, please go ahead and make an alternative PR.

Your PR absolutely fixes the host resolver thread handle leak; there's nothing wrong with it.  But from a more general perspective, it's suddenly clear that how you (or anyone at the C layer) should use aws_thread_clean_up() is very unclear and I would like the full solution to make that consistent as well.

My intent with the managed threads is that they should be something that you just create and forget about, including both joins and clean up (detach).  I think it's unfortunate that the original thread api didn't distinguish between detach and clean up and instead conflated the two but that's what we have to live with now I guess.

So, in summary, issue is spot-on, analysis and verification steps spot-on, PR fixes the issue, but I would like to go with mine because I think it makes the thread contract clearer (otherwise everyone using a managed thread still needs to call clean up after launching even though clean up is considered an end-of-lifecycle operation everywhere in the aws-c-* libraries).  We'll see what others says in review; I certainly understand that an either/or contract (if managed do this else do that) is not optimal.

kankri:
Yes, I agree that it makes sense to clarify the contract and it is redundant for `aws_thread_launch()` to return an "active" thread object for the caller when creating a managed thread. Your PR awslabs/aws-c-common#814 solves this by calling `aws_thread_clean_up()` before returning from `aws_thread_launch()`. So I'm happy if this is closed in favor of your 2 pull requests.

bretambrose:
Thanks again for tracking this down.  I've merged the hopefully-clarifying multi-repo fix for this in common and io and made new releases.  The C++ CRT release may be delayed slightly (a day or two) in order to bring in other fixes.

PR #403, opened 2021-06-24
Address override_default_trust_store() functions still using default trust store.

graebm:
### Issue
The “OverrideDefaultTrustStore” functions did not actually override the system trust store on all platforms. Instead, the functions augmented the system trust store with a user-supplied CA on Linux and Apple platforms.

### Description of changes
The "OverrideDefaultTrustStore" functions actually override the system's default trust store on all platforms.


source/tls_channel_handler.c
```diff
@@ -62,82 +62,49 @@ void aws_tls_ctx_options_clean_up(struct aws_tls_ctx_options *options) {
     AWS_ZERO_STRUCT(*options);
 }
 
-static int s_load_null_terminated_buffer_from_cursor(
```
graebm, line 65: We don't need this function anymore.
A while back Bret made is so we turn this into a null-terminated string right before passing it into s2n, instead of fastidiously trying to keep an untracked 0 at the end of the aws_byte_buf in N different places

source/tls_channel_handler.c
```diff
@@ -62,82 +62,49 @@ void aws_tls_ctx_options_clean_up(struct aws_tls_ctx_options *options) {
...
-    /* PEM is not supported on iOS */
-#if !defined(AWS_OS_IOS)
-    if (!options) {
-        return AWS_OP_SUCCESS;
-    }
-    if (options->allocator) {
-        if (aws_sanitize_pem(&options->ca_file, options->allocator) |
-            aws_sanitize_pem(&options->certificate, options->allocator) |
-            aws_sanitize_pem(&options->private_key, options->allocator)) {
-            return AWS_OP_ERR;
```
graebm, line 99: Instead of re-sanitizing N pem files whenever 1 of them is set, I just made sure to aws_sanitize_pem() each of these files when we set them.

justinboswell (approved):

source/tls_channel_handler.c
```diff
@@ -146,16 +115,29 @@ int aws_tls_ctx_options_init_client_mtls_from_path(
...
-    s_tls_ctx_options_pem_sanitize(options);
+
+    if (aws_sanitize_pem(&options->private_key, allocator)) {
+        AWS_LOGF_ERROR(AWS_LS_IO_TLS, "static: Invalid private key. File must contain PEM encoded data");
+        goto error;
+    }
+
     return AWS_OP_SUCCESS;
+error:
+    aws_tls_ctx_options_clean_up(options);
```
bretambrose, line 137: I went ahead and verified that this was actually safe, and it is, but I'm wondering are there other places in the whole code base that should be calling cleanup on the init object on a failure.  I have it in my mind that we do that for new/delete/release creation patterns but was surprised by this.  It makes sense but requires a zero-struct at the end of clean up.

source/tls_channel_handler.c
```diff
@@ -326,54 +299,87 @@ void aws_tls_ctx_options_set_minimum_tls_version(
...
+        ca_path_tmp = aws_string_new_from_c_str(options->allocator, ca_path);
+        if (!ca_path_tmp) {
+            goto error;
         }
     }
 
     if (ca_file) {
-        if (aws_byte_buf_init_from_file(&options->ca_file, options->allocator, ca_file)) {
-            return AWS_OP_ERR;
+        if (options->ca_file.capacity) {
```
bretambrose, line 325: why capacity vs. len?

source/tls_channel_handler.c
```diff
@@ -326,54 +299,87 @@ void aws_tls_ctx_options_set_minimum_tls_version(
...
- * The following API is deprecated due to inconsistent override versus append system/library trust
- * behavior between Windows, Mac, and Unix based platforms
- */
 int aws_tls_ctx_options_override_default_trust_store_from_path(
     struct aws_tls_ctx_options *options,
     const char *ca_path,
     const char *ca_file) {
 
+    struct aws_string *ca_path_tmp = NULL;
+    struct aws_byte_buf ca_file_tmp;
```
bretambrose, line 308: Let's add a note here that on success, ca_file_tmp gets copied into a persistent structure by value, so there's no need to clean it up.  This is sort of a weird (in that we don't do it often) pattern but I can see its value in making the change atomically all-or-nothing.

bretambrose (approved):

source/tls_channel_handler.c
graebm: our error handling patterns are not consistent through the whole codebase. I've definitely settled on this as the preferred strategy, which is why I deployed it here.

Yeah, our clean_up() functions are supposed to be idempotent, and zero-struct at the end of clean_up() really helps with that.

source/tls_channel_handler.c
graebm: we discussed. I'm changing all `.len` and `.capacity` checks to use a helper function instead.

There were all kinds of bugs (leaks, unexpected behavior, etc) that could happen if someone passed in a 0-length file or cursor

bretambrose (approved):

PR #405, opened 2021-07-23
Add TLS 1.3 support and PQ TLS 1.3 support

alexw91:
*Issue #, if available:*
N/A

*Description of changes:*
Adds regular TLS 1.3 support and PQ TLS 1.3 support. 

`CloudFront-TLS-1-0-2016` is a much better default than `ELBSecurityPolicy-TLS-1-1-2017-01` for CRT clients, it has both more algorithms for wider backwards compatibility, and has more modern algorithms for higher security.

`CloudFront-TLS-1-0-2016` has the following benefits over  `ELBSecurityPolicy-TLS-1-1-2017-01`:
 - Supports TLS 1.3
 - Support for RSA PSS Signatures
 - Support for x25519 Elliptic Curves
 - Supports ChaCha-Poly Ciphers
 - Supports TLS 1.0

Longer term, I think s2n should have CRT-specific TLS Security policies for consistency across different TLS versions, but this is a good step in the right direction.


alexw91:
Splitting this into two Pull Requests.

PR #408, opened 2021-07-30
PKCS#11 public API 1st pass

graebm:
First pass at public-facing API for interacting with PKCS#11 libraries.
No implementation yet.
Note that I'm merging into a `pkcs11` feature-development branch.

Design decisions:
- User must manually create `aws_pkcs11_lib_handle`.
  - This gives the user control over when and how the library is initialized and finalized.
- Use options-struct to pass args into `aws_tls_ctx_options_init_client_mtls_with_pkcs11()`
  - This gives us flexibility to add more optional args in the future, without changing the function signature or introducing a billion overloaded variants of the function.
  - Things I'm not sure about, that this options-struct gives us flexibility to figure out in the future:
    - Are customers storing certificates in PKCS#11, or as normal files on disk? I see demos doing both.
    - Are customers identifying their keys by Label or ID? PKCS#11 doesn't require that either attribute be set.
  - I do hate that I made an args-like options struct to help initialize a class-like options struct though 😡


include/aws/io/tls_channel_handler.h
```diff
@@ -279,6 +280,57 @@ AWS_IO_API int aws_tls_ctx_options_init_client_mtls(
     const struct aws_byte_cursor *cert,
     const struct aws_byte_cursor *pkey);
 
+/**
+ * This struct exists as a graceful way to pass many arguments when
+ * calling init-with-pkcs11 functions on aws_tls_ctx_options (this also makes
+ * it easy to introduce optional arguments in the future).
+ * Instances of this struct should only exist briefly on the stack.
+ * Instructions for binding this to high-level languages:
+ * - Python: The members of this struct should be the keyword args to the init-with-pkcs11 functions.
```
JonathanHenson, line 289: I love you Mike

include/aws/io/tls_channel_handler.h
```diff
@@ -279,6 +280,57 @@ AWS_IO_API int aws_tls_ctx_options_init_client_mtls(
...
+
+    /**
+     * ID of slot containing token.
+     */
+    uint32_t slot_id;
+
+    /**
+     * PIN for logging into the token (UTF-8).
+     */
+    const char *pin;
```
JonathanHenson, line 308: the pin likely comes from user-software land like a gui prompt or a button press right? in that case, should this be an aws_string?

JonathanHenson (approved):
just a question on if string literals are actually the only thing at play here

include/aws/io/pkcs11.h
```diff
@@ -0,0 +1,64 @@
...
+ */
+struct aws_pkcs11_lib_handle;
+
+/**
+ * Options for aws_pkcs11_lib_handle_new()
+ */
+struct aws_pkcs11_lib_handle_options {
+    /**
+     * Name of PKCS#11 library file to load.
+     * Set to NULL if your application is compiled with PKCS#11 symbols linked in.
```
DavidOgunsAWS, line 22: Will the user be able to create handles to a linked PKCS#11 library and dynamically loaded ones?

This isn't something to solve now, but it's a subtle implementation consideration for these handles.

DavidOgunsAWS (approved):

include/aws/io/pkcs11.h
graebm: Yeah the idea is to eventually support:
1) dynamically loaded (based on filepath at runtime)
2) dynamically linked (because their program against it at build time)
3) statically linked

I think we'll get 1 & 2 immediately. 3 will take some more work but we can support it someday

include/aws/io/tls_channel_handler.h
graebm: This options-struct only exists for passing args on the stack. I don't intend to re-use the same struct for actual storage on the heap

include/aws/io/tls_channel_handler.h
```diff
@@ -279,6 +280,67 @@ AWS_IO_API int aws_tls_ctx_options_init_client_mtls(
...
+     * The certificate must be PEM formatted and UTF-8 encoded.
+     * Set NULL if passing in certificate by some other means.
+     */
+    struct aws_byte_cursor *cert_file_contents;
+
+    /**
+     * Label of private key on token (UTF-8).
+     * Set NULL if the private key has no label.
+     */
+    struct aws_byte_cursor *pkey_pkcs11_label;
```
prateek-y, line 328: Should we instead expect full pkcs11 URI ?

PR #412, opened 2021-08-06
Implement PKCS#11 library initialization

graebm:
Establish standards for logging, error handling, etc before handing off implementation:
- Be VERY thorough about logging the underlying PKCS#11 error code and name
  -   Ex: `C_GetInfo() failed. PKCS#11 error: CKR_CRYPTOKI_NOT_INITIALIZED (0x00000190). AWS error: AWS_IO_PKCS11_ERROR.`
- But just 1 catch-all `AWS_IO_PKCS11_ERROR` for now
  - We can always add more specific errors later, but there are nearly 100 named PKCS#11 errors
- Use `TEST_PKCS11_LIB=<path>` env-var to enable tests
  - Seemed simpler than passing through CMAKE variables
  - We might need to add more variables later like _SLOT, _PKEY_LABEL, etc
  - I'll tackle getting these to run in CI later...

Also, shorten ~~aws_pkcs11_lib_handle~~ -> `aws_pkcs11_lib`


justinboswell (approved):

source/pkcs11.c
```diff
@@ -12,3 +20,359 @@
...
+
+/* Callback for PKCS#11 library to lock a mutex.
+ * Described in PKCS11-base-v2.40 section 3.7 */
+static CK_RV s_pkcs11_lock_mutex(CK_VOID_PTR mutex_ptr) {
+    if (mutex_ptr == NULL) {
+        return CKR_GENERAL_ERROR;
+    }
+
+    struct aws_mutex *mutex = mutex_ptr;
+    if (aws_mutex_lock(mutex)) {
```
prateek-y, line 200: What is the behavior when this mutex has been already acquired? Do we block or fail?

source/pkcs11.c
```diff
@@ -12,3 +20,359 @@
...
+    aws_shared_library_clean_up(&pkcs11_lib->shared_lib);
+    aws_mem_release(pkcs11_lib->allocator, pkcs11_lib);
+}
+
+struct aws_pkcs11_lib *aws_pkcs11_lib_new(
+    struct aws_allocator *allocator,
+    const struct aws_pkcs11_lib_options *options) {
+
+    /* Create the struct */
+    struct aws_pkcs11_lib *pkcs11_lib = aws_mem_calloc(allocator, 1, sizeof(struct aws_pkcs11_lib));
```
prateek-y, line 268: Curios, how are memory allocation failures handled?

source/pkcs11.c
```diff
@@ -12,3 +20,359 @@
...
+        filename = aws_string_c_str(filename_storage);
+    }
+
+    AWS_LOGF_DEBUG(
+        AWS_LS_IO_PKCS11,
+        "Loading PKCS#11. file:'%s' C_Initialize:%s",
+        filename ? filename : "<MAIN PROGRAM>",
+        options->omit_initialize ? "omit" : "yes");
+
+    if (aws_shared_library_init(&pkcs11_lib->shared_lib, filename)) {
```
prateek-y, line 289: Is it ok to pass NULL filename here?

source/pkcs11.c
```diff
@@ -12,3 +20,359 @@
...
+            .flags = CKF_OS_LOCKING_OK,
+        };
+
+        CK_RV rv = pkcs11_lib->function_list->C_Initialize(&init_args);
+        if (rv != CKR_OK) {
+            s_raise_ck_error(pkcs11_lib, "C_Initialize", rv);
+            goto except;
+        }
+
+        pkcs11_lib->should_finalize = true;
```
prateek-y, line 324: do we need to call finalize even if initialize fails, to do partial cleanup?

source/pkcs11.c
```diff
@@ -12,3 +20,359 @@
...
+    }
+
+    /* Get info about the library and log it.
+     * This will be VERY useful for diagnosing user issues. */
```
prateek-y, line 328: Nice

source/pkcs11.c
```diff
@@ -12,3 +20,359 @@
 #define CK_DECLARE_FUNCTION_POINTER(returnType, name) returnType(CK_PTR name)
 #define CK_CALLBACK_FUNCTION(returnType, name) returnType(CK_PTR name)
 #include <aws/io/private/pkcs11/v2.40/pkcs11.h>
+
+/* Return c-string for PKCS#11 CKR_* contants */
+const char *s_ckr_str(CK_RV rv) {
+    /* clang-format off */
+    switch (rv) {
```
prateek-y, line 27: How do you get this list?

source/pkcs11.c
graebm: It just calls through to [pthread_mutex_lock()](https://man7.org/linux/man-pages/man3/pthread_mutex_lock.3p.html#ERRORS) on Unix or [AcquireSRWLockExclusive()](https://docs.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-acquiresrwlockexclusive) on Windows, so it's going to do whatever they would do.

So if the mutex has been acquired on another thread, it will block.
If it has already been acquired on this thread it will deadlock.

source/pkcs11.c
graebm: if the allocation would fail, the process is aborted.

For a long time we wrote code that tried to handle out-of-memory. But it was **extremely cumbersome** and the error handling code we wrote was largely untested. We later decided to just make it fatal, for the sake of keeping our code maintainable and testable.

source/pkcs11.c
graebm: NULL is allowed.
You do this if the symbol can be found in THIS PROGRAM
So like, if you link against libsofthsm2.so file at build time, so the OS automatically loads it along with your program, then you can do: `dlsym(dlopen(NULL), "C_GetFunctionList")` to get at the symbols at runtime.

(I had to do experiments to confirm it worked this way)

graebm: It's ok to load the library multiple times. from the [docs](https://man7.org/linux/man-pages/man3/dlopen.3.html)
> The dynamic linker maintains reference counts for object handles, so a dynamically loaded shared object is not deallocated until dlclose() has been called on it as many times as dlopen() has succeeded on it

source/pkcs11.c
graebm: huh. good question.

usually in C APIs, if an "init" call fails it's like nothing ever happened, so my instinct would be that it's not necessary

Looking at [SoftHSM's implementation](https://github.com/opendnssec/SoftHSMv2/blob/3de59817e17bd1873f1e802757b5b38a1c7ca898/src/lib/SoftHSM.cpp#L613), if the `C_Initialize()` call fails then `C_Finalize()` will also fail with `CKR_CRYPTOKI_NOT_INITIALIZED`

source/pkcs11.c
graebm: I copy-pasted from [pkcs11t.h](https://github.com/oasis-tcs/pkcs11/blob/0c3ecef6c90dcd8249ba64cc99d50d9f8f73124e/published/2-40-errata-1/pkcs11t.h#L1039)

source/pkcs11.c
```diff
@@ -12,3 +20,359 @@
...
+        "Loading PKCS#11. file:'%s' C_Initialize:%s",
+        filename ? filename : "<MAIN PROGRAM>",
+        options->omit_initialize ? "omit" : "yes");
+
+    if (aws_shared_library_init(&pkcs11_lib->shared_lib, filename)) {
+        goto except;
+    }
+
+    /* Find C_GetFunctionList() and call it to get the list of pointers to all the other functions */
+    CK_C_GetFunctionList get_function_list = NULL;
```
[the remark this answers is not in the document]
graebm: I don't think that's possible without actually calling each function? From [the docs](http://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/os/pkcs11-base-v2.40-os.html):

> If a particular function in the Cryptoki API is not supported by a library, then the function pointer for that function in the library’s CK_FUNCTION_LIST structure should point to a function stub which simply returns CKR_FUNCTION_NOT_SUPPORTED.

source/pkcs11.c
prateek-y: makes sense

source/pkcs11.c
prateek-y: so with a new version, we would need to update this as well.

source/pkcs11.c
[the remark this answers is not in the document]
prateek-y: Hopefully we will be fine, as we just need a handful of functions.

include/aws/io/io.h
```diff
@@ -134,6 +134,7 @@ enum aws_io_errors {
     AWS_IO_TLS_ALERT_NOT_GRACEFUL,
     AWS_IO_MAX_RETRIES_EXCEEDED,
     AWS_IO_RETRY_PERMISSION_DENIED,
+    AWS_IO_PKCS11_ERROR,
```
prateek-y, line 137: Do you see a distinction here between user and service errors? Like CKR_BUFFER_TOO_SMALL is our problem and CKR_PIN_INCORRECT is users problem.

prateek-y (approved):

include/aws/io/io.h
graebm: Yeah, this is the part of the PR I'm least confident in, and totally open to changing it going forward.

Each of our C libraries is limited to 1024 unique error codes. Adding every single PKCS#11 error code like: `CKR_UNWRAPPING_KEY_TYPE_INCONSISTENT -> AWS_IO_PKCS11_ERROR_UNWRAPPING_KEY_TYPE_INCONSISTENT`
would take up about 100 slots, so it's possible.

Hopefully issues like CKR_BUFFER_TOO_SMALL won't surface because we'll make the calls correctly. But anything can happen because each vendor has their own implementation.

PR #413, opened 2021-08-09
PKCS#11 CI - WIP

graebm:


tests/pkcs11_test.c
```diff
@@ -3,13 +3,59 @@
  * SPDX-License-Identifier: Apache-2.0.
  */
 
+/**
```
justinboswell, line 6: Do you think PKCS11.md in the root or something might be a good place for this? Especially so we can link to it from downstream packages?

.builder/actions/pkcs11_test_setup.py
```diff
@@ -0,0 +1,159 @@
...
+        self._setenv('TEST_PKCS11_TOKEN_LABEL', 'my-test-token')
+        self._setenv('TEST_PKCS11_PIN', '0000')
+        self._setenv('TEST_PKCS11_PKEY_LABEL', 'my-test-key')
+        self._setenv('TEST_PKCS11_CERT_FILE', os.path.join(resources_dir, 'unittests.crt'))
+        self._setenv('TEST_PKCS11_CA_FILE', os.path.join(resources_dir, 'unittests.crt'))
+
+
+    def _find_softhsm_lib(self):
+        """Return path to SoftHSM2 shared lib, or None if not found"""
+        for lib_dir in ['lib64', 'lib']: # search lib64 before lib
```
justinboswell, line 79: This won't work for LD_LIBRARY_PATH or an ld.conf script, which some package managers will screw with.

something like `ldconfig -p` is what you'd normally use to find where the system thinks libraries are

justinboswell (commented):
LGTM, just a few issues

tests/pkcs11_test.c
graebm: Great idea. Doing it.

PR #418, opened 2021-08-18
Improve error message

ilevyor:
*Issue #, if available:*

*Description of changes:*

Create better error messages for windows and OSX errors. See following article for explanation of windows error changes:

https://docs.microsoft.com/en-us/windows/win32/seccrypto/retrieving-error-messages
https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-formatmessage


source/windows/windows_pki_utils.c
```diff
@@ -123,12 +127,16 @@ int aws_load_cert_from_system_cert_store(const char *cert_path, HCERTSTORE *cert
         *cert_store, X509_ASN_ENCODING | PKCS_7_ASN_ENCODING, 0, CERT_FIND_HASH, &cert_hash, NULL);
 
     if (!*certs) {
+        WCHAR wszMsgBuff[512];
```
justinboswell, line 130: This feels like a function, e.g. `bool aws_win_format_message(char *buffer, size_t buffer_size, DWORD lastError)`

source/darwin/secure_transport_tls_channel_handler.c
```diff
@@ -540,7 +540,11 @@ static int s_process_write_message(
 
     if (status != noErr) {
         AWS_LOGF_DEBUG(
-            AWS_LS_IO_TLS, "id=%p: SSLWrite failed with OSStatus error code %d.", (void *)handler, (int)status);
+            AWS_LS_IO_TLS,
+            "id=%p: SSLWrite failed with OSStatus error code %d (%s).",
+            (void *)handler,
+            (int)status,
+            CFStringGetCStringPtr(SecCopyErrorMessageString(status, NULL), kCFStringEncodingUTF8));
```
justinboswell, line 547: Is this the only place where we need to decode an error in the secure transport code? Feels like there might be more?

graebm: Also, here's some complication you need to  worry about

From the [SecCopyErrorMessageString docs](https://developer.apple.com/documentation/security/1394686-seccopyerrormessagestring?language=objc):
> returns NULL if no string is available

and also

> Call the CFRelease function to release this object when you are finished using it

And beware what you pass to [CFRelease](https://developer.apple.com/documentation/corefoundation/1521153-cfrelease?language=objc):
> This value must not be NULL. 

oh yeah, [CFStringGetCStringPtr](https://developer.apple.com/documentation/corefoundation/1542133-cfstringgetcstringptr?language=objc) can fail too:
> Whether or not this function returns a valid pointer or NULL depends on many factors, all of which depend on how the string was created and its properties
so maybe we should use [CFStringGetCString](https://developer.apple.com/documentation/corefoundation/1542721-cfstringgetcstring?language=objc) instead which copies into a buffer, but won't randomly fail.

so you'll probably need to do a bunch of horrible error checking like
```
CFStringRef msg_cfstring = SecCopyMessageString(status, NULL);
const char msg_buf[512];
bool msg_buf_valid = false;
if (msg_cfstring) {
    msg_buf_valid = CFStringGetCString(msg_cfstring, msg_buf, sizeof(msg_buf), kCFStringEncodingUTF8);
}
if (!msg_buf_valid) {
   strncpy(msg_buf, "Unknown", sizeof(msg_buf));
}

AWS_LOGF_DEBUG("blah blah %s",...., msg_buf);

if (msg_cfstring) {
    CFRelese(msg_cfstring);
}


```

source/windows/windows_pki_utils.c
```diff
@@ -88,11 +88,15 @@ int aws_load_cert_from_system_cert_store(const char *cert_path, HCERTSTORE *cert
         CERT_STORE_PROV_SYSTEM_A, 0, (HCRYPTPROV)NULL, CERT_STORE_OPEN_EXISTING_FLAG | store_val, store_path);
 
     if (!*cert_store) {
+        WCHAR wszMsgBuff[512];
+        FormatMessage(
+            FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS, NULL, GetLastError(), 0, wszMsgBuff, 512, NULL);
```
graebm, line 93: You should probably store off the value of GetLastError() like:

`int win_error = GetLastError();`

 Each time you call any Windows function there's a  chance GetLastError()'s value may change. So there's a chance that the call to FormatMessage() will change its value before you call GetLastError() again when logging a few lines below

source/windows/windows_pki_utils.c
```diff
@@ -88,11 +88,15 @@ int aws_load_cert_from_system_cert_store(const char *cert_path, HCERTSTORE *cert
         CERT_STORE_PROV_SYSTEM_A, 0, (HCRYPTPROV)NULL, CERT_STORE_OPEN_EXISTING_FLAG | store_val, store_path);
 
     if (!*cert_store) {
+        WCHAR wszMsgBuff[512];
```
graebm, line 91: WCHAR means 16bit "WIDE CHAR"
AWS_LOGF "%s" expects traditional 8 bit `char`

source/windows/windows_pki_utils.c
graebm: Yeah, consider a helper function that handles logging WindowsErrors. That makes it tougher to include a bunch of context in a single log statement, but it's ok if you log multiple times about one error. so like you have a helper function like:

```
// I always log: "Windows system function %s() failed with error code %d (%s)"
void s_log_win_err(const char *function_name, int win_err)
```

and you call that once to report the failed Windows call, then log again about the context it happened in. so the user's logs look like:
```
[ERROR] Windows system function CertOpenStore() failed with error code 5 (ACCESS_DENIED)
[ERROR] invalid certficate path /a/b/c/d. Failed to load cert store.
```

source/channel.c
```diff
@@ -864,12 +864,13 @@ int aws_channel_slot_shutdown(
...
         "id=%p: shutting down slot %p, with handler %p "
-        "in %s direction with error code %d",
+        "in %s direction with error code %d (%s)",
         (void *)slot->channel,
         (void *)slot,
         (void *)slot->handler,
         (dir == AWS_CHANNEL_DIR_READ) ? "read" : "write",
-        err_code);
+        err_code,
+        aws_error_str(err_code));
```
graebm, line 873: Personal preference, but I prefer `aws_error_name()` over `aws_error_str()`. But that's just me feel free to disagree

aws_error_name() will get you:

`id=0x12341234 shutting down slot 0x98769876, with handler 0x45674567 in read direction with error code  34 (AWS_ERROR_INVALID_ARGUMENT)`

aws_error_str(err_code) will get you:

`id=0x12341234 shutting down slot 0x98769876, with handler 0x45674567 in read direction with error code 34 (aws-c-common: AWS_ERROR_INVALID_ARGUMENT, An invalid argument was passed to a function.)`

source/windows/windows_pki_utils.c
graebm: maybe "%ws" is for printf'ing wide strings?

source/darwin/darwin_error_message.c
```diff
@@ -0,0 +1,29 @@
+#include "darwin_error_message.h"
+#include <Security/SecCertificate.h>
+
+void aws_darwin_log_message(
+    enum aws_log_level log_level,
+    aws_log_subject_t subject,
+    char *function_name,
```
bretambrose, line 7: Minor: this can probable be const char *

source/channel.c
justinboswell: I second this, use either aws_error_name or debug_str

justinboswell (approved):
Fix-n-ship

source/windows/windows_error_message.c
```diff
@@ -0,0 +1,21 @@
...
+void aws_win_log_message(
+    enum aws_log_level log_level,
+    aws_log_subject_t subject,
+    char *function_name,
+    DWORD last_error) {
+    WCHAR buffer[512]; // Buffer for text.
+    DWORD dw_chars;
+    dw_chars = FormatMessage(
+        FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS, NULL, last_error, 0, (LPSTR)buffer, 512, NULL);
+    AWS_LOGF(
```
bretambrose, line 14: Ideally I think we'd want to convert the message back to utf-8.  I'm not at all confident that our logging will properly handle wide string arguments.

source/windows/host_resolver.c
```diff
@@ -35,6 +36,7 @@ int aws_default_dns_resolve(
     int res_error = GetAddrInfoA(hostname_cstr, NULL, &hints, &result);
 
     if (res_error) {
+        aws_win_log_message(AWS_LL_ERROR, AWS_LS_IO_DNS, "GetAddrInfoA()", res_error);
         AWS_LOGF_ERROR(AWS_LS_IO_DNS, "static: getaddrinfo failed with error_code %d", res_error);
```
graebm, line 40: maybe just don't even bother with the 2nd LOGF statement. It's just a repeat of the new 1st one

source/windows/iocp/iocp_event_loop.c
```diff
@@ -203,11 +204,13 @@ struct aws_event_loop *aws_event_loop_new_default_with_options(
...
         1);                   /* NumberOfConcurrentThreads */
     if (impl->iocp_handle == NULL) {
+        int last_error = GetLastError();
+        aws_win_log_message(AWS_LL_FATAL, AWS_LS_IO_EVENT_LOOP, "CreateIoCompletionPort()", last_error);
         AWS_LOGF_FATAL(
             AWS_LS_IO_EVENT_LOOP,
             "id=%p: CreateIOCompletionPort failed with error %d",
             (void *)event_loop,
-            (int)GetLastError());
+            (int)last_error);
```
graebm, line 213: no point in the 2nd LOGF. The 1st is saying the same thing, but better

source/windows/iocp/iocp_event_loop.c
```diff
@@ -533,11 +536,13 @@ static int s_connect_to_io_completion_port(struct aws_event_loop *event_loop, st
...
 
     if (!iocp_associated) {
+        int last_error = GetLastError();
+        aws_win_log_message(AWS_LL_ERROR, AWS_LS_IO_EVENT_LOOP, "CreateIoCompletionPort()", last_error);
         AWS_LOGF_ERROR(
             AWS_LS_IO_EVENT_LOOP,
             "id=%p: CreateIoCompletionPort() failed with error %d",
             (void *)event_loop,
-            (int)GetLastError());
+            last_error);
```
graebm, line 545: no point in 2nd LOGF

source/windows/iocp/socket.c
```diff
@@ -596,14 +597,16 @@ static int s_ipv4_stream_connection_success(struct aws_socket *socket) {
...
         AWS_LOGF_ERROR(
             AWS_LS_IO_SOCKET,
             "id=%p handle=%p: failed to determine connection error %d",
             (void *)socket,
             (void *)socket->io_handle.data.handle,
-            (int)WSAGetLastError());
+            (int)last_error);
         socket->state = ERRORED;
-        int error = s_determine_socket_error(WSAGetLastError());
+        int error = s_determine_socket_error(last_error);
```
graebm, line 609: ok. this is my last comment. you're seeing the pattern. If the 2nd log statement is a repeat of the 1st, just cut it. If the 2nd log statement has new information (like this one reports id and handle) then log the aws-error, not just the %d of what you logged before

PR #425, opened 2021-09-16
Find PKCS#11 private key / begin s2n integration

graebm:
**Design decision:**
Use a single PKCS#11 session for all TLS connections on this s2n_ctx. We do this because PKCS#11 tokens may only support a [limited number of sessions](http://docs.oasis-open.org/pkcs11/pkcs11-ug/v2.40/cn02/pkcs11-ug-v2.40-cn02.html#_Toc406759996). Only one thread can interact with a session at a time, so this is potentially a big bottleneck.

For now, we'll keep it simple, and just use a lock to ensure the session has only 1 user at a time. If this turns out to be a major bottleneck in reality, there are many possible ways to optimize, including:
- put the single session on its own thread outside the event-loops
- 1 session per connection (depending on max supported sessions)
- 1 session per event-loop (depending on max supported sessions)

**Also lots of little helper functions:**
  - `aws_pkcs11_lib_find_slot_with_token()`
  - `aws_pkcs11_lib_open_session()`
  - `aws_pkcs11_lib_close_session()`
  - `aws_pkcs11_lib_login_user()`
  - `aws_pkcs11_lib_find_private_key()`


graebm:
A few CI checks are failing right now. I suggest we ignore them for now, so we can move ahead on the feature branch. They'll be fixed of course before 
- CI / linux-compat (opensuse-leap): failing due to docker image missing `sudo`
- CI / linux-compat (rhel8-x64): failing due to docker image missing `sudo`
- CI / linux-shared-libs: failing due to loading 2 different libcrypto.so when built this way. but probably in the real world no one is building this way

include/aws/io/private/pkcs11_private.h
```diff
@@ -7,13 +7,61 @@
...
+
 /**
  * pkcs11_private.h
  * This file declares symbols that are private to aws-c-io but need to be
  * accessed from multiple .c files.
  */
 
+/* aws_pkcs11_t: equivalent to CK_SLOT_ID / CK_SESSION_HANDLE / CK_OBJECT_HANDLE.
+ * including PKCS#11 headers is weird, so outside of pkcs11.c we'll just use this typedef instead */
+typedef unsigned long aws_pkcs11_t;
```
prateek-y, line 20: Is it better to include a hint on what type it is, "aws_pkcs11_t" does not tell me anything about it.

include/aws/io/tls_channel_handler.h
```diff
@@ -191,6 +191,19 @@ struct aws_tls_ctx_options {
...
      */
     void *ctx_options_extension;
+
+    /**
+     * Set if using PKCS#11 for private key operations.
+     * See aws_tls_ctx_pkcs11_options for more details.
+     */
+    struct {
+        struct aws_pkcs11_lib *lib;                  /* required */
+        struct aws_string *user_pin;                 /* optional */
```
prateek-y, line 201: user_pin is optional ? Is it not required for sign/decrypt operations?

source/pkcs11.c
```diff
@@ -398,3 +426,312 @@ void aws_pkcs11_lib_release(struct aws_pkcs11_lib *pkcs11_lib) {
...
+ */
+int aws_pkcs11_lib_find_slot_with_token(
+    struct aws_pkcs11_lib *pkcs11_lib,
+    const aws_pkcs11_t *match_slot_id,
+    const struct aws_string *match_token_label,
+    aws_pkcs11_t *out_slot_id) {
+
+    CK_SLOT_ID *slot_id_array = NULL;
+    CK_SLOT_ID *candidate = NULL;
+    CK_TOKEN_INFO info;
```
prateek-y, line 445: Dont you need to zero initialize this?

source/s2n/s2n_tls_channel_handler.c
```diff
@@ -924,6 +950,48 @@ static int s2n_monotonic_clock_time_nanoseconds(void *context, uint64_t *time_in
     return 0;
 }
 
+static int s_tls_ctx_pkcs11_setup(struct s2n_ctx *s2n_ctx, const struct aws_tls_ctx_options *options) {
+    /* anything initialized in this function is cleaned up during s_s2n_ctx_destroy()
+     * so don't worry about cleaning up unless it's some tmp heap allocation */
+
+    s2n_ctx->pkcs11.lib = aws_pkcs11_lib_acquire(options->pkcs11.lib);
```
prateek-y, line 957: What happens when this function fails?

source/s2n/s2n_tls_channel_handler.c
```diff
@@ -924,6 +950,48 @@ static int s2n_monotonic_clock_time_nanoseconds(void *context, uint64_t *time_in
     return 0;
 }
 
+static int s_tls_ctx_pkcs11_setup(struct s2n_ctx *s2n_ctx, const struct aws_tls_ctx_options *options) {
+    /* anything initialized in this function is cleaned up during s_s2n_ctx_destroy()
+     * so don't worry about cleaning up unless it's some tmp heap allocation */
+
```
prateek-y, line 956: Is there a need to sanitize options here, to make sure it has everything we need.

prateek-y: nvm, I see the validation later.

source/pkcs11.c
```diff
@@ -398,3 +426,312 @@ void aws_pkcs11_lib_release(struct aws_pkcs11_lib *pkcs11_lib) {
...
+ */
+int aws_pkcs11_lib_find_private_key(
+    struct aws_pkcs11_lib *pkcs11_lib,
+    aws_pkcs11_t session_handle,
+    const struct aws_string *match_label,
+    aws_pkcs11_t *out_key_object_handle) {
+
+    bool success = false;
+
+    /* whether C_FindObjectsFinal() must be run */
```
prateek-y, line 649: How do you decide this? Isnt it required always?

prateek-y (commented):
Looks perfect, we might hit issues while running this with greengrass, as it will do its own init/finalize and looks like there are gotchas while calling some of the session/login API's multiple times with different flags.

include/aws/io/tls_channel_handler.h
graebm: it's NULL if the token uses a "protected authentication path" (token itself has a PIN pad)

I clarified the comment

source/pkcs11.c
graebm: Good idea
I was assuming the PKCS#11 lib would do that, but better to be safe

source/s2n/s2n_tls_channel_handler.c
graebm: this function cannot fail. It's just a atomic increment, that does a "return this" at the end, for the sake of ergonomics

source/s2n/s2n_tls_channel_handler.c
graebm: check out the changes in`source/tls_channel_handler.c`

they're sanitized in the platform-agnostic code that copies the data from `aws_tls_ctx_pkcs11_options` into `aws_tls_ctx_options`

I'll add a comment to that effect

source/pkcs11.c
graebm: added clarifying comments

prateek-y (approved):

include/aws/io/private/pkcs11_private.h
graebm: I flip-flopped a lot on the best way forward here. I just replaced it with `unsigned long` in the "private API".

I tried it every way: exposing the actual PKCS#11 headers, unique typedefs for `aws_pkcs11_session_handle_t`, `aws_pkcs11_slot_id_t`, and just using `unsigned long`.

`unsigned long` felt the least confusing

graebm:
> Looks perfect, we might hit issues while running this with greengrass, as it will do its own init/finalize and looks like there are gotchas while calling some of the session/login API's multiple times with different flags.

Yeah, we'll probably need to add an `omit_login` option, for cases where some other part of the application has already logged in. Same way we have an `omit_initialization` option

source/pkcs11.c
```diff
@@ -398,3 +427,316 @@ void aws_pkcs11_lib_release(struct aws_pkcs11_lib *pkcs11_lib) {
...
+            session_handle);
+        aws_raise_error(AWS_IO_PKCS11_ERROR);
+        goto except;
+    }
+
+    /* Success! */
+    AWS_LOGF_TRACE(AWS_LS_IO_PKCS11, "id=%p session=%lu: Found private key.", (void *)pkcs11_lib, session_handle);
+    *out_key_object_handle = found_objects[0];
+    success = true;
+    goto finally;
```
bretambrose, line 727: can remove/clean

source/s2n/s2n_tls_channel_handler.c
```diff
@@ -924,6 +950,50 @@ static int s2n_monotonic_clock_time_nanoseconds(void *context, uint64_t *time_in
...
+    /* PKCS#11 options were already sanitized (ie: check for required args) in tls_channel_handler.c */
+
+    /* anything initialized in this function is cleaned up during s_s2n_ctx_destroy()
+     * so don't worry about cleaning up unless it's some tmp heap allocation */
+
+    s2n_ctx->pkcs11.lib = aws_pkcs11_lib_acquire(options->pkcs11.lib); /* cannot fail */
+    aws_mutex_init(&s2n_ctx->pkcs11.session_lock);
+
+    unsigned long match_slot_id = options->pkcs11.slot_id;
+    unsigned long slot_id;
```
bretambrose, line 963: Feel like this is going to raise a warning somewhere about uninitialized variables with non-omniscient compilers.

include/aws/io/private/pkcs11_private.h
```diff
@@ -15,5 +17,47 @@
...
+ * Find the slot that meets all criteria:
+ * - has a token
+ * - if match_slot_id is non-null, then slot IDs must match
+ * - if match_token_label is non-null, then labels must match
+ * The function fails unless it finds exactly one slot meeting all criteria.
+ */
+AWS_IO_API
+int aws_pkcs11_lib_find_slot_with_token(
+    struct aws_pkcs11_lib *pkcs11_lib,
+    const unsigned long *match_slot_id,
```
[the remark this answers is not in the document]
graebm: gread point, I'll use uint64_t here
do all the weird casting checks in pkcs11.c

include/aws/io/private/pkcs11_private.h
```diff
@@ -15,5 +17,47 @@
...
+    const struct aws_string *match_token_label,
```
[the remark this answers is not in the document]
graebm: they don't need to be null-terminated

it was just more convenient to pass `aws_string*` than `aws_byte_cursor*`, because everywhere that calls this function is storing an `aws_string*`, and this is a private function

if this was made into a public function, it should use `aws_byte_cursor*` instead.

include/aws/io/private/pkcs11_private.h
```diff
@@ -15,5 +17,47 @@
...
+    unsigned long *out_slot_id);
+
+AWS_IO_API
+int aws_pkcs11_lib_open_session(
+    struct aws_pkcs11_lib *pkcs11_lib,
+    unsigned long slot_id,
+    unsigned long *out_session_handle);
+
+AWS_IO_API
+void aws_pkcs11_lib_close_session(struct aws_pkcs11_lib *pkcs11_lib, unsigned long session_handle);
```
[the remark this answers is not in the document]
graebm: because the calling code is shutting things down, and wouldn't be able to do anything differently if an error was reported.

I'm just not a fan of returning errors if the calling code can't reasonably do anything in response

source/pkcs11.c
```diff
@@ -398,3 +427,316 @@ void aws_pkcs11_lib_release(struct aws_pkcs11_lib *pkcs11_lib) {
...
+        s_raise_ck_error(pkcs11_lib, "C_GetSlotList", rv);
+        goto except;
+    }
+
+    for (size_t i = 0; i < num_slots; ++i) {
+        CK_SLOT_ID slot_id_i = slot_id_array[i];
+
+        /* if specific slot_id requested, and this isn't it, then skip */
+        if ((match_slot_id != NULL) && (*match_slot_id != slot_id_i)) {
+            AWS_LOGF_TRACE(
```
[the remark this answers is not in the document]
graebm: sticking with TRACE. this prints once per slot and this [link](http://www.pkiglobe.org/pkcs11_terminology.html) says:
> In case of HSMs, there could be hundreds or more slots are available> 

there's a DEBUG at the end of the function, that tells which slot we ended up picking

source/pkcs11.c
```diff
@@ -398,3 +427,316 @@ void aws_pkcs11_lib_release(struct aws_pkcs11_lib *pkcs11_lib) {
...
+
+        /* this slot is a candidate! */
+
+        /* be sure there's only one candidate */
+        if (candidate != NULL) {
+            AWS_LOGF_ERROR(
+                AWS_LS_IO_PKCS11,
+                "id=%p: Failed to choose PKCS#11 token, multiple tokens match search criteria",
+                (void *)pkcs11_lib);
+            aws_raise_error(AWS_IO_PKCS11_ERROR);
```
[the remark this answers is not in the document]
graebm: Added AWS_IO_PKCS11_TOKEN_NOT_FOUND

source/pkcs11.c
```diff
@@ -398,3 +427,316 @@ void aws_pkcs11_lib_release(struct aws_pkcs11_lib *pkcs11_lib) {
...
+
+        /* the new candidate! */
+        candidate = &slot_id_array[i];
+        memcpy(&info, &token_info_i, sizeof(CK_TOKEN_INFO));
+    }
+
+    if (candidate == NULL) {
+        AWS_LOGF_ERROR(
+            AWS_LS_IO_PKCS11, "id=%p: Failed to find PKCS#11 token which matches search criteria", (void *)pkcs11_lib);
+        aws_raise_error(AWS_IO_PKCS11_ERROR);
```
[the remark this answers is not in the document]
graebm: using the same AWS_IO_PKCS11_TOKEN_NOT_FOUND if we find 0 or if multiple tokens match

I tried naming it more generically, but it sounded awkward:
AWS_IO_PKCS11_SELECT_TOKEN_FAILED?
AWS_IO_PKCS11_CANNOT_PICK_TOKEN?

source/pkcs11.c
```diff
@@ -398,3 +427,316 @@ void aws_pkcs11_lib_release(struct aws_pkcs11_lib *pkcs11_lib) {
...
+    CK_SLOT_ID *candidate = NULL;
+    CK_TOKEN_INFO info;
+    AWS_ZERO_STRUCT(info);
+    bool success = false;
+
+    /* query number of slots with tokens */
+    CK_ULONG num_slots = 0;
+    CK_RV rv = pkcs11_lib->function_list->C_GetSlotList(CK_TRUE /*tokenPresent*/, NULL /*pSlotList*/, &num_slots);
+    if (rv != CKR_OK) {
+        s_raise_ck_error(pkcs11_lib, "C_GetSlotList", rv);
```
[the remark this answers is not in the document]
graebm: nah. I only see one that shows up multiple times. too much boilerplate

source/pkcs11.c
```diff
@@ -398,3 +427,316 @@ void aws_pkcs11_lib_release(struct aws_pkcs11_lib *pkcs11_lib) {
...
+
+int aws_pkcs11_lib_login_user(
+    struct aws_pkcs11_lib *pkcs11_lib,
+    unsigned long session_handle,
+    const struct aws_string *optional_user_pin) {
+
+    CK_UTF8CHAR_PTR pin = NULL;
+    CK_ULONG pin_len = 0;
+    if (optional_user_pin) {
+        if (optional_user_pin->len > ULONG_MAX) {
```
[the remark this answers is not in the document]
graebm: keeping this. casting without checking first is creepy

source/pkcs11.c
graebm: ok. using our usual goto behaviors now

bretambrose (approved):

PR #434, opened 2021-10-05
Handle connection failure during async PKCS#11 operations

graebm:

### DIFF
--- a/source/s2n/s2n_tls_channel_handler.c
+++ b/source/s2n/s2n_tls_channel_handler.c
@@ -571,7 +571,7 @@ static void s_delayed_shutdown_task_fn(struct aws_channel_task *channel_task, vo
 
 /* This task performs the PKCS#11 private key operations.
  * This task is scheduled because the s2n async private key operation is not allowed to complete synchronously */
-pstatic void s_s2n_pkcs11_async_pkey_task(
+static void s_s2n_pkcs11_async_pkey_task(
     struct aws_channel_task *channel_task,
     void *arg,
     enum aws_task_status status) {
@@ -593,7 +593,8 @@ pstatic void s_s2n_pkcs11_async_pkey_task(
     AWS_LOGF_TRACE(AWS_LS_IO_TLS, "id=%p: Running PKCS#11 async pkey task", (void *)handler);
 
     /* We check all s2n_async_pkey_op functions for success,
-     * but they shouldn't fail if they're called correctly */
+     * but they shouldn't fail if they're called correctly.
+     * Even if the output is bad, the failure will happen later in s2n_negotiate() */
 
     uint32_t input_size = 0;
     if (s2n_async_pkey_op_get_input_size(op, &input_size)) {
--- a/source/s2n/s2n_tls_channel_handler.c
+++ b/source/s2n/s2n_tls_channel_handler.c
@@ -716,7 +716,7 @@ static int s_s2n_pkcs11_async_pkey_callback(struct s2n_connection *conn, struct 
     (void)conn;
 
     /* Schedule a task to do further work.
-     * We can't do it now because the s2n async private key operation cannot complete synchronously */
+     * We can't just do the work now because the s2n async private key operation cannot complete synchronously */
     AWS_LOGF_TRACE(AWS_LS_IO_TLS, "id=%p: async pkey callback received, scheduling PKCS#11 task", (void *)handler);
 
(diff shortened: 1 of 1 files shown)


justinboswell (approved):

source/s2n/s2n_tls_channel_handler.c
```diff
@@ -319,6 +320,11 @@ static void s_on_negotiation_result(
 static int s_drive_negotiation(struct aws_channel_handler *handler) {
     struct s2n_handler *s2n_handler = (struct s2n_handler *)handler->impl;
 
+    if (s2n_handler->negotiation_failed || s2n_handler->negotiation_finished) {
+        AWS_LOGF_TRACE(AWS_LS_IO_TLS, "id=%p: Connection is not in a state to continue negotiation", (void *)handler);
```
JonathanHenson, line 324: This will not happen often and if it does the log message isn't going to be your bottleneck. I wouldn't mind an assert here, but at a minimum i think this is ERROR or WARN

JonathanHenson: also AWS_UNLIKELY would probably apply here as well?

source/s2n/s2n_tls_channel_handler.c
```diff
@@ -577,29 +585,36 @@ static void s_s2n_pkcs11_async_pkey_task(
...
+        AWS_LOGF_ERROR(AWS_LS_IO_TLS, "id=%p: Failed querying s2n async pkey op size", (void *)handler);
+        aws_raise_error(AWS_IO_TLS_ERROR_NEGOTIATION_FAILURE);
         goto error;
     }
 
     input_data = aws_mem_acquire(handler->alloc, input_size);
     if (s2n_async_pkey_op_get_input(op, input_data, input_size)) {
-        aws_raise_error(AWS_ERROR_INVALID_STATE);
+        AWS_LOGF_ERROR(AWS_LS_IO_TLS, "id=%p: Failed querying s2n async pkey input", (void *)handler);
+        aws_raise_error(AWS_IO_TLS_ERROR_NEGOTIATION_FAILURE);
```
JonathanHenson, line 609: this seems more like an io error or a syscall error. would that be more instructive? to the user?

JonathanHenson (approved):
FIX or disagree-n-ship

source/s2n/s2n_tls_channel_handler.c
```diff
@@ -60,7 +60,8 @@ struct s2n_handler {
     aws_tls_on_error_fn *on_error;
     void *user_data;
     bool advertise_alpn_message;
-    bool negotiation_finished;
+    bool negotiation_finished; /* true after negotiation succeeds */
```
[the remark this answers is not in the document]
graebm: yeah, I had the same thoughts. I'll go with negotiation_succeeded.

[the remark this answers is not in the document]
graebm: actually, went with NEGOTIATION_ONGOING/SUCCEEDED/FAILED enum

I debated adding NOT_STARTED and SHUTDOWN_COMPLETE states, but that made me feel like I should add a lot more state checking in a lot more places and I'd like to keep these changes minimal, if possible

source/s2n/s2n_tls_channel_handler.c
```diff
@@ -674,21 +704,19 @@ static void s_s2n_pkcs11_async_pkey_task(
...
 
 static int s_s2n_pkcs11_async_pkey_callback(struct s2n_connection *conn, struct s2n_async_pkey_op *op) {
-    /* TODO:
-     * - should we just store the op, and invoke PKCS11 stuff from negotiate()?
-         removes possibility of channel dying before task runs
-     */
     struct s2n_handler *s2n_handler = s2n_connection_get_ctx(conn);
     struct aws_channel_handler *handler = &s2n_handler->handler;
 
     AWS_ASSERT(conn == s2n_handler->connection);
```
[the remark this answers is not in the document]
graebm: channel tasks always run before the channel is destroyed. That's why we made "channel tasks" a thing distinct from "vanilla event loop tasks". I'll add a comment

source/s2n/s2n_tls_channel_handler.c
graebm: I'll go back to INVALID_STATE.
Yeah, I wasn't sure what to do here. But talking with some s2n engineers, these should never fail unless we're doing something really terribly wrong

PR #444, opened 2021-10-13
Input Stream API change

TingDaoK:
API change:
- Refcount `aws_input_stream`
- API change, every down stream makes its own input stream implementation needs to be updated.
- Introduced `release/acquire` vtable function for language bindings that need to use their own refcounts.

### will not merge until we fixed the downstream that creates its own inputstream

C:
- https://github.com/awslabs/aws-c-s3/pull/175
- https://github.com/awslabs/aws-c-http/pull/342
- https://github.com/awslabs/aws-c-auth/pull/159

CRTs:
- https://github.com/awslabs/aws-crt-python/pull/336
- https://github.com/awslabs/aws-crt-nodejs/pull/301
- https://github.com/awslabs/aws-crt-java/pull/447
- https://github.com/awslabs/aws-crt-cpp/pull/341
- https://github.com/awslabs/aws-crt-dotnet/pull/81
- https://github.com/awslabs/aws-crt-ffi/pull/40

**For aws-crt-*, we will need to update them when they pull the latest submodules**


include/aws/io/stream.h
```diff
@@ -38,17 +39,29 @@ struct aws_input_stream_vtable {
...
-    void (*destroy)(struct aws_input_stream *stream);
+    void (*impl_destroy)(struct aws_input_stream *stream);
 };
 
-struct aws_input_stream {
+struct aws_input_stream_options {
     struct aws_allocator *allocator;
-    void *impl;
     struct aws_input_stream_vtable *vtable;
+    void *impl;
```
graebm, line 48: debatable/trivial: instead of passing in `void *impl`, what if we passed in `size_t sizeof_impl` and then new() did `aws_mem_acquire_many()` to create both the base-struct and impl in one go?

source/stream.c
```diff
@@ -10,6 +10,13 @@
 
 #include <errno.h>
 
+struct aws_input_stream {
+    struct aws_allocator *allocator;
+    void *impl;
+    struct aws_input_stream_vtable *vtable;
```
graebm, line 16: trivial: const `const struct aws_input_stream_vtable *vtable`

source/stream.c
```diff
@@ -59,19 +66,43 @@ int aws_input_stream_get_length(struct aws_input_stream *stream, int64_t *out_le
...
+}
+
+struct aws_input_stream *aws_input_stream_new(const struct aws_input_stream_options *options) {
+    AWS_PRECONDITION(options);
+    AWS_PRECONDITION(options->allocator);
 
-        stream->vtable->destroy(stream);
+    struct aws_input_stream *input_stream = aws_mem_calloc(options->allocator, 1, sizeof(struct aws_input_stream));
+    if (!input_stream) {
+        return NULL;
```
graebm, line 88: allocations cannot fail. remove this check

source/stream.c
```diff
@@ -188,52 +219,46 @@ static int s_aws_input_stream_byte_cursor_get_length(struct aws_input_stream *st
...
-    aws_mem_acquire_many(
-        allocator,
-        2,
-        &input_stream,
-        sizeof(struct aws_input_stream),
-        &impl,
-        sizeof(struct aws_input_stream_byte_cursor_impl));
-
-    if (!input_stream) {
+    if (!impl) {
```
graebm, line 241: remove check for allocation failure

source/stream.c
```diff
@@ -278,84 +303,92 @@ static int s_aws_input_stream_file_get_status(struct aws_input_stream *stream, s
...
 
-    struct aws_input_stream *input_stream = NULL;
-    struct aws_input_stream_file_impl *impl = NULL;
-
-    aws_mem_acquire_many(
-        allocator, 2, &input_stream, sizeof(struct aws_input_stream), &impl, sizeof(struct aws_input_stream_file_impl));
+    struct aws_input_stream_file_impl *impl = aws_mem_calloc(allocator, 1, sizeof(struct aws_input_stream_file_impl));
 
-    if (!input_stream) {
+    if (!impl) {
```
graebm, line 332: remove OOM check

source/stream.c
```diff
@@ -278,84 +303,92 @@ static int s_aws_input_stream_file_get_status(struct aws_input_stream *stream, s
...
 struct aws_input_stream *aws_input_stream_new_from_open_file(struct aws_allocator *allocator, FILE *file) {
-    struct aws_input_stream *input_stream = NULL;
-    struct aws_input_stream_file_impl *impl = NULL;
+    struct aws_input_stream_file_impl *impl = aws_mem_calloc(allocator, 1, sizeof(struct aws_input_stream_file_impl));
 
-    aws_mem_acquire_many(
-        allocator, 2, &input_stream, sizeof(struct aws_input_stream), &impl, sizeof(struct aws_input_stream_file_impl));
-
-    if (!input_stream) {
+    if (!impl) {
```
graebm, line 359: remove OOM check

source/stream.c
```diff
@@ -61,17 +68,41 @@ int aws_input_stream_get_length(struct aws_input_stream *stream, int64_t *out_le
 
 void aws_input_stream_destroy(struct aws_input_stream *stream) {
     if (stream != NULL) {
-        AWS_ASSERT(stream->vtable && stream->vtable->destroy);
+        AWS_ASSERT(stream->vtable && stream->vtable->impl_destroy);
 
-        stream->vtable->destroy(stream);
+        stream->vtable->impl_destroy(stream);
+        aws_mem_release(stream->allocator, stream);
```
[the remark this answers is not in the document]
graebm: oops. this comment is from weeks ago...

include/aws/io/stream.h
TingDaoK: hmmm, I feel it is a bit weird as the `impl` will still need a callback to destroy (To clean up resource besides of the memory allocation). So, the destroy callback will not need to clean up the memory seems like a bit weird to me.

TwistedTwigleg (commented):
Looks good to me!
There was only one minor thing I noticed, and it's more of a question than anything.

Because I'm not experienced with this code yet, I would wait for another review just to make sure it is all good, as it's entirely possible I missed something.

source/stream.c
```diff
@@ -59,19 +66,37 @@ int aws_input_stream_get_length(struct aws_input_stream *stream, int64_t *out_le
     return stream->vtable->get_length(stream, out_length);
 }
 
-void aws_input_stream_destroy(struct aws_input_stream *stream) {
+void s_aws_input_stream_destroy(struct aws_input_stream *stream) {
```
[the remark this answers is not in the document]
TingDaoK: yeah, good catch. It should be static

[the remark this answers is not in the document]
TingDaoK: Well, what other function starts with `s_` but not being static? I didn't notice any from a quick search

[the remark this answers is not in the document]
TwistedTwigleg: I just looked and now I cannot find it. I thought it was in “stream.c” or maybe the transport handler, but I looked again just now in all the changed files and didn’t see any.

I’ll take another look in the morning though to be sure, but maybe I just thought I saw multiple 🤔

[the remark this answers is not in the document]
TwistedTwigleg: I looked again and realized what it was: I was the static `struct` creations that were throwing me off, as thought the class inheritance was the name. For example, I thought `aws_channel_handler_vtable` was the name of the struct in `static struct aws_channel_handler_vtable s_handler_vtable`, not `s_handler_vtable`.

Apologizes for the confusion! I looked through all of the files and can confirm all the statics now start with `s_`.

bretambrose (commented):
This pattern is a bit different than what we've done in the past and I'm unsure that there's a need for that difference.

What I'd prefer to see:

aws_input_stream stays public (remove allocator, add ref count)
aws_input_stream is embedded as a value within the actual implementations, can remove the get_impl function as well
destroy is removed from the vtable because the refcount-to-zero callback -- set at subclass creation time -- will do the full cleanup.

source/stream.c
```diff
@@ -59,19 +59,13 @@ int aws_input_stream_get_length(struct aws_input_stream *stream, int64_t *out_le
...
-    }
-}
-
 /*
  * cursor stream implementation
  */
 
 struct aws_input_stream_byte_cursor_impl {
+    struct aws_allocator *allocator;
+    struct aws_input_stream base;
```
graebm, line 68: trivial: usually we make `base` is the very first member

graebm (approved):

include/aws/io/stream.h
```diff
@@ -39,6 +39,17 @@ struct aws_input_stream_vtable {
...
+    /**
+     * Optional. If not set, the default aws_ref_count_acquire/release will be used.
+     * Set for high level language binding that has its own refcounting implementation and needs to be kept alive from
+     * C.
+     * When the C level needs to keep the high level language binding alive, while binding as the user also holds the
+     * reference to the C object, it will result in a recursive reference.
+     * The function here is for C Object to use the high level refcount implementation directly.
+     */
+    struct aws_input_stream *(*acquire)(struct aws_input_stream *stream);
+    struct aws_input_stream *(*release)(struct aws_input_stream *stream);
```
graebm, line 52: this doesn't need to return anything
you can do that from aws_input_stream_acquire()

include/aws/io/stream.h
```diff
@@ -38,17 +39,47 @@ struct aws_input_stream_vtable {
...
     int (*get_length)(struct aws_input_stream *stream, int64_t *out_length);
-    void (*destroy)(struct aws_input_stream *stream);
+
+    /**
+     * Optional. If not set, the default aws_ref_count_acquire/release will be used.
+     * Set for high level language binding that has its own refcounting implementation and needs to be kept alive from
+     * C.
+     * When the C level needs to keep the high level language binding alive, while binding as the user also holds the
+     * reference to the C object, it will result in a recursive reference.
+     * The function here is for C Object to use the high level refcount implementation directly.
```
graebm, line 49: This sentence doesn't make a ton of sense, just remove it. previous sentence says it nicely

graebm: Add comment about how the refcount isn't used if you override acquire()/release()

graebm (approved):
fix & ship

include/aws/io/stream.h
bretambrose: This is something I grumped about earlier and twisted Dengke's arm to include.  I prefer the return value of always-null to simplify non-final cleanup (just assign the result) and swaps

TingDaoK: yeah, we agreed that the c function should return the pointer, but the implementation doesn't need to.

Issue #401, opened 2021-06-21
aws-c-io includes Windows.h in public headers


pitrou:
The `Windows.h` header has started being included in aws-c-io public headers, notably `aws/io/event_loop.h`. This in turns gets exposed into aws-sdk-cpp public headers, and can have repercussions on downstream users (especially due to Windows' habit of defining many macros with common names).

bretambrose:
Fixed in https://github.com/awslabs/aws-c-io/pull/454

pitrou:
Thank you!

PR #453, opened 2021-11-11
Remove windows.h inclusion from public headers

AMZN-hgoffin:
As of AWS C++ SDK 1.9, windows.h is being pulled in via aws/io/event_loop.h, which is causing lots of pre-existing code to suddenly pull in a bunch of ill-behaved Windows SDK macros such as `#define GetMessage GetMessageA`.

This pull request hides windows.h as an internal-only dependency, duplicating the Windows `struct _OVERLAPPED` layout into `struct aws_win32_OVERLAPPED` and adding a centralized type-cast function to avoid the proliferation of unsafe casts throughout the codebase.

This resolves https://github.com/awslabs/aws-c-io/issues/401

### DIFF
--- a/include/aws/io/event_loop.h
+++ b/include/aws/io/event_loop.h
@@ -25,7 +25,6 @@ struct aws_task;
 struct aws_thread_options;
 
 #if AWS_USE_IO_COMPLETION_PORTS
-#    include <Windows.h>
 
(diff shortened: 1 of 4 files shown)


bretambrose:
Merged in https://github.com/awslabs/aws-c-io/pull/454

PR #459, opened 2021-12-20
Windows cert chain now acts like rest of crt

sbSteveK:
Changed windows_pki_utils.c to process certificate chains in the same order as all other platforms.


source/windows/windows_pki_utils.c
```diff
@@ -603,7 +603,7 @@ int aws_import_key_pair_to_cert_context(
             aws_raise_error(AWS_ERROR_SYS_CALL_FAILURE);
         }
 
-        if (i != cert_count - 1 || !add_result) {
+        if (i != 0 || !add_result) {
```
graebm, line 606: Let's build a new unit test that stresses using a certificate-chain

JonathanHenson: we could probably just use a cert chain for the testing-resources unit test certs.

JonathanHenson (approved):
Could you generate a new certificate chain for unit tests? To make the change work seamlessly you can simply replace the files that are there with some that use a chain.

PR #460, opened 2021-12-21
Remove platform ifdefs from tls_channel_handler.h

graebm:
**Issue**
It was annoying and error prone having to mirror these ifdefs around every function that could call these functions, and every function that could call those functions, etc, etc, etc. There were multiple compilation errors in aws-crt-cpp on iOS because we didn't exactly match the nested pattern of ifdefs previously seen here.

**Description of changes:**
Instead of ifdef-ing out platform-specific functions in the H file, just ifdef in the C file, and raise a runtime error on unsupported platforms.

PS: Merry Christmas @bretambrose 

### DIFF
--- a/include/aws/io/tls_channel_handler.h
+++ b/include/aws/io/tls_channel_handler.h
@@ -411,10 +411,12 @@ AWS_IO_API int aws_tls_ctx_options_init_default_server(
     struct aws_byte_cursor *pkey);
 
 /**
- * Initializes options for use with mutual tls in client mode. This function only works on
- * Windows. cert_reg_path is the path to a system
+ * Initializes options for use with mutual tls in client mode.
+ * cert_reg_path is the path to a system
  * installed certficate/private key pair. Example:
  * CurrentUser\\MY\\<thumprint>
+ *
+ * NOTE: This only works on Windows.
  */
 AWS_IO_API int aws_tls_ctx_options_init_client_mtls_from_system_path(
     struct aws_tls_ctx_options *options,
@@ -422,10 +424,12 @@ AWS_IO_API int aws_tls_ctx_options_init_client_mtls_from_system_path(
     const char *cert_reg_path);
 
 /**
- * Initializes options for use with server mode. This function only works on
- * Windows. cert_reg_path is the path to a system
+ * Initializes options for use with server mode.
+ * cert_reg_path is the path to a system
  * installed certficate/private key pair. Example:
  * CurrentUser\\MY\\<thumprint>
+ *
+ * NOTE: This only works on Windows.
  */
 AWS_IO_API int aws_tls_ctx_options_init_default_server_from_system_path(
     struct aws_tls_ctx_options *options,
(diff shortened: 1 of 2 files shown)


bretambrose:
Is there a #pragma (or cross platform macro we could make that uses them) we could embed in the unimplemented bodies to also make some compile-time error/warning output?

graebm:
> Is there a #pragma (or cross platform macro we could make that uses them) we could embed in the unimplemented bodies to also make some compile-time error/warning output?

We could conditionally mark the function deprecated, so the compiler warns when it's used on an unsupported platform. But I think that kind of markup needs to go on the function *declaration* in the H file, I don't think you can do it on the definition in the C file. So then any "wrapper function" would have to disable the warning. So like: in aws-crt-cpp `TlsContextOptions` has functions that mirror these, so THOSE function declarations would need the same conditional deprecated warnings, but then in the .cpp file we'd need to disable deprecated warnings for the calls in the the C layer. And then the MQTT builder has functions that call into THOSE functions. So we'd need to do the same thing: put the conditional deprecated tags on the wrapper functions, but then disable deprecated warnings in the cpp file.

This is starting to feel complex and fragile, similarly to what we already have here.

Also, this only works in C++. The other languages will still have the same runtime-errors-only

-------------
Alternatively I could put an AWS_ASSERT() in the unimplemented parts of the C file. So you'd get a more severe shock if you compile in debug

bretambrose:
> > Is there a #pragma (or cross platform macro we could make that uses them) we could embed in the unimplemented bodies to also make some compile-time error/warning output?
> 
> We could conditionally mark the function deprecated, so the compiler warns when it's used on an unsupported platform. But I think that kind of markup needs to go on the function _declaration_ in the H file, I don't think you can do it on the definition in the C file. So then any "wrapper function" would have to disable the warning. So like: in aws-crt-cpp `TlsContextOptions` has functions that mirror these, so THOSE function declarations would need the same conditional deprecated warnings, but then in the .cpp file we'd need to disable deprecated warnings for the calls in the the C layer. And then the MQTT builder has functions that call into THOSE functions. So we'd need to do the same thing: put the conditional deprecated tags on the wrapper functions, but then disable deprecated warnings in the cpp file.
> 
> This is starting to feel complex and fragile, similarly to what we already have here.
> 
> Also, this only works in C++. The other languages will still have the same runtime-errors-only
> 
> Alternatively I could put an AWS_ASSERT() in the unimplemented parts of the C file. So you'd get a more severe shock if you compile in debug

Yeah, never mind on that question.

bretambrose (approved):

graebm:
Opinion: is `AWS_ERROR_UNIMPLEMENTED` too vague for someone that doesn't have logging turned on? Should we have a more specific error type named something like `AWS_ERROR_UNAVAILABLE_ON_PLATFORM`?

bretambrose:
> Opinion: is `AWS_ERROR_UNIMPLEMENTED` too vague for someone that doesn't have logging turned on? Should we have a more specific error type named something like `AWS_ERROR_UNAVAILABLE_ON_PLATFORM`? `AWS_ERROR_PLATFORM_NOT_SUPPORTED`?

I like platform not supported, makes the real problem clearer.

graebm:
> > Opinion: is `AWS_ERROR_UNIMPLEMENTED` too vague for someone that doesn't have logging turned on? Should we have a more specific error type named something like `AWS_ERROR_UNAVAILABLE_ON_PLATFORM`? `AWS_ERROR_PLATFORM_NOT_SUPPORTED`?
> 
> I like platform not supported, makes the real problem clearer.

https://github.com/awslabs/aws-c-common/pull/882

PR #463, opened 2022-01-03
Fix some sporadically failing tests

bretambrose:
Pinned event loop test was seeing sporadic crashes in the assert because that test only uses/needs a single handler.  Added a dummy handler to address.

Also fixed race condition in shutdown of ca_override test.

### DIFF
--- a/source/channel.c
+++ b/source/channel.c
@@ -978,7 +978,7 @@ int aws_channel_slot_on_handler_shutdown_complete(
 }
 
 size_t aws_channel_slot_downstream_read_window(struct aws_channel_slot *slot) {
-    AWS_ASSERT(slot->adj_right);
+    AWS_ASSERT(slot->adj_right || !slot->channel->read_back_pressure_enabled);
     return slot->channel->read_back_pressure_enabled ? slot->adj_right->window_size : SIZE_MAX;
 }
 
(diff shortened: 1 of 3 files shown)


source/channel.c
```diff
@@ -978,7 +978,7 @@ int aws_channel_slot_on_handler_shutdown_complete(
 }
 
 size_t aws_channel_slot_downstream_read_window(struct aws_channel_slot *slot) {
-    AWS_ASSERT(slot->adj_right);
+    AWS_ASSERT(slot->adj_right || !slot->channel->read_back_pressure_enabled);
     return slot->channel->read_back_pressure_enabled ? slot->adj_right->window_size : SIZE_MAX;
```
graebm, line 982: This doesn't seem safe.

It's checking whether there's a slot to the right, because the channel needs to be sure that something will be there to receive the data that gets read.

graebm: well, this specific change seems OK in isolation

but `socket_channel_handler.c:s_do_read()` uses this to see how much data to read from the socket, and that function is assuming that a handler will be there to receive the data it reads.

If you need to make this change, you might need to revisit code that calls it, and see if any of that needs changing

I'm not sure the exact callstack you were failing with...

graebm: maybe s_do_read() should just early-out if there's no handler to the right?

bretambrose: Well, it reads any data and tries to pass it in the read direction.  That's going to fail and so it frees the io message and merrily continues on.

graebm (approved):

PR #465, opened 2022-01-11
Change windows cert chain processing to act like other platforms

sbSteveK:
Change windows certificate chain processing to work the same way as all other platforms.


graebm (commented):
I'm cool with this change, but uncomfortable releasing it until we have a test confirming that it works like we expect

sbSteveK:
Sounds good to me. Will add the unit of certificate chain TLS as a part of the ticket =D

PR #467, opened 2022-01-17
Add option to bind to an interface

cobookman:
*Issue #, if available:* N/A

*Description of changes:* Adds optional parameter which specifies a network interface to bind to. Forcing network traffic to use just that single interface.


bretambrose:
This is a good feature request, but we cannot put a byte cursor in this structure as it gets copied by value across threads.  A likely solution would be to make it a fixed-size array.

waahm7:
Superseded by #647

PR #468, opened 2022-01-26
Make get length const

bretambrose:
for *reasons*


include/aws/io/stream.h
```diff
@@ -37,7 +37,7 @@ struct aws_input_stream_vtable {
      */
     int (*read)(struct aws_input_stream *stream, struct aws_byte_buf *dest);
     int (*get_status)(struct aws_input_stream *stream, struct aws_stream_status *status);
-    int (*get_length)(struct aws_input_stream *stream, int64_t *out_length);
+    int (*get_length)(const struct aws_input_stream *stream, int64_t *out_length);
```
graebm, line 40: I mean, if for some implementation this means going to the end of the stream, and then rewinding, it's not const then

bretambrose: On second thought, it was a silly idea.

PR #469, opened 2022-01-26
Fix Certificate Chains for Windows

sbSteveK:
Fixed certificate chains for Windows.

Changed TLS tests to use a ca_root that generates an end user certificate for the server.

Added a test for certificate chains.

Added a script that generates new certificates for tests.


tests/resources/generateCerts.sh
```diff
@@ -0,0 +1,67 @@
...
+            -set_serial 02 \
+            -subj '/C=US/ST=Washington/L=Seattle/O=Amazon/OU=SDKs/CN=localhost/emailAddress=aws-sdk-common-runtime@amazon.com'
+
+yes | openssl ca -config ca_root.cnf \
+            -extensions server_cert \
+            -days 824 -notext -md sha256 \
+            -in server.csr \
+            -out server.crt
+
+cat server.crt ca_root.crt > certChain.crt
```
graebm, line 37: trivial: name suggestion ~certChain.crt~ -> `server_chain.crt`
be clear it's related to the server.crt, and snake_case like the other files :|

tests/resources/generateCerts.sh
```diff
@@ -0,0 +1,67 @@
...
+mkdir certGeneration
+cp unittests.key certGeneration/unittests.key
+cp unittests.conf certGeneration/unittests.conf
+cp ca_root.cnf certGeneration/ca_root.cnf
+cd certGeneration
+
+touch index.txt
+echo 1000 > serial
+
+openssl genrsa -out ca_root.key 2048
```
graebm, line 14: trivial/optional: we don't *need* to regenerate the keys, just the certs, but you could leave this in as a comment as to how it was made

source/windows/windows_pki_utils.c
```diff
@@ -603,7 +603,7 @@ int aws_import_key_pair_to_cert_context(
             aws_raise_error(AWS_ERROR_SYS_CALL_FAILURE);
         }
 
-        if (i != cert_count - 1 || !add_result) {
+        if (i != 0 || !add_result) {
```
graebm, line 606: trivial: rename this PR something like "Fix certificate chains on windows"
because that's the real important change from the perspective of someone seeing this merge in the commit history.
The fact that you shuffled around some files for testing isn't as important from an outsider perspective, even if that's most of the changes in this PR

graebm (approved):

PR #476, opened 2022-03-09
Deprecate macOS keychain path function

graebm:
### Issue

These were causing compiler warnings on macOS 12
```
source/darwin/darwin_pki_utils.c:52:36: warning: 'SecKeychainOpen' is deprecated: first deprecated in macOS 12.0 - Custom keychain management is no longer supported [-Werror,-Wdeprecated-declarations]
```
### Research

These deprecated functions only happen if someone calls `aws_tls_ctx_options_set_keychain_path()`. This function doesn't seem widely used. It's only bound in aws-crt-cpp, no other aws-crt-xyz libraries expose it.

### Description of changes:
- Mark `aws_tls_ctx_options_set_keychain_path()` deprecated
- Disable the compiler warnings so we can still compile with warnings-as-errors

### DIFF
--- a/source/darwin/darwin_pki_utils.c
+++ b/source/darwin/darwin_pki_utils.c
@@ -18,6 +18,12 @@ static struct aws_mutex s_sec_mutex = AWS_MUTEX_INIT;
 
 #if !defined(AWS_OS_IOS)
 
+#    pragma clang diagnostic push
+/* macOS 12.0 starting marking SecKeychainOpen() and SecKeychainUnlock() as deprecated
+ * because "Custom keychain management is no longer supported".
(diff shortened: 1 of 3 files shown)


source/darwin/darwin_pki_utils.c
```diff
@@ -18,6 +18,12 @@ static struct aws_mutex s_sec_mutex = AWS_MUTEX_INIT;
 
 #if !defined(AWS_OS_IOS)
 
+#    pragma clang diagnostic push
```
JonathanHenson, line 21: should we guard this with version macros?

JonathanHenson (approved):

PR #479, opened 2022-03-15
Added ECC support for PKCS11

JamieHunter:
PKCS11 feature was missing ECC support. This commit adds ECC support.

*Issue #, if available:*

*Description of changes:*
Added PKCS11 ECC capabilities to S2N callback
Added PKCS11 ECC testing certificates and test scenarios


bretambrose (commented):
Without understanding the details, would the existence of https://github.com/awslabs/aws-c-cal/blob/main/include/aws/cal/private/der.h (currently private) let you remove some of the DER helper functions if it was made public?

JamieHunter:
> Without understanding the details, would the existence of https://github.com/awslabs/aws-c-cal/blob/main/include/aws/cal/private/der.h (currently private) let you remove some of the DER helper functions if it was made public?

Yes, most likely they would.

source/pkcs11.c
```diff
@@ -1177,6 +1187,136 @@ static int s_pkcs11_sign_rsa(
     return success ? AWS_OP_SUCCESS : AWS_OP_ERR;
 }
 
+/*
+ * Basic ASN.1 (DER) encoding of header -- sufficient for ECDSA
+ */
+static int s_asn1_enc_prefix(struct aws_byte_buf *buffer, uint8_t identifier, size_t length) {
+    if (((identifier & 0x1f) == 0x1f) || (length > 0x7f)) {
+        AWS_LOGF_ERROR(AWS_LS_IO_PKCS11, "Unable to encode ASN.1 (DER) header 0x%02x %lu", identifier, (long)length);
```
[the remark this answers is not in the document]
JamieHunter: I was wondering about that - I didn't see %zu used elsewhere (probably didn't look hard enough), and wasn't sure if it would break in older versions of C compiler.

source/pkcs11.c
```diff
@@ -1177,6 +1187,136 @@ static int s_pkcs11_sign_rsa(
...
+        bigint--;
+    }
+    // header - indicate integer
+    bool success = s_asn1_enc_prefix(buffer, 0x02, trim_len) == AWS_OP_SUCCESS;
+    if (trim_len > length) {
+        // add leading zero that was not in original buffer
+        trim_len--;
+        bigint++;
+        uint8_t zero = 0;
+        success = success && aws_byte_buf_write(buffer, &zero, 1);
```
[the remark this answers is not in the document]
JamieHunter: Thanks, I was looking for a write_byte, and missed write_u8.

sckulkarni246:
Hello AWS team,

Do you know when this feature would be merged into main?
Also, will this feature be a part of the next device SDK release?

Regards,
Shashank

include/aws/io/io.h
```diff
@@ -240,6 +240,8 @@ enum aws_io_errors {
 
     AWS_ERROR_IO_PINNED_EVENT_LOOP_MISMATCH,
 
+    AWS_ERROR_PKCS11_ENCODING_ERROR,
```
graebm, line 243: You should also add an entry in [io.c](https://github.com/awslabs/aws-c-io/blob/be2952e5ea2048cba35cd07e521ceb78f7587114/source/io.c#L280) with plain text for the error

source/pkcs11.c
```diff
@@ -1177,6 +1187,137 @@ static int s_pkcs11_sign_rsa(
...
+
+    if (s_pkcs11_sign_helper(
+            pkcs11_lib, session_handle, key_handle, mechanism, digest_data, allocator, &part_signature) !=
+        AWS_OP_SUCCESS) {
+        goto error;
+    }
+
+    size_t num_bytes = part_signature.len / 2;
+    aws_byte_buf_init(&r_part, allocator, num_bytes + 4);
+    aws_byte_buf_init(&s_part, allocator, num_bytes + 4);
```
graebm, line 1284: Are "r" and "s" always exactly the same length? is splitting exactly in the middle always the right thing to do?

is it worth checking whether `part_signature.len` is an odd number and failing explicitly here? (I could also see the argument that it doesn't matter because the handshake will fail eventually anyway)

graebm (commented):
Looks pretty good. I'm guessing the failing tests are due to changing the test certs

source/pkcs11.c
JamieHunter: I will add some comments to this effect (yes, they are always same length)

graebm (approved):
🥳

PR #487, opened 2022-05-19
Add ability to read bound address

bgklika:
*Issue #, if available:*
https://github.com/awslabs/aws-c-io/issues/488

When start IPC server on automatically configured port (0).
After socket is bound that change add ability to read real address and port number which socket bound to.

*Description of changes:*
added functions 
`aws_socket_get_bound_address()`
`aws_server_bootstrap_get_bound_port_of_socket_listener()`
added test
`s_test_bind_on_0_port_ipv4_ok`


include/aws/io/channel_bootstrap.h
```diff
@@ -294,6 +294,14 @@ AWS_IO_API int aws_server_bootstrap_set_alpn_callback(
 AWS_IO_API struct aws_socket *aws_server_bootstrap_new_socket_listener(
     const struct aws_server_socket_channel_bootstrap_options *bootstrap_options);
 
+/**
+ * Get local bound port for network sockets.
+ * @retval -1 on errors
+ */
+AWS_IO_API int aws_server_bootstrap_get_bound_port_of_socket_listener(
```
bretambrose, line 301: This function seems redundant given the other one.

include/aws/io/socket.h
```diff
@@ -176,11 +176,17 @@ AWS_IO_API int aws_socket_connect(
  */
 AWS_IO_API int aws_socket_bind(struct aws_socket *socket, const struct aws_socket_endpoint *local_endpoint);
 
+/**
+ * TCP/UDP only. Get bound port number.
+ */
+AWS_IO_API int aws_socket_get_bound_address(struct aws_socket *socket, struct aws_socket_endpoint *local_address);
```
bretambrose, line 182: I think the socket can be const.

Comment seems inaccurate given that it's getting more than the port.

source/posix/socket.c
```diff
@@ -840,6 +840,101 @@ int aws_socket_bind(struct aws_socket *socket, const struct aws_socket_endpoint
...
+                    AF_INET, &s->sin_addr, local_address->address, sizeof(local_address->address))) {
+                AWS_LOGF_DEBUG(
+                    AWS_LS_IO_SOCKET,
+                    "id=%p fd=%d: local endpoint %s:%d",
+                    (void *)socket,
+                    socket->io_handle.data.fd,
+                    local_address->address,
+                    port);
+            } else {
+                AWS_LOGF_WARN(
```
bretambrose, line 884: This seems like a failure path but isn't treated as one.

bretambrose (commented):
I started to go through the PR but the more I thought about it the more I have concerns about the overall approach.  Given the way it gets used downstream in the other PRs I feel like there are memory model safety/visibility issues as well as potential race conditions on shutdown.  Also, these queries can side-affect the state of the socket cross-thread which is also worrying.

Based on the downstream usage it's used only in the server listener constructor.  I wish there was a way we could do this that didn't look like "call this at anytime" because there are definitely times where it would be a disaster.

include/aws/io/channel_bootstrap.h
bgklika: It depends on strategy.

For strict consistency, the best case is reading the actual port directly in aws_socket_bind() and updating the field of struct aws_socket_endpoint *local_endpoint. Unfortunately local_endpoint argument is defined as a pointer to const struct.
Removing a const specifier means drop backward compatibility. Also changes in aws_socket_bind() can affect any existing applications which do not expect the port in local_endpoint will be changed by aws_socket_bind(). Additional third output argument means changing function signature and updating all aws_socket_bind() callers, in all existing applications.

For strict backwards compatibility best is to implement an isolated chain of functions to provide new functions and keep unmodified all existing applications and library functions used by it. Only new applications which use new functions can cause problems even when bugs exist in new code.

Because I'm newbie in aws-crt-java and it's 18 submodules, I selected the second approach.

bgklika:
It depends on strategy.

For strict consistency, the best case is reading the actual port directly in aws_socket_bind() and updating the field of struct aws_socket_endpoint *local_endpoint. Unfortunately local_endpoint argument is defined as a pointer to const struct.
Removing a const specifier means drop backward compatibility. Also changes in aws_socket_bind() can affect any existing applications which do not expect the port in local_endpoint will be changed by aws_socket_bind(). Additional third output argument means changing function signature and updating all aws_socket_bind() callers, in all existing applications.

For strict backwards compatibility best is to implement an isolated chain of functions to provide new functions and keep unmodified all existing applications and library functions used by it. Only new applications which use new functions can cause problems even when bugs exist in new code.

Because I'm newbie in aws-crt-java and it's 18 submodules, I selected the second approach.

include/aws/io/channel_bootstrap.h
bgklika: It is wrapper function for aws_event_stream_rpc_server_listener_get_bound_port of aws-c-event-stream. Please take a look https://github.com/awslabs/aws-c-event-stream/pull/78 

Unfortunately  event_stream_rpc_server.c does not use aws_socket_ symbols directly.
I just follow existing path of calls.

include/aws/io/socket.h
bgklika: I thinking about const but unfortunalety windows implementation has a `socket->state = ERRORED;`
I copied that behaviour from aws_socket_bind() implementation.

bgklika: Note updated in dfab913a3c6fead5d66c9f498fcbac74d24e4c34

source/posix/socket.c
bgklika: Resolved in b13f39ac0d52312cd319668db11bb4a4b167098c

graebm:
Follow-up pull request is here: https://github.com/awslabs/aws-c-io/pull/491

PR #492, opened 2022-05-31
Add additional logging around cross-thread scheduling

TwistedTwigleg:
*Description of changes:*

Adds additional logging around scheduling an event immediately for cross-thread scheduling. This will allow us to better see when cross-thread scheduling is occurring and, if it is occurring, why it might not be immediately executing. This does not change any functionality and is purely just logging changes.

_____________


bretambrose (commented):
I think these statements misinterpret the context a little.  Neither of the added else-clauses represents a failure path.  

The empty list check determines whether or not the event loop already has a task scheduled to slurp the cross-thread queue into the event loop's scheduler; a false check here isn't a failure.

Similarly, the should_cancel check only fails if the channel is still running which is the happy, expected path, meaning the task is going to get executed properly.

TwistedTwigleg:
Thank you for the clarification! I didn't realize these were not failure paths. I adjusted the logging accordingly 👍

TwistedTwigleg:
The issue this logging was for helping debug has been fixed. Closing this PR...

PR #494, opened 2022-05-31
New API to serialize channel task execution by serialized channel task submission

bretambrose:
Review Note:

The diff is confusing, but the old s_register_pending_task function has been cleanly split into three functions:
* s_reset_pending_channel_task - resets the submitted channel task
* s_register_pending_task_in_event_loop - submits the task directly to the event loop scheduler; only valid when called from the appropriate event loop thread
* s_register_pending_task_cross_thread - submits the task to the cross thread queue, always safe to call


TwistedTwigleg (approved):
Awesome! 🎉

PR #501, opened 2022-07-16
Custom TLS key operation support

TwistedTwigleg:
*Description of changes:*

Adds custom key operation support for the TLS handshake when connecting with MQTT, to allow for making your own private key operations.

This PR adds the support needed to expose this functionality to the CRT layers, as well as adjusting the PKCS11 implementation to use the new custom_key_operation workflow.

_________________


source/pkcs11_private.h
```diff
@@ -152,16 +138,49 @@ AWS_IO_API
...
+AWS_IO_API
+void aws_pkcs11_tls_op_handler_destroy(struct aws_pkcs11_tls_op_handler *op_handler);
+
+/**
+ * Performs the PKCS11 TLS private key operation. This is called automatically when performing a MQTT TLS handshake.
+ */
+AWS_IO_API
+void aws_pkcs11_tls_op_handler_do_operation(
+    struct aws_custom_key_op_handler *handler,
+    struct aws_tls_key_operation *operation);
```
graebm, line 167: don't need this in header file, it can be a static function in the .c file

source/pkcs11_private.h
```diff
@@ -152,16 +138,49 @@ AWS_IO_API
...
+    struct aws_byte_cursor cert_file_path,
+    struct aws_byte_cursor cert_file_contents);
 
 /**
- * Given enum, return string like: AWS_TLS_SIGNATURE_RSA -> "RSA"
+ * Removes a reference from the passed-in PKCS11 TLS operation handler. When the reference count reaches
+ * zero, the PKCS11 TLS operation handler will be destroyed.
+ */
+AWS_IO_API
+void aws_pkcs11_tls_op_handler_destroy(struct aws_pkcs11_tls_op_handler *op_handler);
```
graebm, line 159: don't need this either, there's already a destroy() vtable function

source/pkcs11_private.h
```diff
@@ -152,16 +138,49 @@ AWS_IO_API
 int aws_pkcs11_asn1_enc_ubigint(struct aws_byte_buf *const buffer, struct aws_byte_cursor bigint);
 
 /**
- * Given enum, return string like: AWS_TLS_HASH_SHA256 -> "SHA256"
+ * Creates a new PKCS11 TLS operation handler with a reference count set to 1.
  */
 AWS_IO_API
-const char *aws_tls_hash_algorithm_str(enum aws_tls_hash_algorithm hash);
+struct aws_pkcs11_tls_op_handler *aws_pkcs11_tls_op_handler_new(
```
graebm, line 144: just return a generic aws_tls_op_handler

source/pkcs11_tls_op_handler.c
```diff
@@ -0,0 +1,242 @@
...
+static struct aws_custom_key_op_handler *s_aws_custom_key_op_handler_new(
+    struct aws_allocator *allocator,
+    struct aws_pkcs11_tls_op_handler *pkcs11_handler) {
+
+    struct aws_custom_key_op_handler *key_op_handler = aws_custom_key_op_handler_new(allocator);
+    key_op_handler->impl = (void *)pkcs11_handler;
+    key_op_handler->vtable = &s_aws_custom_key_op_handler_vtable;
+
+    return key_op_handler;
+}
```
graebm, line 111: debatable: not worth having this helper function

source/s2n/s2n_tls_channel_handler.c
```diff
@@ -642,68 +642,176 @@ static enum aws_tls_hash_algorithm s_s2n_to_aws_hash_algorithm(s2n_tls_hash_algo
...
+    aws_channel_task_init(
+        &operation->completion_task,
+        s_tls_key_operation_completion_task,
+        operation,
+        "tls_key_operation_completion_task");
+    aws_channel_schedule_task_now(s2n_handler->slot->channel, &operation->completion_task);
+}
+
+void aws_tls_key_operation_complete(struct aws_tls_key_operation *operation, struct aws_byte_cursor output) {
+    AWS_LOGF_DEBUG(
```
graebm, line 742: add null-check

source/s2n/s2n_tls_channel_handler.c
```diff
@@ -642,68 +642,176 @@ static enum aws_tls_hash_algorithm s_s2n_to_aws_hash_algorithm(s2n_tls_hash_algo
...
+        AWS_LS_IO_TLS,
+        "id=%p: TLS key operation complete with %zu bytes of output data",
+        (void *)operation->s2n_handler,
+        output.len);
+    s_tls_key_operation_complete_common(operation, 0, &output);
+}
+
+void aws_tls_key_operation_complete_with_error(struct aws_tls_key_operation *operation, int error_code) {
+    if (error_code == 0) {
```
graebm, line 751: add null-check

include/aws/io/tls_channel_handler.h
```diff
@@ -706,6 +841,65 @@ AWS_IO_API struct aws_byte_buf aws_tls_handler_protocol(struct aws_channel_handl
...
+ * Failure to do so will stall the TLS connection indefinitely and leak memory.
+ */
+AWS_IO_API
+void aws_tls_key_operation_complete_with_error(struct aws_tls_key_operation *operation, int error_code);
+
+/**
+ * Returns the input data that needs to be operated on by the custom key operation.
+ */
+AWS_IO_API
+struct aws_byte_cursor aws_tls_key_operation_get_input(const struct aws_tls_key_operation *operation);
```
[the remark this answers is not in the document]
TwistedTwigleg: Looking at this, I'm not sure it would reduce a copy. Not that I'm against the change, I implemented it as a test but used `git stash` instead of committing it.

We are using `get_input` in two places: [s_aws_pkcs11_tls_op_handler_do_operation](https://github.com/awslabs/aws-c-io/blob/TT-key-op-prototype/source/pkcs11_tls_op_handler.c#L54) and in [s_aws_custom_key_op_handler_perform_operation](https://github.com/awslabs/aws-crt-java/blob/TT-key-op-prototype/src/native/custom_key_op_handler.c#L27).
* In the custom key op handler, we get the byte cursor and immediately call `aws_jni_byte_array_from_cursor`, which populates a JNI array with this data.
  * If we switch this to an `aws_byte_buf`, then I think we actually get an extra copy - as we have to initialize a new aws_byte_buf, populate it with the data from the operation, and then send that data to JNI using `aws_copy_native_array_to_java_byte_array` and manually pointing to pointer to it. However, under the hood, this is what `aws_jni_byte_array_from_cursor` does, **but** it does this directly from the byte cursor itself. I think if we switch this to a byte cursor, we are actually going to end up allocating an additional buffer in this case.
* In the PKCS11 handler stuff, depending on the operation we call either `aws_pkcs11_lib_decrypt` or `aws_pkcs11_lib_sign`.
  * In `aws_pkcs11_lib_decrypt` we read the data straight from the `aws_byte_cursor` itself through it's pointer and length variables. In this case, similar to the custom key operation, I think we would make an extra allocation if we use an aws_byte_buf, since it would copy the data rather than just passing the cursor around.
    * Here though, I can see a possible benefit: It would allow us to know for sure that the memory is there when we go to pass it to the PKCS11 library and we can ensure it doesn't disappear until we are done, while with the cursor, it is theoretically possible it will disappear if on other threads and such (unlikely though). It would also allow us to replace the checks using `AWS_ASSERT(cursor.len <= ULONG_MAX)` with `aws_byte_buf_is_valid`.
  * In `aws_pkcs11_lib_sign` there are different stuff happening depending if it's an RSA signature or a ECDSA. For RSA, we just get the data from the pointer and length properties in the cursor and call `s_pkcs11_sign_helper`, while in ESCA it only is used in a call to `s_pkcs11_sign_helper`. In `s_pkcs11_sign_helper`. it gets the data from the cursor using the pointer and length properties directly.
    * Like with the decrypting though, I can see the benefit of using an `aws_byte_buf` for preventing the memory from possibly being cleared when we do not expect it and for data validation, but I do not think it would remove a copy in this case either.

In both cases, I don't see how the change would reduce a copy. From what I can tell, what would just change the data type we are populating from a cursor to a buffer, which adds some additional memory management code but guarantees it will stick around.

I do not think the change is necessarily a bad thing, especially with the extra guarantees it provides for memory not disappearing if it's no another thread, but from a reducing copies perspective, based on my current experience I do not think it would do this based on what I've seen (and my admittedly limited experience).

____________

Writing this done now so I do not forget - I'm totally cool with going over this in our code review - I just don't want to forget my thoughts on it and so I wanted to capture it while it was fresh.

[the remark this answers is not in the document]
graebm: good research. I'm with you

source/s2n/s2n_tls_channel_handler.c
```diff
@@ -697,17 +697,9 @@ static void s_tls_key_operation_complete_common(
...
-    }
-
-    // Ensure this can only be called once and exactly once. If called again, log an error.
+    /* Ensure this can only be called once and exactly once. */
     size_t complete_count = aws_atomic_fetch_add(&operation->complete_count, 1);
-    if (complete_count != 0) {
-        AWS_LOGF_ERROR(AWS_LS_IO_TLS, "Operation id=%p: Operation is already complete!", (void *)operation);
-        return;
-    }
+    AWS_FATAL_ASSERT(complete_count == 0);
```
graebm, line 702: trivial: writing fatal asserts like this will give users a bit of a clue what's going on
`AWS_FATAL_ASSERT(complete_count == 0 && "TLS key operation marked complete multiple times");`

source/tls_channel_handler.c
```diff
@@ -244,21 +238,19 @@ int aws_tls_ctx_options_init_client_mtls_with_pkcs11(
...
+        pkcs_has_slot_id ? &pkcs_slot_id : NULL);
 
     if (pkcs11_handler == NULL) {
         goto error;
     }
 
-    // CLEANUP
+    /* CLEANUP */
     if (pkcs_lib != NULL) {
         aws_pkcs11_lib_release(pkcs_lib);
```
graebm, line 255: trivial/style: don't have aws_tls_ctx_options_init_client_mtls_with_pkcs11() function do acquire()/release() on the pkcs11_lib. It's going to be "held" by the aws_pkcs11_tls_op_handler()

source/tls_channel_handler.c
```diff
@@ -244,21 +238,19 @@ int aws_tls_ctx_options_init_client_mtls_with_pkcs11(
             aws_string_new_from_cursor(allocator, &pkcs11_options->private_key_object_label);
     }
 
-    struct aws_pkcs11_tls_op_handler *pkcs11_handler = aws_pkcs11_tls_op_handler_new(
+    struct aws_custom_key_op_handler *pkcs11_handler = aws_pkcs11_tls_op_handler_new(
```
graebm, line 241: (commenting on lines above this) you can move all the validation and transforming of byte-cursors into strings over to the aws_pkcs11_tls_op_handler_new() function, instead of having it in tls_channel_handler.c

source/tls_channel_handler.c
```diff
@@ -272,8 +264,27 @@ int aws_tls_ctx_options_init_client_mtls_with_pkcs11(
         aws_string_destroy(pkcs_private_key_object_label);
     }
 
+    struct aws_byte_buf tmp_cert_buf;
+    if (aws_byte_buf_init(&tmp_cert_buf, allocator, 0) != AWS_OP_SUCCESS) {
```
graebm, line 268: this is one of those functions that can't fail, so don't need error checking

actually you don't need this function call at all, you double-initialized this byte-buf

all you need is the aws_byte_buf_init_copy_from_cursor() call

include/aws/io/tls_channel_handler.h
```diff
@@ -356,17 +468,9 @@ struct aws_tls_ctx_pkcs11_options {
...
     struct aws_byte_cursor private_key_object_label;
 
-    /**
-     * Certificate's file path on disk (UTF-8).
-     * The certificate must be PEM formatted and UTF-8 encoded.
-     * Zero out if passing in certificate by some other means (such as file contents).
-     */
-    struct aws_byte_cursor cert_file_path;
-
     /**
```
graebm, line 471: you can't remove this, this is something users might set directly. it would break the API

source/tls_channel_handler.c
```diff
@@ -272,8 +264,27 @@ int aws_tls_ctx_options_init_client_mtls_with_pkcs11(
...
+        AWS_LOGF_ERROR(AWS_LS_IO_TLS, "Could not allocate byte buffer for custom key operation certificate");
+        aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
+        goto error;
+    }
+    if (pkcs11_options->cert_file_contents.ptr == NULL) {
+        AWS_LOGF_ERROR(AWS_LS_IO_TLS, "Certificate file contents are empty");
+        return AWS_OP_ERR;
+    } else {
+        if (aws_byte_buf_init_copy_from_cursor(&tmp_cert_buf, allocator, pkcs11_options->cert_file_contents)) {
```
graebm, line 277: you don't need this at all. if user passes file_contents as byte_cursor, you can just pass that through to aws_tls_ctx_options_init_client_mtls_with_custom_key_operations().

you'll only need this tmp buffer for reading in the contents of a file

source/tls_channel_handler.c
```diff
@@ -272,8 +264,27 @@ int aws_tls_ctx_options_init_client_mtls_with_pkcs11(
...
+            AWS_LOGF_ERROR(AWS_LS_IO_TLS, "Could not allocate byte buffer for custom key operation certificate");
+            return AWS_OP_ERR;
```
graebm, line 279: this function allocates temporary things.
you can't just return AWS_OP_ERR if things go wrong, you need to goto

source/tls_channel_handler.c
```diff
@@ -134,85 +135,135 @@ int aws_tls_ctx_options_init_client_mtls_from_path(
...
     }
+    int custom_key_result;
 
-    /* certificate required, but there are multiple ways to pass it in */
-    if ((pkcs11_options->cert_file_path.ptr != NULL) && (pkcs11_options->cert_file_contents.ptr != NULL)) {
-        AWS_LOGF_ERROR(
-            AWS_LS_IO_TLS, "static: Both certificate filepath and contents are specified. Only one may be set.");
-        aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
+    if ((pkcs11_options->cert_file_contents.ptr != NULL) && (pkcs11_options->cert_file_path.ptr != NULL)) {
         goto error;
```
graebm, line 218: still need to log about how certificate was passed in wrong
still need to aws_raise_error()

source/tls_channel_handler.c
```diff
@@ -134,85 +135,135 @@ int aws_tls_ctx_options_init_client_mtls_from_path(
...
+            options, allocator, pkcs11_handler, &pkcs11_options->cert_file_contents);
+    } else {
+        struct aws_byte_buf tmp_cert_buf;
         struct aws_string *tmp_string = aws_string_new_from_cursor(allocator, &pkcs11_options->cert_file_path);
-        int op = aws_byte_buf_init_from_file(&options->certificate, allocator, aws_string_c_str(tmp_string));
+        int op = aws_byte_buf_init_from_file(&tmp_cert_buf, allocator, aws_string_c_str(tmp_string));
         aws_string_destroy(tmp_string);
+
         if (op != AWS_OP_SUCCESS) {
+            aws_byte_buf_clean_up(&tmp_cert_buf);
```
graebm, line 229: I don't like it when there are multiple places in a function that clean something up.

maybe move the declaration to the top of the function:
```
struct aws_byte_buf tmp_cert_buf;
AWS_ZERO_STRUCT(tmp_cert_buf);
```

and then just always clean it up at the end of the function

source/tls_channel_handler.c
```diff
@@ -134,85 +135,135 @@ int aws_tls_ctx_options_init_client_mtls_from_path(
...
+     */
+    aws_custom_key_op_handler_release(pkcs11_handler);
+
+    return custom_key_result;
 
 error:
+
+    if (pkcs11_handler != NULL) {
+        aws_custom_key_op_handler_release(pkcs11_handler);
+    }
```
graebm, line 254: same thing where I don't like it that 2 code paths in this file release the pkcs11_handler

source/tls_channel_handler.c
```diff
@@ -739,3 +790,94 @@ void aws_tls_ctx_release(struct aws_tls_ctx *ctx) {
...
+        aws_ref_count_release(&key_op_handler->ref_count);
+    }
+    return NULL;
+}
+
+void aws_custom_key_op_handler_perform_operation(
+    struct aws_custom_key_op_handler *key_op_handler,
+    struct aws_tls_key_operation *operation) {
+    if (key_op_handler != NULL) {
+        if (key_op_handler->vtable != NULL && key_op_handler->vtable->on_key_operation != NULL) {
```
graebm, line 879: don't do any of these null checks

we are in REAL trouble if this function doesn't successfully pass ownership along

it SHOULD just crash

graebm (approved):

source/tls_channel_handler.c
```diff
@@ -155,33 +152,21 @@ int aws_tls_ctx_options_init_client_mtls_with_custom_key_operations(
...
-        aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
-        goto error;
-    }
+    AWS_ASSERT(cert_file_contents != NULL);
     aws_byte_buf_init_copy_from_cursor(&options->certificate, allocator, *cert_file_contents);
 
     /* Make sure the certificate is set and valid */
     if (aws_sanitize_pem(&options->certificate, allocator)) {
         AWS_LOGF_ERROR(AWS_LS_IO_TLS, "static: Invalid certificate. File must contain PEM encoded data");
+        aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
```
graebm, line 169: again, aws_sanitize_pem() raises errors

source/tls_channel_handler.c
```diff
@@ -209,52 +194,63 @@ int aws_tls_ctx_options_init_client_mtls_with_pkcs11(
         &pkcs11_options->private_key_object_label,
         pkcs11_options->slot_id);
 
+    struct aws_byte_buf tmp_cert_buf;
+    AWS_ZERO_STRUCT(tmp_cert_buf);
+    bool success = true;
```
graebm, line 199: again, set `success = false` to start
because it's easier to maintain

source/tls_channel_handler.c
```diff
@@ -209,52 +194,63 @@ int aws_tls_ctx_options_init_client_mtls_with_pkcs11(
...
     } else {
-        struct aws_byte_buf tmp_cert_buf;
         struct aws_string *tmp_string = aws_string_new_from_cursor(allocator, &pkcs11_options->cert_file_path);
         int op = aws_byte_buf_init_from_file(&tmp_cert_buf, allocator, aws_string_c_str(tmp_string));
         aws_string_destroy(tmp_string);
 
         if (op != AWS_OP_SUCCESS) {
-            aws_byte_buf_clean_up(&tmp_cert_buf);
-            goto error;
+            aws_raise_error(AWS_ERROR_INVALID_STATE);
```
graebm, line 222: aws_byte_buf_init_from_file() would have raised a good error already

source/pkcs11_tls_op_handler.c
```diff
@@ -0,0 +1,228 @@
...
+
+    /* private_key_object_label is optional */
+    if (match_private_key_label->ptr != NULL) {
+        pkcs_private_key_object_label = aws_string_new_from_cursor(allocator, match_private_key_label);
+    }
+
+    CK_SLOT_ID slot_id;
+    if (aws_pkcs11_lib_find_slot_with_token(pkcs11_handler->lib, match_slot_id, pkcs_token_label, &slot_id /*out*/)) {
+        success = false;
+        aws_raise_error(AWS_ERROR_INVALID_STATE);
```
graebm, line 180: don't re-raise

TwistedTwigleg:
Thanks for all the reviews and help! Merging into `main`...

PR #503, opened 2022-08-01
Fix building with clang

Biswa96:
This fixes two errors.
1. error: use of undeclared identifier 'EPIPE'.
   This is fixed by including errno.h file.
2. error: use of undeclared identifier 'win_error'.
   This is fixed by declaring win_error variable before goto label.


bretambrose:
Merged as part of https://github.com/awslabs/aws-c-io/pull/504

PR #508, opened 2022-08-12
Tolerate s2n_init() being called before aws_io_library_init()

graebm:
**ISSUE:**
A customer reported a failure when aws-c-io called `s2n_init()`. The customer's code was already using S2N-TLS, and they were already calling `s2n_init()` themselves. They were pulling in aws-c-io via a deep dependency chain and weren't even aware that anything else in their application was using S2N-TLS.

In the past, the failing call to `s2n_init()` wasn't being checked, we did start checking for success in this PR https://github.com/awslabs/aws-c-io/pull/461

**DESCRIPTION OF CHANGES**
If S2N-TLS has already been initialized, then aws-c-io does not call `s2n_init()` or `s2n_cleanup()`

### DIFF
--- a/source/s2n/s2n_tls_channel_handler.c
+++ b/source/s2n/s2n_tls_channel_handler.c
@@ -181,8 +181,7 @@ void aws_tls_init_static_state(struct aws_allocator *alloc) {
      * This can cause a crash if s2n is compiled into a shared library and
      * that library is unloaded before the appexit handler runs. */
     if (s2n_disable_atexit() != S2N_SUCCESS) {
-
-        /* If this failed, then s2n is already initialized */
+        AWS_LOGF_DEBUG(AWS_LS_IO_TLS, "static: s2n is already initialized");
         s_s2n_initialized_externally = true;
     }
 
--- a/source/s2n/s2n_tls_channel_handler.c
+++ b/source/s2n/s2n_tls_channel_handler.c
@@ -181,15 +181,12 @@ void aws_tls_init_static_state(struct aws_allocator *alloc) {
      * This can cause a crash if s2n is compiled into a shared library and
      * that library is unloaded before the appexit handler runs. */
     if (s2n_disable_atexit() != S2N_SUCCESS) {
-        if (s2n_errno == S2N_ERR_INITIALIZED) {
-            s_s2n_initialized_externally = true;
-        } else {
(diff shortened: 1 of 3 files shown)


source/s2n/s2n_tls_channel_handler.c
```diff
@@ -168,23 +168,30 @@ static const char *s_determine_default_pki_ca_file(void) {
...
 }
 
+/* If s2n is already initialized, then we don't call s2n_init() or s2n_cleanup() ourselves */
+static bool s_s2n_initialized_externally;
+
 void aws_tls_init_static_state(struct aws_allocator *alloc) {
     (void)alloc;
     AWS_LOGF_INFO(AWS_LS_IO_TLS, "static: Initializing TLS using s2n.");
 
-    setenv("S2N_ENABLE_CLIENT_MODE", "1", 1);
```
ny0312, line 175: We don't need this anymore?

source/s2n/s2n_tls_channel_handler.c
```diff
@@ -168,23 +168,30 @@ static const char *s_determine_default_pki_ca_file(void) {
...
-    setenv("S2N_DONT_MLOCK", "1", 1);
-
     /* Disable atexit behavior, so that s2n_cleanup() fully cleans things up.
      *
      * By default, s2n uses an ataexit handler and doesn't fully clean up until the program exits.
      * This can cause a crash if s2n is compiled into a shared library and
      * that library is unloaded before the appexit handler runs. */
-    s2n_disable_atexit();
+    if (s2n_disable_atexit() != S2N_SUCCESS) {
```
ny0312, line 183: This assumes that the only reason why `s2n_disable_atexit` would fail is that s2n is already initialized. This is correct now, but risks regression imho.

I would explicitly call out/document this assumption, with a link to the source of assumption. E.g. [link](https://code.amazon.com/packages/S2N/blobs/f9dd19fdc78e2d4d95a7b8f6326c3d84ed61d40b/--/third-party-src/api/s2n.h#L211-L212)

source/s2n/s2n_tls_channel_handler.c
graebm: S2N doesn't use that variable for anything anymore https://github.com/aws/s2n-tls/issues/771

ny0312 (approved):

TingDaoK (approved):

PR #513, opened 2022-08-29
Fix for heap access violation

TwistedTwigleg:
*Description of changes:*

Fixes a heap access violation found when implementing AppVerifier into the C++ V2 SDK CI. AppVerifier finds the issue at the end of running a sample and, according to the logs provided by AppVerifier, it is caused by an access violation caused by calling `CertFreeCertificateContext` in `windows/secure_channel_tls_handler.c` in the `s_secure_channel_ctx_destroy` function.

The issue seems to be that `CryptDestroyKey` frees the private key certificate context and so then when `CertFreeCertificateContext` is called to free the private key certificate context further down, it leads to the access violation since it is already free. Freeing the private certificate key store first, before the private key, causes the same issue when trying to free the private key.

The fix implemented is to check if the private key was loaded from the private certificate store and, if it was, not delete the certificate store on shutdown.

_______________


source/windows/secure_channel_tls_handler.c
```diff
@@ -1841,7 +1842,15 @@ static void s_secure_channel_ctx_destroy(struct secure_channel_ctx *secure_chann
...
     if (secure_channel_ctx->pcerts) {
-        CertFreeCertificateContext(secure_channel_ctx->pcerts);
+        /**
+         * Only free the private certificates if the private key is NOT
+         * from the certificate context because freeing the private key
+         * using CryptDestroyKey frees the certificate context and then
+         * trying to access it leads to a access violation.
+         */
+        if (secure_channel_ctx->private_key_from_pcerts == false) {
+            CertFreeCertificateContext(secure_channel_ctx->pcerts);
```
graebm, line 1852: Good sleuthing, just nitpicking variable names. 

maybe rename ~private_key_from_pcerts~ -> `should_free_pcerts` or `pcerts_is_root_object`

That variable name describes the exact bug you're fixing now.

But the real underlying issue is that in one code path pcerts is the root object and needs to be freed. In another code path it's not the root object, so it's not the thing we free. In the future there might be some 3rd code path where pcerts is or isn't a root object, so let's use a variable name that we could reuse there as well

TwistedTwigleg: Good idea - I'll change the name to `should_free_pcerts` so it has better re-usability 👍

TwistedTwigleg: Fixed!

graebm (approved):

TwistedTwigleg:
Thanks! Merging into `main`...

PR #515, opened 2022-09-06
Removes AWS_FATAL_ASSERT in s_aws_event_loop_group_shutdown_async

waahm7:
*Issue #, if available:* https://github.com/aws/aws-sdk-cpp/issues/1809

*Description of changes:*
- Removes fatal assert in s_aws_event_loop_group_shutdown_async

### DIFF
--- a/source/event_loop.c
+++ b/source/event_loop.c
@@ -54,7 +54,7 @@ static void s_event_loop_destroy_async_thread_fn(void *thread_data) {
     aws_thread_current_at_exit(s_event_loop_group_thread_exit, el_group);
 }
 
-static int s_aws_event_loop_group_shutdown_async(struct aws_event_loop_group *el_group) {
+static void s_aws_event_loop_group_shutdown_async(struct aws_event_loop_group *el_group) {
 
     /* It's possible that the last refcount was released on an event-loop thread,
      * so we would deadlock if we waited here for all the event-loop threads to shut down.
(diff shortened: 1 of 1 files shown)


source/event_loop.c
```diff
@@ -63,16 +63,14 @@ static void s_aws_event_loop_group_shutdown_async(struct aws_event_loop_group *e
     struct aws_thread cleanup_thread;
     AWS_ZERO_STRUCT(cleanup_thread);
 
-    AWS_FATAL_ASSERT(aws_thread_init(&cleanup_thread, el_group->allocator) == AWS_OP_SUCCESS);
+    aws_thread_init(&cleanup_thread, el_group->allocator);
```
waahm7, line 66: This always return Success.

source/event_loop.c
```diff
@@ -63,16 +63,14 @@ static void s_aws_event_loop_group_shutdown_async(struct aws_event_loop_group *e
...
 
     struct aws_thread_options thread_options;
     AWS_ZERO_STRUCT(thread_options);
     thread_options.cpu_id = -1;
     thread_options.join_strategy = AWS_TJS_MANAGED;
 
-    AWS_FATAL_ASSERT(
-        aws_thread_launch(&cleanup_thread, s_event_loop_destroy_async_thread_fn, el_group, &thread_options) ==
-        AWS_OP_SUCCESS);
+    aws_thread_launch(&cleanup_thread, s_event_loop_destroy_async_thread_fn, el_group, &thread_options);
```
waahm7, line 73: This can only fail if we cannot create a thread. As the function is void, we cann’t return an AWS_OP_ERR without changing the return type. We use this function as a completion call back during aws_ref_count_init at https://github.com/awslabs/aws-c-io/blob/main/source/event_loop.c#L116 which requires the return type to be void.

bretambrose: If this fails and shutdown is blocking (aws_thread_join_all_managed is used without a timeout) then it will be an indefinite wait.

bretambrose (approved):

PR #523, opened 2022-10-24
Fix HostResolver Callback

waahm7:
*Issue:*
default_resolve_host return value is not consistent. If it returns an error, it may or may not invoke the callback with an error code.

*Description of changes:*
Make it consistent that it should return success if it successfully triggered the callback.


source/host_resolver.c
```diff
@@ -1474,7 +1474,6 @@ static int default_resolve_host(
             res(resolver, host_name, AWS_OP_SUCCESS, &callback_address_list, user_data);
```
graebm, line 1474: found another place where the callbacks can fire, even though setup fails:
on line 1411, `create_and_init_host_entry()` sets up a `struct host_entry` and if something goes wrong at any point while setting it up, calls `s_clean_up_host_entry()`, which iterates over a list of callbacks and invokes them.

We should NOT fire those callbacks unless the `host_entry` was completely set up.

graebm: also I see that we're never checking the return value of `aws_thread_launch()`

source/host_resolver.c
```diff
@@ -1474,7 +1474,6 @@ static int default_resolve_host(
...
         } else {
             res(resolver, host_name, aws_last_error(), NULL, user_data);
```
graebm, line 1476: I have no idea what value `aws_last_error()` will return here
bug?

source/host_resolver.c
waahm7: Good catch, fixed.

source/host_resolver.c
```diff
@@ -1346,10 +1346,7 @@ static inline int create_and_init_host_entry(
     new_host_entry->resolution_config = *config;
     aws_condition_variable_init(&new_host_entry->entry_signal);
 
-    if (aws_thread_init(&new_host_entry->resolver_thread, resolver->allocator)) {
-        goto setup_host_entry_error;
-    }
-
```
waahm7, line 1352: Thread init can never fail. We should just make it void.

source/host_resolver.c
```diff
@@ -1473,8 +1475,9 @@ static int default_resolve_host(
         if (aws_array_list_length(&callback_address_list)) {
             res(resolver, host_name, AWS_OP_SUCCESS, &callback_address_list, user_data);
         } else {
+            /* aws_last_error will be set by aws_host_address_copy or aws_array_list_push_back if something goes wrong.
+             */
             res(resolver, host_name, aws_last_error(), NULL, user_data);
```
TingDaoK, line 1480: Maybe it's better to return error instead of firing the callback? The function failed synced, the return code is more proper to handle the synced failure.

waahm7: Makes sense, I think we should be consistent and just update aws-c-common guidelines to include callback rules or something.

source/host_resolver.c
```diff
@@ -24,14 +24,14 @@ int aws_host_address_copy(const struct aws_host_address *from, struct aws_host_a
...
-        return AWS_OP_ERR;
+        return aws_raise_error(AWS_IO_DNS_QUERY_FAILED);
     }
 
     to->host = aws_string_new_from_string(to->allocator, from->host);
 
     if (!to->host) {
         aws_string_destroy((void *)to->address);
-        return AWS_OP_ERR;
+        return aws_raise_error(AWS_IO_DNS_QUERY_FAILED);
```
graebm, line 34: either undo these changes, or delete these lines entirely.

If these were to fail, it would be due to AWS_ERROR_OOM (out of memory). Reporting AWS_IO_DNS_QUERY_FAILED instead is a much less useful error for the user.

On the other hand, it's no longer possible for aws_string_new_from_string() to fail, because we no longer allow memory allocation to fail.

source/host_resolver.c
```diff
@@ -1473,8 +1475,9 @@ static int default_resolve_host(
         if (aws_array_list_length(&callback_address_list)) {
             res(resolver, host_name, AWS_OP_SUCCESS, &callback_address_list, user_data);
         } else {
-            res(resolver, host_name, aws_last_error(), NULL, user_data);
-            result = AWS_OP_ERR;
+            int last_error = aws_last_error();
+            int error_code = (last_error != AWS_ERROR_SUCCESS) ? last_error : AWS_IO_DNS_QUERY_FAILED;
```
graebm, line 1479: this is not a good way to check whether aws_last_error has been set

nowhere in our codebase do we ever bother to reset aws_last_error back to 0/SUCCESS. When a thread starts, it's 0, but after that it just holds whatever error code was last set, even if it's an old stale value from some other function called long before the function we're currently in

Whenever we call `aws_last_error()`, we need to be sure that something in our current callstack set the error. Usually we ensure that by immediately doing `goto error` after a function reports an error. But this code seems to do a whole bunch of stuff ... and at the end if things aren't perfect it's just kind of hoping (???) that an error was set along the way.

source/host_resolver.c
```diff
@@ -1473,8 +1475,9 @@ static int default_resolve_host(
...
+            return error_code;
```
graebm, line 1480: our `int` functions return either AWS_OP_SUCCESS or AWS_OP_ERROR (and then users call aws_last_error() to get the actual error code). We never return the actual error code itself from an `int`-returning function

graebm: also ... how are we not leaking memory by suddenly doing a return here?

isn't there an aws_array_list_clean_up() that still needs to happen?

source/host_resolver.c
graebm: agree, but disagree

I agree with:
- we should be consistent
- we should update guidelines

But disagree on changing it as Dengke suggested.

In a perfect world, the callback would ALWAYS fire asynchronously. The synchronous failure via aws_raise_error() would be reserved for cases where it's impossible to do the lookup (the input arguments are invalid, thread can't be launched, etc). In cases where we failed to find any results, we'd still report that via callback, even if we knew the answer synchronously.

So let's take one of these paths:
- ALWAYS report results asynchronously. Change this code so we report success and failure asynchronously. We can schedule a task on the event_loop_group to do this. See comment `NOTE: we don't use el_group yet, but we will in the future`
- OR keep changes minimal and continue reporting both success and failure results synchronously

graebm: Here's a long blog post explaining why callbacks that are SOMETIMES sync and SOMETIMES async are bad. You gotta always be one or always be the other:
https://blog.ometer.com/2011/07/24/callbacks-synchronous-and-asynchronous/

graebm: added it to the guidelines:
https://github.com/awslabs/aws-c-common/pull/945

source/host_resolver.c
waahm7: Thanks, that's really informative.

source/host_resolver.c
waahm7: Thanks, I have fixed it.

source/host_resolver.c
waahm7: Thanks, fixed.

source/host_resolver.c
waahm7: I have updated it again, now that I understand more about C error handling.

graebm (approved):

PR #534, opened 2022-12-16
Rename variables in socket tests, for clarity

graebm:
**Issue:**
I could not follow what these tests were doing. Server stuff was named "incoming, and client stuff was named "outgoing", but both sides were sending and receiving data.

**Description of changes:**
Rename server things "server" and client things "client"

### DIFF
--- a/source/channel_bootstrap.c
+++ b/source/channel_bootstrap.c
@@ -1348,7 +1348,7 @@ struct aws_socket *aws_server_bootstrap_new_socket_listener(
         AWS_LS_IO_CHANNEL_BOOTSTRAP,
         "id=%p: attempting to initialize a new "
         "server socket listener for %s:%d",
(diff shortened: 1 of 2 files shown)


source/channel_bootstrap.c
```diff
@@ -1348,7 +1348,7 @@ struct aws_socket *aws_server_bootstrap_new_socket_listener(
         AWS_LS_IO_CHANNEL_BOOTSTRAP,
         "id=%p: attempting to initialize a new "
         "server socket listener for %s:%d",
-        (void *)server_connection_args->bootstrap,
+        (void *)bootstrap_options->bootstrap,
```
graebm, line 1351: previously, this was always logging "NULL" because the member variable hadn't been set yet

waahm7 (approved):

PR #541, opened 2023-01-13
Errno usage pass

bretambrose:
A pass through all uses of errno.  There were a number of places where a system call was made and then the errno check for it occurred after other things that might side-affect the value of errno (like logging).

This PR establishes a new pattern for system call error checking.  The following approaches are acceptable:
```
int result = system_call(...);
int errno_value = errno;
if (result) {
   ... // do stuff with errno_value
}
```

or if there's a concern with thread local access performance?:

```
int result = system_call(...);
if (result) {
   int errno_value = errno;
   ... // do stuff with errno_value
}
```

Applying this pattern universally sometimes leads to what looks like trivial separation:

```
int result = system_call(...);
if (result) {
   int errno_value = errno;
   return aws_raise_error(translate_io_error(errno_value));
}
```

but it's important to follow the pattern everywhere since skipping it in the above case makes it far easier for someone to
accidentally break things in the future by inserting logic between the system call and the aws_raise_error() call, ala:

```
int result = system_call(...);
if (result) {
   AWS_LOGF_ERROR(...);
   return aws_raise_error(translate_io_error(errno));
}
``` 

### DIFF
--- a/source/posix/pipe.c
+++ b/source/posix/pipe.c
@@ -276,7 +276,7 @@ int aws_pipe_read(struct aws_pipe_read_end *read_end, struct aws_byte_buf *dst_b
     ssize_t read_val = read(read_impl->handle.data.fd, dst_buffer->buffer + dst_buffer->len, num_bytes_to_read);
 
     if (read_val < 0) {
-        int errno_value = errno;
+        int errno_value = errno; /* Always cache errno before potential side-effect */
         if (errno_value == EAGAIN || errno_value == EWOULDBLOCK) {
             return aws_raise_error(AWS_IO_READ_WOULD_BLOCK);
         }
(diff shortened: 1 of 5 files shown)


source/posix/pipe.c
```diff
@@ -276,10 +276,11 @@ int aws_pipe_read(struct aws_pipe_read_end *read_end, struct aws_byte_buf *dst_b
     ssize_t read_val = read(read_impl->handle.data.fd, dst_buffer->buffer + dst_buffer->len, num_bytes_to_read);
 
     if (read_val < 0) {
-        if (errno == EAGAIN || errno == EWOULDBLOCK) {
+        int errno_value = errno;
```
graebm, line 279: trivial: suggest we be super duper obvious that this is to avoid bugs, vs just being a style choice
```suggestion
        int errno_value = errno; /* logging may reset error, so cache it */
```

graebm (approved):

PR #542, opened 2023-01-17
Fix: Generate random callback

waahm7:
*Description of changes:*
- Introduce a generate random callback which has user_data as well. We can't use this callback in Swift without user_data.
- Add shutdown callback for `exponential_backoff_strategy`.


include/aws/io/retry_strategy.h
```diff
@@ -108,8 +113,20 @@ struct aws_exponential_backoff_retry_options {
...
     uint64_t (*generate_random)(void);
+
+    /*
+     * By default this will be set to use aws_device_random. If you want something else, set it here.
+     */
+    generate_random_fn *generate_random_impl;
+    /**
+     * Optional user data for the generate random generate_random_impl.
+     */
+    void *generate_random_user_data;
```
graebm, line 129: I'd like to see the code that's using this.

I'm worried that ... now that we're passing a `void *` from another language to C. that other language is going to need to know when C is done using that `void *`. Like, do we need to add a shutdown callback too?

include/aws/io/retry_strategy.h
```diff
@@ -108,8 +113,20 @@ struct aws_exponential_backoff_retry_options {
...
+
+    /** Deprecated. Use generate_random_impl instead
+     * By default this will be set to use aws_device_random. If you want something else, set it here.
+     * */
     uint64_t (*generate_random)(void);
+
+    /*
+     * By default this will be set to use aws_device_random. If you want something else, set it here.
+     */
+    generate_random_fn *generate_random_impl;
```
graebm, line 125: ah, the eternal "what do I call v2" question

include/aws/io/retry_strategy.h
```diff
@@ -35,6 +35,11 @@ typedef void(aws_retry_strategy_on_retry_token_acquired_fn)(
  */
 typedef void(aws_retry_strategy_on_retry_ready_fn)(struct aws_retry_token *token, int error_code, void *user_data);
 
+/**
+ * Optional function to supply your own generate random implementation
+ */
+typedef uint64_t(generate_random_fn)(void *user_data);
```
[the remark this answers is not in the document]
waahm7: Thanks, done.

include/aws/io/retry_strategy.h
waahm7: Any better ideas?

include/aws/io/retry_strategy.h
waahm7: Thanks, that is a great catch. I have added a shutdown callback and relevant changes in Swift at https://github.com/awslabs/aws-crt-swift/pull/147/files

graebm (approved):

PR #546, opened 2023-02-01
Purge address cache

waahm7:
\*Description of changes:*
- Adds a new function to purge cache of a specific address.


source/host_resolver.c
```diff
@@ -300,6 +318,7 @@ static void s_clear_default_resolver_entry_table(struct default_host_resolver *r
     aws_hash_table_clear(table);
 }
 
+/* TODO: add a callback to notify when purge has completed */
```
waahm7, line 321: I will work on this in next PR.

source/host_resolver.c
```diff
@@ -547,6 +567,61 @@ static inline void process_records(
...
+    /* Success if entry doesn't exist in cache. */
+    if (element == NULL) {
+        aws_mutex_unlock(&default_host_resolver->resolver_lock);
+        if (options->on_host_purge_complete_callback != NULL) {
+            /*Sechedule completion callback asynchronouly */
+            struct host_purge_callback_options *purge_callback_options =
+                aws_mem_calloc(default_host_resolver->allocator, 1, sizeof(struct host_purge_callback_options));
+            purge_callback_options->allocator = default_host_resolver->allocator;
+            purge_callback_options->on_host_purge_complete = options->on_host_purge_complete_callback;
+            purge_callback_options->on_host_purge_complete_user_data = options->user_data;
```
waahm7, line 597: Should we keep the resolver alive until callback has fired? The callback doesn’t use resolver, so it is fine. However, it is possible that host_resolver destroy callback is fired before purge is complete.

include/aws/io/host_resolver.h
```diff
@@ -76,10 +81,21 @@ struct aws_host_listener;
 
 struct aws_host_listener_options;
 
+struct aws_host_resolver_purge_host_options {
+    struct aws_host_resolver *resolver;
```
TingDaoK, line 85: I don't know, but I prefer to have the `resolver` as a separate parameter instead included in the options. feel more natural and we mostly follow that pattern.

include/aws/io/host_resolver.h
```diff
@@ -76,10 +81,21 @@ struct aws_host_listener;
...
+    /* the host to purge the cache for */
+    const struct aws_string *host;
+    /* Callback to invoke when the purge is complete */
+    aws_on_host_purge_complete_fn *on_host_purge_complete_callback;
+    /* user_data will be passed as it is in the callback. */
+    void *on_host_purge_complete_user_data;
```
TingDaoK, line 91: trivial: just rename it to `user_data` or `complete_user_data` as the structure name already have `host_purge`.

source/host_resolver.c
```diff
@@ -80,6 +82,12 @@ int aws_host_resolver_purge_cache(struct aws_host_resolver *resolver) {
     return resolver->vtable->purge_cache(resolver);
 }
 
+int aws_host_resolver_purge_host_cache(const struct aws_host_resolver_purge_host_options *options) {
+    AWS_ASSERT(options && options->resolver);
```
TingDaoK, line 86: This a external API, people can really pass NULL to this function or invalid options.

We should do a raise error instead for proper error handling.

source/host_resolver.c
```diff
@@ -547,6 +567,61 @@ static inline void process_records(
...
+                aws_mem_calloc(default_host_resolver->allocator, 1, sizeof(struct host_purge_callback_options));
+            purge_callback_options->allocator = default_host_resolver->allocator;
+            purge_callback_options->on_host_purge_complete = options->on_host_purge_complete_callback;
+            purge_callback_options->on_host_purge_complete_user_data = options->on_host_purge_complete_user_data;
+
+            struct aws_task *task = aws_mem_calloc(default_host_resolver->allocator, 1, sizeof(struct aws_task));
+            aws_task_init(
+                task, s_purge_host_cache_callback_task, purge_callback_options, "async_purge_host_address_task");
+
+            struct aws_event_loop *loop = aws_event_loop_group_get_next_loop(default_host_resolver->event_loop_group);
```
TingDaoK, line 603: ASSERT on `loop!=NULL`

source/host_resolver.c
```diff
@@ -547,6 +567,61 @@ static inline void process_records(
...
 
+static void s_purge_host_cache_callback_task(struct aws_task *task, void *arg, enum aws_task_status status) {
+    (void)status;
+    struct host_purge_callback_options *options = arg;
+    aws_mem_release(options->allocator, task);
+    options->on_host_purge_complete(options->on_host_purge_complete_user_data);
+    aws_mem_release(options->allocator, options);
+}
+
+static int resolver_purge_host_cache(const struct aws_host_resolver_purge_host_options *options) {
```
TingDaoK, line 578: This function doesn't fail at all, can we use void return type?

Rename to `s_resolver_purge_host_cache` as it's a static function. If there are other functions in this code based not following that naming pattern, please fix them as well.

source/host_resolver.c
```diff
@@ -547,6 +567,61 @@ static inline void process_records(
...
+        return AWS_OP_SUCCESS;
+    }
+
+    struct host_entry *host_entry = element->value;
+    AWS_FATAL_ASSERT(host_entry);
+
+    /* Setup the on_host_purge_complete callback. */
+    aws_mutex_lock(&host_entry->entry_lock);
+    host_entry->on_host_purge_complete = options->on_host_purge_complete_callback;
+    host_entry->on_host_purge_complete_user_data = options->on_host_purge_complete_user_data;
```
TingDaoK, line 615: Is there any chance for `host_entry->on_host_purge_complete` already be set? If not, can we ASSERT on that? I think that will be more clear in the code that the callback will not be override and lead to a never-invoked complete callback.

source/host_resolver.c
```diff
@@ -1211,6 +1286,10 @@ static void aws_host_resolver_thread(void *arg) {
     aws_array_list_clean_up(&new_address_list);
     aws_array_list_clean_up(&expired_address_list);
 
+    /* trigger the purge complete callback */
+    if (host_entry->on_host_purge_complete != NULL) {
+        host_entry->on_host_purge_complete(host_entry->on_host_purge_complete_user_data);
```
TingDaoK, line 1291: Why don't we invoke the callback from `s_on_host_entry_shutdown_completion`? I haven't really went through the code about where to invoke the callback, just asking.

include/aws/io/host_resolver.h
waahm7: Thanks, updated.

source/host_resolver.c
waahm7: Thanks, makes sense. I have updated it to throw invalid argument error.

source/host_resolver.c
waahm7: Thanks, fixed.

source/host_resolver.c
waahm7: I thought about using a void, however, this is an interface. In future, there might be other implementations that throw an error. I have fixed the naming.

source/host_resolver.c
waahm7: I have added an assert. We remove it from the hash table after setting these. So next time we will not find the entry in the cache.

source/host_resolver.c
waahm7: The callback is for when host entry is removed from the cache, not when it has shutdown successfully. `aws_thread_current_at_exit` might fail and the callback will never be invoked if we do it in `s_on_host_entry_shutdown_completion`.

include/aws/io/host_resolver.h
waahm7: Thanks, fixed.

source/host_resolver.c
```diff
@@ -80,6 +82,28 @@ int aws_host_resolver_purge_cache(struct aws_host_resolver *resolver) {
     return resolver->vtable->purge_cache(resolver);
 }
 
+int aws_host_resolver_purge_host_cache(
+    struct aws_host_resolver *resolver,
+    const struct aws_host_resolver_purge_host_options *options) {
+    if (!resolver) {
```
TingDaoK, line 88: Let's follow `aws_host_resolver_add_host_listener`. Use `AWS_PRECONDITION` to NULL check on resolver, and resolver vtable, and if there no implementation of `purge_host_cache`, raise error.

Validate the options from the implementation.

TingDaoK (approved):
fix and ship

PR #547, opened 2023-02-09
Check TLS error code on MacOS

TwistedTwigleg:
*Description of changes:*

Adjusts the code for checking the TLS minimum version setting on MacOS to log an error when the error code returned by the MacOS API is not successful. Also adds warning log about TLS 1.3 support on MacOS.

__________


source/darwin/secure_transport_tls_channel_handler.c
```diff
@@ -830,22 +830,28 @@ static struct aws_channel_handler *s_tls_handler_new(
...
-            SSLSetProtocolVersionMin(secure_transport_handler->ctx, kTLSProtocol12);
+            setProtocolStatus = SSLSetProtocolVersionMin(secure_transport_handler->ctx, kTLSProtocol12);
+            AWS_LOGF_TRACE(AWS_LS_IO_TLS, "Attempted to set minimum TLS to: 1.2");
             break;
         case AWS_IO_TLSv1_3:
 #if TLS13_AVAILABLE
-            SSLSetProtocolVersionMin(secure_transport_handler->ctx, kTLSProtocol13);
+            AWS_LOGF_WARN(AWS_LS_IO_TLS, "TLS 1.3 currently not supported. Will use system default TLS instead");
+            setProtocolStatus = SSLSetProtocolVersionMin(secure_transport_handler->ctx, kTLSProtocol13);
 #else
```
waahm7, line 855: The log statement doesn’t seem accurate. Should we update the Tls13_Available flag for now to indicate that it is not supported?

TwistedTwigleg: Good idea! That would be better than logging or throwing a fatal error, as then it wouldn't break anything and would be more easily visible to customers.

TwistedTwigleg:
Closing this PR. If we want to handle TLS 1.3 not being supported on MacOS, just printing a warning is probably not the best solution. This was intended as a short-term solution but given we haven't had any major issues, we can probably just leave it until a full solution is devised and implemented.

For this reason, closing this PR...

PR #555, opened 2023-03-14
Build Emscripten version using poll()

ahmedyarub:
Added a check for Emscripten/WASM build and code required to replace epoll with poll. Tested building the library but not running it.


graebm:
What are you trying to do, that you need to run this in WASM?

We can't take a PR like this without extensive testing, adding CI checks to ensure it runs now and forever. We're a tiny team that owns a large number of libraries and we don't have the bandwidth to take a new feature that vastly increases our testing surface area, unless there's a very strong case for it. You're free to fork, if you need to add these features for your own project

ahmedyarub:
> What are you trying to do, that you need to run this in WASM?
> 
> We can't take a PR like this without extensive testing, adding CI checks to ensure it runs now and forever. We're a tiny team that owns a large number of libraries and we don't have the bandwidth to take a new feature that vastly increases our testing surface area, unless there's a very strong case for it. You're free to fork, if you need to add these features for your own project

I'm creating a WASM application that is used to manage AWS resources. AWS SDK cannot be be built for WASM because epoll is still not implemented in WASM AFAIK. Do you want me to add tests and CI checks?

graebm:
As far as I know, WASM can't interact with OS things like sockets and threads, and these libraries have been built with the assumption they can talk directly to the OS and create such low level resources.

If you're ultimately running in a Javascript environment, I'd suggest using the [AWS SDK for Javascript](https://aws.amazon.com/sdk-for-javascript/)

ahmedyarub:
I'm creating a cross-platform application using Qt 6 that runs on Windows, Linux, MacOS, iOS, Android, and WASM so using a JavaScript library might not be the best option. A couple of years ago I created a version of my application that used the C++ SDK and proxified the connections to WebSockify which worked without any problems. Emscripten itself even has unit-tests for poll() and they definitely work. For more info you can refer to https://emscripten.org/docs/porting/networking.html#emulated-posix-tcp-sockets-over-websockets.
If you look closely at the changes I made you would see that I didn't change anything in previous implementations. The new code is enabled only for Emscripten-based builds. One more benefit here is that it would work on any other platform that doesn't have epoll.

graebm:
We don't have the bandwidth to take on this feature right now, sorry. You can add these changes in a fork to unblock yourself.

PR #559, opened 2023-04-06
DNS Ping Proposal

bretambrose:
https://github.com/aws/aws-iot-device-sdk-python-v2/discussions/399

A set of changes designed to allow us to safely disable the DNS "pinging" functionality on a per-connection basis, while keeping the changes uncoupled to the particulars of the conditionally-undesired behavior and leaving default behavior unchanged.

* Moves host resolution time from system clock to high res clock
* Removes DNS host listener functionality completely.  It is no longer used as of https://github.com/awslabs/aws-c-s3/pull/136
* Un-entangle max-ttl from the DNS refresh frequency.
* * Add support for customizing resolve frequency.  
* * Rework inner resolver thread exit condition to reflect these changes.
* Host resolution configuration can now be overridden at connection kick off time
* Make DNS TTL test more reliable by using a mock clock

Follow-up PRs in aws-c-http and aws-c-mqtt add host resolution override support to their public connection establishment options:
* aws-c-http - https://github.com/awslabs/aws-c-http/pull/434
* aws-c-mqtt - https://github.com/awslabs/aws-c-mqtt/pull/269

Aws-c-mqtt PR also uses the host resolution override to, by default, disable all unsolicited DNS queries when using MQTT, either directly or over websockets.


TwistedTwigleg (approved):
Looks good to me. 👍

DmitriyMusatkin:
`Removes DNS host listener functionality completely. It is no longer used since S3 endpoint DNS now returns large groups of records`

Is this assertion true? from testing it seems that MVA DNS for s3 now returns a max of 8 ips and for 100 gbps nics min amount of ips we want is around 10. Will this cause perf degradation since we are not acquiring enough ips?

source/host_resolver.c
```diff
@@ -20,6 +20,7 @@
 #include <inttypes.h>
 
 const uint64_t NS_PER_SEC = 1000000000;
+const size_t AWS_DEFAULT_DNS_TTL = 30;
```
DmitriyMusatkin, line 23: i know this is not a new constant introduced by this code, but why 30? a lot of aws services ive seen use at least 60. are we refreshing dns too much with default config?

bretambrose: Assertion was incorrect about root reason.  host listener functionality was removed far earlier here: https://github.com/awslabs/aws-c-s3/pull/136

Will update PR commentary, but s3 client will be unaffected

bretambrose:
> `Removes DNS host listener functionality completely. It is no longer used since S3 endpoint DNS now returns large groups of records`
> 
> Is this assertion true? from testing it seems that MVA DNS for s3 now returns a max of 8 ips and for 100 gbps nics min amount of ips we want is around 10. Will this cause perf degradation since we are not acquiring enough ips?


Assertion was incorrect about root reason.  host listener functionality was removed far earlier here: https://github.com/awslabs/aws-c-s3/pull/136

Will update PR commentary, but s3 client will be unaffected

source/host_resolver.c
bretambrose: I don't what, if any, reasons were behind the original choice.

JonathanHenson (approved):
discussed in my office hours, We can move the host resolver config init function to return an instance via value since it's trivially copyable and constructable. 

The cvar wakeup can go outside the mutex lock/unlock. Otherwise looks good to me. Fix and ship.

PR #567, opened 2023-04-21
Add a test for invalid cert

waahm7:
*Description of changes:*
- Adds debug CI for each platform and a test for invalid cert import.

### DIFF
--- a/source/windows/windows_pki_utils.c
+++ b/source/windows/windows_pki_utils.c
@@ -568,15 +568,14 @@ int aws_import_key_pair_to_cert_context(
         return AWS_OP_ERR;
     }
 
-    if (aws_array_list_init_dynamic(&private_keys, alloc, 1, sizeof(struct aws_byte_buf))) {
-        aws_array_list_clean_up(&certificates);
-        return AWS_OP_ERR;
-    }
-
     if (aws_decode_pem_to_buffer_list(alloc, public_cert_chain, &certificates)) {
         AWS_LOGF_ERROR(
             AWS_LS_IO_PKI, "static: failed to decode cert pem to buffer list with error %d", (int)aws_last_error());
         goto clean_up;
+    }
+
+    if (aws_array_list_init_dynamic(&private_keys, alloc, 1, sizeof(struct aws_byte_buf))) {
+        goto cleanup;
     }
 
     if (aws_decode_pem_to_buffer_list(alloc, private_key, &private_keys)) {
--- a/source/windows/windows_pki_utils.c
+++ b/source/windows/windows_pki_utils.c
@@ -575,7 +575,7 @@ int aws_import_key_pair_to_cert_context(
     }
 
     if (aws_array_list_init_dynamic(&private_keys, alloc, 1, sizeof(struct aws_byte_buf))) {
-        goto cleanup;
+        goto clean_up;
     }
 
     if (aws_decode_pem_to_buffer_list(alloc, private_key, &private_keys)) {
--- a/source/windows/windows_pki_utils.c
+++ b/source/windows/windows_pki_utils.c
@@ -568,13 +568,14 @@ int aws_import_key_pair_to_cert_context(
         return AWS_OP_ERR;
     }
 
+    if (aws_array_list_init_dynamic(&private_keys, alloc, 1, sizeof(struct aws_byte_buf))) {
+        aws_array_list_clean_up(&certificates);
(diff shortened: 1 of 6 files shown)


source/windows/windows_pki_utils.c
```diff
@@ -568,16 +568,17 @@ int aws_import_key_pair_to_cert_context(
         return AWS_OP_ERR;
     }
 
+    if (aws_array_list_init_dynamic(&private_keys, alloc, 1, sizeof(struct aws_byte_buf))) {
+        aws_array_list_clean_up(&certificates);
+        return AWS_OP_ERR;
+    }
+
```
graebm, line 575: Our general approach to clean_up() functions is: they should tolerate being passed zeroed-out stuff, so we can naively clean stuff up even if it wasn't fully initialized, to avoid weird fragile cleanup bugs like this...

But aws_cert_chain_clean_up() was hitting an ASSERT because it tries to iterate over an array-list, but the array-list is zeroed out. And when `aws_array_list_size()` is called, it freaks about because the array-list isn't initialized.

I'd propose a different solution: aws_cert_chain_clean_up() should tolerate a zeroed-out list, and the easiest way to achieve that is to make aws_array_list_size() tolerate being called on a zeroed out list. 

Instead of: ~AWS_PRECONDITION(aws_array_list_is_valid(list));~
We can do: `AWS_PRECONDITION(AWS_IS_ZEROED(*list) || aws_array_list_is_valid(list));`

graebm: this goes against our general rule of: the only functions that should tolerate being passed NULL, or zeroed-out-but-uninitialized struct are `clean_up() / release() / destroy()` functions

But since cleaning up an `aws_array_list` requires the user to first iterate over its members and clean them up, I think we should make an exception and let `aws_array_list_size()`

waahm7: Thanks, updated.

tests/tls_handler_test.c
```diff
@@ -2189,6 +2189,41 @@ static int s_test_concurrent_cert_import(struct aws_allocator *allocator, void *
 
 AWS_TEST_CASE(test_concurrent_cert_import, s_test_concurrent_cert_import)
 
+static int s_test_invalid_cert_import(struct aws_allocator *allocator, void *ctx) {
+    (void)ctx;
+    (void)allocator;
+    /* temporarily disable this on linux until we can make CRYPTO_zalloc behave and stop angering ASan */
+#    if defined(__linux__)
+    return AWS_OP_SUCCESS;
+#    endif
```
graebm, line 2198: wtf?
there's got to be another way to solve this
a test that never runs on our most popular OS isn't a great idea

graebm (commented):
looks good except 1 thing

tests/tls_handler_test.c
waahm7: I copied it from `s_test_concurrent_cert_import` test. The leaks are in S2N and I am not sure if we can do anything about it apart from filing a ticket to them.
```
=================================================================
==6208==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 88 byte(s) in 1 object(s) allocated from:
    #0 0x7fdc33b72110 in malloc (/usr/lib64/libasan.so.4+0xdc110)
    #1 0x5acef3 in s2n_mem_malloc_no_mlock_impl /root/aws-c-io/build/deps/s2n/utils/s2n_mem.c:126

Indirect leak of 1051 byte(s) in 8 object(s) allocated from:
    #0 0x7fdc33b72110 in malloc (/usr/lib64/libasan.so.4+0xdc110)
    #1 0x5acef3 in s2n_mem_malloc_no_mlock_impl /root/aws-c-io/build/deps/s2n/utils/s2n_mem.c:126

SUMMARY: AddressSanitizer: 1139 byte(s) leaked in 9 allocation(s).
```

DmitriyMusatkin: what is asan issue with CRYPTO_zalloc?
if we cant resolve it we can add annotations for ignore asan on that function either through annotations or dynamically.

DmitriyMusatkin: we can dynamically disable asan for other peoples code using approaches similar to https://github.com/awslabs/aws-c-io/blob/main/builder.json#L25.
lets file issue with them as well, cause if it triggers asan for us it will trigger asan for other people using our libs or s2n directly

PR #569, opened 2023-05-08
WIP | Fix warnings in public headers

waahm7:


include/aws/io/tls_channel_handler.h
```diff
@@ -254,8 +258,6 @@ struct aws_tls_negotiated_protocol_message {
     struct aws_byte_buf protocol;
 };
 
-static const int AWS_TLS_NEGOTIATED_PROTOCOL_MESSAGE = 0x01;
```
waahm7, line 257: As far as I understand, a `static const int` will always trigger a warning (C5264) in our header checker, indicating that it is unused. This is because it is only used in the `.c` files.

graebm (approved):

PR #573, opened 2023-05-30
aws_async_input_stream

graebm:
An async API for reading a stream of bytes.

### DIFF
--- a/include/aws/io/async_stream.h
+++ b/include/aws/io/async_stream.h
@@ -90,7 +90,8 @@ struct aws_async_input_stream *aws_async_input_stream_release(struct aws_async_i
  * Returns a future, which will contain an error code if something went wrong,
  * or a result bool indicating whether EOF has been reached.
  *
- * WARNING: Do not call read() again until the previous read() is done.
+ * WARNING: The buffer must have space available.
+ * WARNING: Do not read again until the previous read is complete.
  */
 AWS_IO_API
 struct aws_future_bool *aws_async_input_stream_read(struct aws_async_input_stream *stream, struct aws_byte_buf *dest);
@@ -101,6 +102,9 @@ struct aws_future_bool *aws_async_input_stream_read(struct aws_async_input_strea
  * It may complete synchronously. It may complete on another thread.
  * Returns a future, which will contain an error code if something went wrong,
  * or a result bool indicating whether EOF has been reached.
+ *
+ * WARNING: The buffer must have space available.
+ * WARNING: Do not read again until the previous read is complete.
  */
 AWS_IO_API
 struct aws_future_bool *aws_async_input_stream_read_to_fill(
--- a/include/aws/io/async_stream.h
+++ b/include/aws/io/async_stream.h
@@ -12,6 +12,8 @@
  *       this would avoid a copy in the use-cases we know of, but it's more complex
  * TODO: vtable acquire()/release()?
  * TODO: protect against simultaneous reads?
+ * TODO: check results of vtable->read() (i.e. 0 byte reads not allowed)?
+ *       this would require 1 or 2 additional allocations per read
  */
 
 #include <aws/io/io.h>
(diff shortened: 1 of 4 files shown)


source/async_stream.c
```diff
@@ -0,0 +1,194 @@
...
+
+/* Call read() in a loop.
+ * It would be simpler to set a completion callback for each read() call,
+ * but this risks our call stack growing large if there are many small, synchronous, reads.
+ * So be complicated and loop until a read() ) call is actually async,
+ * and only then set the completion callback (which is this same function, where we resume looping). */
+static void s_async_stream_fill_job_loop(void *user_data) {
+    struct aws_async_input_stream_fill_job *job = user_data;
+
+    while (true) {
```
waahm7, line 80: If the underlying stream is sync and data is not available, we will just hang here. Should we introduce a time-off / sleep of some sort? Recently, Swift had a similar issue where CPU usage went to 100%.

graebm: Thanks for pointing this out. aws_input_stream allows 0 byte reads, but the async stream does not. I fixed `s_async_wrapping_synchronous_stream_read()` so that it calls aws_input_stream_read() in a loop if it gets a 0 byte read.

Regarding whether or not we should sleep: I don't think we should. If anyone is hitting this loop and the CPU is spinning at 100%, they should migrate to using an ACTUAL async stream, instead of the old synchronous stream API. This wasn't an option in the past, but now it is!

source/async_stream.c
```diff
@@ -0,0 +1,194 @@
...
+            int error_code = aws_future_bool_get_error(job->read_future);
+            bool eof = error_code ? false : aws_future_bool_get_result(job->read_future);
+            bool reached_capacity = job->dest->len == job->dest->capacity;
+            job->read_future = aws_future_bool_release(job->read_future); /* release and NULL */
+
+            if (error_code || eof || reached_capacity) {
+                /* job complete! */
+                s_async_stream_fill_job_complete(job, eof, error_code);
+                return;
+            }
```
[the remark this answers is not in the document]
graebm: added TODO

source/async_stream.c
```diff
@@ -0,0 +1,209 @@
...
+static const struct aws_async_input_stream_vtable s_async_stream_wrapping_input_stream_vtable = {
+    .destroy = s_async_wrapping_synchronous_stream_destroy,
+    .read = s_async_wrapping_synchronous_stream_read,
+};
+
+struct aws_async_input_stream *aws_async_input_stream_new_from_synchronous(
+    struct aws_allocator *alloc,
+    struct aws_input_stream *source) {
+
+    AWS_PRECONDITION(source);
```
DmitriyMusatkin, line 198: nit: precondition usage seems to be very sporadic. if we are doing it then it should be consistent for all nullable inputs

waahm7 (approved):

include/aws/io/async_stream.h
```diff
@@ -0,0 +1,123 @@
...
+
+/**
+ * Read once from the async stream into the buffer.
+ * The read completes when at least 1 byte is read, the buffer is full, or EOF is reached.
+ * Depending on implementation, the read could complete at any time.
+ * It may complete synchronously. It may complete on another thread.
+ * Returns a future, which will contain an error code if something went wrong,
+ * or a result bool indicating whether EOF has been reached.
+ *
+ * WARNING: Do not call read() again until the previous read() is done.
```
DmitriyMusatkin, line 93: nit: i would add the same warning to the function below as well

DmitriyMusatkin (approved):

source/async_stream.c
```diff
@@ -0,0 +1,209 @@
...
+
+/* Data to perform the aws_async_input_stream_read_to_fill() job */
+struct aws_async_input_stream_fill_job {
+    struct aws_allocator *alloc;
+    struct aws_async_input_stream *stream;
+    struct aws_byte_buf *dest;
+    /* Future for each read() step */
+    struct aws_future_bool *read_future;
+    /* Future to set when this job completes */
+    struct aws_future_bool *my_future;
```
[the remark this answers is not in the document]
graebm: going with `on_complete_future` as the name to give all "final futures" in "job" structs
it will appear in actual code as `fill_job->on_complete_future`

source/async_stream.c
graebm: consistently using it on all args to public functions now
not bothering to check in internal functions

PR #581, opened 2023-06-27
Cleanup aws-lc thread locals in event loop threads

bretambrose:
Requires >= v1.11.0 of aws-lc

Part of the fix for https://github.com/awslabs/aws-crt-nodejs/issues/452

### DIFF
--- a/source/linux/epoll_event_loop.c
+++ b/source/linux/epoll_event_loop.c
@@ -16,8 +16,6 @@
 
 #include <sys/epoll.h>
 
-#include <openssl/thread.h>
-
 #include <errno.h>
 #include <limits.h>
 #include <unistd.h>
--- a/source/linux/epoll_event_loop.c
+++ b/source/linux/epoll_event_loop.c
@@ -15,6 +15,8 @@
 
 #include <sys/epoll.h>
 
+#include <openssl/thread.h>
+
 #include <errno.h>
 #include <limits.h>
 #include <unistd.h>
@@ -562,7 +564,7 @@ static int aws_event_loop_listen_for_io_events(int epoll_fd, struct epoll_event 
     return epoll_wait(epoll_fd, events, MAX_EVENTS, timeout);
 }
 
-static void s_aws_cleanup_aws_lc_thread_local_state(void *user_data) {
+static void s_aws_epoll_cleanup_aws_lc_thread_local_state(void *user_data) {
     (void)user_data;
 
 #if defined(OPENSSL_IS_AWSLC)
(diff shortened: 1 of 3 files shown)


source/linux/epoll_event_loop.c
```diff
@@ -576,6 +584,8 @@ static void aws_event_loop_thread(void *args) {
         return;
     }
 
+    aws_thread_current_at_exit(s_aws_cleanup_aws_lc_thread_local_state, NULL);
```
graebm, line 587: oye.... what's the right place to call this from?
You picked epoll because we use that on Linux
But BSD uses kqueue

`s2n_cleanup()` is similar in that you need to call it on any thread that interacts with S2N, to avoid leaks...
[Here](https://github.com/awslabs/aws-c-io/blob/c3f95ebf3d6d1608e1656bb08eca05e303cc4fd3/source/s2n/s2n_tls_channel_handler.c#L1146) is where the s2n_channel_handler registers that to run on any event-loop thread running S2N. We could add the aws-lc call here too? because anywhere we use AWS-LC we're also using S2N?

Or we could be super blunt and just have **EVERY** aws_thread call this ...

Still doesn't cover misc user threads from other languages that might call `aws_tls_ctx_new()` ...

source/linux/epoll_event_loop.c
```diff
@@ -562,6 +562,14 @@ static int aws_event_loop_listen_for_io_events(int epoll_fd, struct epoll_event
     return epoll_wait(epoll_fd, events, MAX_EVENTS, timeout);
 }
 
+static void s_aws_cleanup_aws_lc_thread_local_state(void *user_data) {
+    (void)user_data;
+
+#if defined(OPENSSL_IS_AWSLC)
+    AWSLC_thread_local_clear();
```
graebm, line 569: Does aws-c-io always know where libcrypto's header files are? is aws-c-io making any other libcrypto calls directly?

Should there be some hideous wrapper in aws_cal_thread_cleanup()?

Or could we modify s2n so `s2n_cleanup()` calls this? Since their library already has the concept of "thread-cleanup" while ours do not have that concept so far

source/linux/epoll_event_loop.c
bretambrose: You could sign in an event loop thread that isn't running a connection so I think all event loop threads should unconditionally clean up, not just ones that s2n has touched.

source/linux/epoll_event_loop.c
bretambrose: I would assume so; aws-c-io has a direct link against against aws-c-cal

I don't think trying to push this into s2n_cleanup would be proper, especially since an event loop thread could conceivably invoke arbitrary aws-lc logic without ever involving s2n (consider a plaintext connection where crypto primitives were invoked in callbacks or the like);

bretambrose: Nm, see what you mean now, although I'm disturbed as to why this worked as is.  Checking the addition of `#include <openssl/base.h>` now

graebm (approved):

graebm:
You need to update this line to get the ProofAlarm to pass
https://github.com/awslabs/aws-c-io/blob/c3f95ebf3d6d1608e1656bb08eca05e303cc4fd3/.github/workflows/proof-alarm.yml#L19

PR #582, opened 2023-07-05
Readonly file

TingDaoK:
*Issue #, if available:*

- https://github.com/awslabs/aws-crt-cpp/issues/508

*Description of changes:*

- Use `rb` instead of `r+b` for supporting read only files. 
- We open file for input stream that we only needs to read from.

### DIFF
--- a/tests/stream_test.c
+++ b/tests/stream_test.c
@@ -54,9 +54,15 @@ static struct aws_input_stream *s_create_read_only_file_stream(struct aws_alloca
     FILE *file = aws_fopen(s_test_file_name, "w+b");
     fwrite(s_simple_binary_test, sizeof(uint8_t), sizeof(s_simple_binary_test), file);
     fclose(file);
+#ifdef _WIN32
+    if (_chmod(s_test_file_name, 0444)) {
+        return NULL;
+    }
+#else
     if (chmod(s_test_file_name, 0444)) {
         return NULL;
     }
+#endif
     return aws_input_stream_new_from_file(allocator, s_test_file_name);
 }
 
--- a/tests/stream_test.c
+++ b/tests/stream_test.c
@@ -11,14 +11,14 @@
 
 #include <sys/stat.h>
 
-#ifdef _wIN32
+#ifdef _WIN32
 #    include <io.h>
 #endif
 
-AwS_STATIC_STRING_FROM_LITERAL(s_simple_test, "SimpleTest");
+AWS_STATIC_STRING_FROM_LITERAL(s_simple_test, "SimpleTest");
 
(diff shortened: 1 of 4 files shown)


tests/stream_test.c
```diff
@@ -46,6 +52,24 @@ static struct aws_input_stream *s_create_binary_file_stream(struct aws_allocator
...
 }
 
+static struct aws_input_stream *s_create_read_only_file_stream(struct aws_allocator *allocator) {
+    remove(s_test_file_name);
+
+    FILE *file = aws_fopen(s_test_file_name, "w+b");
+    fwrite(s_simple_binary_test, sizeof(uint8_t), sizeof(s_simple_binary_test), file);
+    fclose(file);
+#ifdef _WIN32
+    if (_chmod(s_test_file_name, 0444)) {
```
graebm, line 62: trivial: doesn't using the same `s_test_file_name` in every test prevent the tests from running in parallel?

maybe tests should not share file names (randomly generated, or just named after the test)

tests/stream_test.c
```diff
@@ -46,6 +52,24 @@ static struct aws_input_stream *s_create_binary_file_stream(struct aws_allocator
     return aws_input_stream_new_from_file(allocator, s_test_file_name);
 }
 
+static struct aws_input_stream *s_create_read_only_file_stream(struct aws_allocator *allocator) {
+    remove(s_test_file_name);
+
+    FILE *file = aws_fopen(s_test_file_name, "w+b");
```
graebm, line 58: Maybe we should do a purge of all uses "w+" and "r+" in this library. I don't think there are any actual cases where we want to do both read and write. The unnecessary "+" encourages people to keep doing that

tests/stream_test.c
graebm, line 62: Windows doesn't have user/group/other, it only has user. I don't know if 444 really works here or not. I would use enum
```suggestion
    if (_chmod(s_test_file_name, _S_IREAD)) {
```

tests/stream_test.c
```diff
@@ -46,6 +52,24 @@ static struct aws_input_stream *s_create_binary_file_stream(struct aws_allocator
...
+    fwrite(s_simple_binary_test, sizeof(uint8_t), sizeof(s_simple_binary_test), file);
+    fclose(file);
+#ifdef _WIN32
+    if (_chmod(s_test_file_name, 0444)) {
+        return NULL;
+    }
+#else
+    if (chmod(s_test_file_name, 0444)) {
```
graebm, line 66: I'm weirded out that this is  `0444` and not `0x444`
But then ... I've never called `chmod()` from C

Maybe just use the enum `S_IRUSR`?

graebm (approved):

PR #587, opened 2023-07-19
Maximum Backoff Implementation

waahm7:
*Description of changes:*
Add a new maximum backoff option with a default value of 20 seconds. This is an alternative implementation of https://github.com/awslabs/aws-c-io/pull/584, as it more closely aligns with the behavior of the SDKs.


include/aws/io/retry_strategy.h
```diff
@@ -112,6 +112,8 @@ struct aws_exponential_backoff_retry_options {
     size_t max_retries;
     /** Scaling factor to add for the backoff. Default is 25ms */
     uint32_t backoff_scale_factor_ms;
+    /** Maximum retry backoff in seconds. Default is 20 seconds */
+    uint32_t maximum_backoff_s;
```
graebm, line 116: trivial: we don't usually abbreviate seconds as "s"

searching public `aws/**/*.h` I found 4 each of variables ending in `_sec` or `_secs`

graebm: debatable: it's a bit weird to mix different units of time (ms and sec). Worth sticking with ms just so it doesn't confuse people?

source/exponential_backoff_retry_strategy.c
```diff
@@ -184,7 +187,8 @@ typedef uint64_t(compute_backoff_fn)(struct exponential_backoff_retry_token *tok
 
 static uint64_t s_compute_no_jitter(struct exponential_backoff_retry_token *token) {
     uint64_t retry_count = aws_min_u64(aws_atomic_load_int(&token->current_retry_count), 63);
-    return aws_mul_u64_saturating((uint64_t)1 << retry_count, token->backoff_scale_factor_ns);
+    return aws_min_u64(
+        token->maximum_backoff_ns, aws_mul_u64_saturating((uint64_t)1 << retry_count, token->backoff_scale_factor_ns));
```
graebm, line 191: trivial: 2 much math 4 1 line
```suggestion
    uint64_t backoff_ns = aws_mul_u64_saturating((uint64_t)1 << retry_count, token->backoff_scale_factor_ns);
    return aws_min_u64(backoff_ns, token->maximum_backoff_ns);
```

source/exponential_backoff_retry_strategy.c
```diff
@@ -199,7 +203,9 @@ static uint64_t s_compute_deccorelated_jitter(struct exponential_backoff_retry_t
         return s_compute_full_jitter(token);
     }
 
-    return s_random_in_range(token->backoff_scale_factor_ns, aws_mul_u64_saturating(last_backoff_val, 3), token);
+    return aws_min_u64(
+        token->maximum_backoff_ns,
+        s_random_in_range(token->backoff_scale_factor_ns, aws_mul_u64_saturating(last_backoff_val, 3), token));
 }
```
graebm, line 209: you'll want to do `min(X, max_backoff)` before you do `random_in_range()`

if you clamp it after the random(), then you'll have a random number 0->infinity, clamped to MAX. So at some point that random number will almost always be insanely large, so it will pretty much always get clamped to MAX, so they're won't be any randomness anymore

waahm7: I am not sure what the right answer is, but I copied the formula from https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/. 

Some thoughts:
- A cap is also in place on the maximum number of retries, which should mitigate instances where the random number becomes exceedingly large.
- Another thing is that `last_backoff_val` cannot be greater than `maximum_backoff`. So at most, it would be `3 * maximum_backoff` and there will be a 66% chance of getting the `maximum_backoff` again. There will still be an element of randomness.

graebm: the blog post formula for decorrelated jitter is:
`sleep = min(cap, random_between(base, sleep * 3))`

they do random before the clamp, like I said

graebm: agreed, it doesn't super duper matter, since we rolled a bunch of random sleeps to get here, so we're already "jittered" from whatever other clients are doing

but it's weird to have a random number generator which, after running for a while, starts returning the same number 2/3 of the time

???

include/aws/io/retry_strategy.h
waahm7: I have changed it to `secs`. I think it's fine to have different units, as we mention them in the name. Updating it to `ms` leads to the question of whether to allow 1ms or have a minimum of 1000ms.

include/aws/io/retry_strategy.h
```diff
@@ -112,6 +112,8 @@ struct aws_exponential_backoff_retry_options {
     size_t max_retries;
     /** Scaling factor to add for the backoff. Default is 25ms */
     uint32_t backoff_scale_factor_ms;
+    /** Maximum retry backoff in seconds. Default is 20 seconds */
+    uint32_t maximum_backoff_secs;
```
graebm, line 116: trivial: the existing `max_retries` already abbreviated "maximum"
```suggestion
    uint32_t max_backoff_secs;
```

graebm (approved):
fix & ship

PR #589, opened 2023-08-07
Fix S2N blocked issue

waahm7:
*Description of changes:*
Fixes the use of `s2n_recv`. The blocked value is only guaranteed to be meaningful if s2n_error_get_type(s2n_errno) == S2N_ERR_T_BLOCKED. If the error type is something else, the blocked value is meaningless and cannot be relied on for any logic. Currently, the return values are all treated the same and if any fatal error occurs we can end up in an endless loop because blocked value can be garbage. 

Example of how to integrate properly [here](https://github.com/aws/s2n-tls/blob/138e46b15fe4105696f6460d8a596bac39ed0ab1/docs/examples/s2n_recv.c#L47). Note that you don't really even need to read the blocked value at all. It's simply there to tell you what type of blocking happened.

Another issue that can occur when multi-record mode is enabled is S2N will fill up the provided buffer entirely and the blocked status will have a "junk" value. Since the return value is >0 and s2n_error_get_type(s2n_errno) is not S2N_ERR_T_BLOCKED. The blocked value does end up with S2N_BLOCKED_ON_READ so, luckily, the loop exits. But as 
 there's no event to wake it back up because it wasn't actually blocked on a socket operation and we will end up never reading the rest of the data.


source/s2n/s2n_tls_channel_handler.c
```diff
@@ -558,7 +558,21 @@ static int s_s2n_handler_process_read_message(
 
         if (read < 0) {
             aws_mem_release(outgoing_read_message->allocator, outgoing_read_message);
-            continue;
+
+            /* the socket blocked so exit from the loop */
+            if (s2n_error_get_type(s2n_errno) == S2N_ERR_T_BLOCKED) {
```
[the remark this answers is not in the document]
waahm7: We do have to make an extra read to make sure that we have read everything.

> But maybe put back that && blocked == S2N_NOT_BLOCKED in the while loop, so we can break out of the loop earlier if we read > 0 data, and there's not more to read.

The blocked value is garbage if read > 0, so we cannot rely on that for any logic.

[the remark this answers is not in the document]
waahm7: I think we do have to make an extra read to make sure that we have read everything.

> But maybe put back that && blocked == S2N_NOT_BLOCKED in the while loop, so we can break out of the loop earlier if we read > 0 data, and there's not more to read.

The blocked value is garbage if `s2n_error_get_type(s2n_errno) != S2N_ERR_T_BLOCKED`, and there will be no error if `read>0`.

include/aws/io/io.h
```diff
@@ -250,6 +250,8 @@ enum aws_io_errors {
     AWS_IO_STREAM_SEEK_UNSUPPORTED,
     AWS_IO_STREAM_GET_LENGTH_UNSUPPORTED,
 
+    AWS_IO_TLS_ERROR_READ_FAILURE,
```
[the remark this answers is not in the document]
waahm7: Thanks, good catch. This is confusing. 

Debateable:  
For Windows: I have updated the error to `AWS_IO_TLS_ERROR_READ_FAILURE`, as `AWS_IO_TLS_ERROR_NEGOTIATION_FAILURE` is incorrect. We can see the logs to see the exact error that occurred.

For Apple: We can update to read failure, but since it has the concept of graceful vs. non-graceful shutdown, so `AWS_IO_TLS_ERROR_ALERT_RECEIVED` kinda makes sense. But also, that is the only place that we throw that error.

source/io.c
```diff
@@ -300,6 +300,10 @@ static struct aws_error_info s_errors[] = {
     AWS_DEFINE_ERROR_INFO_IO(
         AWS_IO_STREAM_GET_LENGTH_UNSUPPORTED,
         "Get length is not supported in the underlying I/O source."),
+    AWS_DEFINE_ERROR_INFO_IO(
+        AWS_IO_TLS_ERROR_READ_FAILURE,
+        "Failed to read during TLS"),
```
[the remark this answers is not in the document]
waahm7: Thanks, updated.

source/s2n/s2n_tls_channel_handler.c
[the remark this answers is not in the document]
graebm: Looking at the s2n source code, `blocked` does appear to have meaning when `> 0` bytes are read.

at the [start of the function](https://github.com/aws/s2n-tls/blob/05ae36ce470f20233821f3cb000140124023b6bb/tls/s2n_recv.c#L137) it's initialized to:
`*blocked = S2N_BLOCKED_ON_READ;`

But then at the end, before returning the number of bytes processed, it [checks to see if there's more data available](https://github.com/aws/s2n-tls/blob/05ae36ce470f20233821f3cb000140124023b6bb/tls/s2n_recv.c#L231-L235):
```
    if (s2n_stuffer_data_available(&conn->in) == 0) {
        *blocked = S2N_NOT_BLOCKED;
    }

    return bytes_read;
```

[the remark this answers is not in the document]
waahm7: The way I understand it is that the issue with having that condition in the while loop for `read > 0` is as follows: it tries to see if we have read everything. If yes, then it sets the status to NOT_BLOCKED, and in the next call, because `blocked == NOT_BLOCKED`, it will return that there was no data available. However, if the buffer was full and there is still data available, the status would be blocked, and we will never read the remaining data.

source/windows/secure_channel_tls_handler.c
```diff
@@ -1144,8 +1144,7 @@ static int s_do_application_data_decrypt(struct aws_channel_handler *handler) {
         } else {
```
graebm, line 1144: I'm noticing that many lines up, if this is false: https://github.com/awslabs/aws-c-io/blob/a3555c86bd10149a1ab0c1b2810756ec54cfbe3a/source/windows/secure_channel_tls_handler.c#L1086

then this function returns AWS_OP_ERR without having ever called `aws_raise_error()`.

Let's add an `else` branch and have it do `AWS_IO_TLS_ERROR_READ_FAILURE` there as well

include/aws/io/io.h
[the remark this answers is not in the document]
graebm: I would also change the Apple code from ~AWS_IO_TLS_ERROR_ALERT_RECEIVED~ to `AWS_IO_TLS_ERROR_READ_FAILURE`.

I mean, it's the same situation of "this function returned any error other than WOULD_BLOCK"

source/s2n/s2n_tls_channel_handler.c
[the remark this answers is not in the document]
graebm: oh yeah, you're right.
we want to bail out of the loop if read > 0 and NOT_BLOCKED

include/aws/io/io.h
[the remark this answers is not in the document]
waahm7: Thanks, updated.

source/s2n/s2n_tls_channel_handler.c
```diff
@@ -573,6 +588,12 @@ static int s_s2n_handler_process_read_message(
         } else {
             aws_mem_release(outgoing_read_message->allocator, outgoing_read_message);
         }
+
+        if (blocked == S2N_NOT_BLOCKED) {
+            /* read is complete so shutdown with success */
+            aws_channel_shutdown(slot->channel, AWS_OP_SUCCESS);
+            return AWS_OP_SUCCESS;
+        }
```
graebm: this is not correct
I'm pretty sure S2N_NOT_BLOCKED means we've processed all available data, so we can exit the loop, but we might get more data from the socket in the future

it doesn't mean the connection is over

waahm7: Yeah, I have fixed it. I confused return value `0` and `read > 0 + NOT_BLOCKED`.

source/windows/secure_channel_tls_handler.c
waahm7: I have added a raise_error in that function.

source/windows/secure_channel_tls_handler.c
```diff
@@ -1111,6 +1111,12 @@ static int s_do_application_data_decrypt(struct aws_channel_handler *handler) {
                         "id=%p: Decrypt ended exactly on the end of the record, resetting buffer.",
                         (void *)handler);
                 }
+            } else {
+                AWS_LOGF_ERROR(
+                    AWS_LS_IO_TLS,
+                    "id=%p: Error decrypting message. Unexpected type of output buffer.",
```
graebm, line 1117: ask henso about this case
if this is the right way to handle it, or if he intended something else to happen

source/s2n/s2n_tls_channel_handler.c
```diff
@@ -573,6 +588,11 @@ static int s_s2n_handler_process_read_message(
         } else {
             aws_mem_release(outgoing_read_message->allocator, outgoing_read_message);
         }
+
+        if (blocked == S2N_NOT_BLOCKED) {
+            /* read is complete */
+            break;
+        }
```
camshaft, line 595: why is this here? you don't need to read the blocked status at all

waahm7: Thanks for reviewing. We allocate a buffer before every call to recv. This is just an optimization that allows us to break out of the loop early if the read is complete. This way, we avoid an unnecessary recv that would return a read less than 0 with blocked status, allowing us to break out of the loop.

camshaft: If you want to make sure the s2n-tls buffer has been drained, it would be much less fragile to use `s2n_peek`

```suggestion
        if (s2n_peek(conn) == 0) {
            /* the s2n-tls buffer has been completely processed */
            break;
        }
```

That being said... do you know what will register epoll interest for future reads? I'm not familiar with CRT but you usually drain the socket until you get blocked and then register epoll at that point.

camshaft: But in general, if you're receiving a lot of data from the socket it's better to loop and eat the extra allocation over waiting for epoll to wake you up for every single record.

source/s2n/s2n_tls_channel_handler.c
```diff
@@ -558,9 +558,24 @@ static int s_s2n_handler_process_read_message(
...
+
+            /* the socket returned a fatal error so shut down */
+            AWS_LOGF_ERROR(
+                AWS_LS_IO_TLS,
+                "id=%p: S2N failed to read with error: %s (%s)",
+                (void *)handler,
+                s2n_strerror(s2n_errno, "EN"),
+                s2n_strerror_debug(s2n_errno, "EN"));
+            aws_channel_shutdown(slot->channel, AWS_IO_TLS_ERROR_READ_FAILURE);
+            return aws_raise_error(AWS_IO_TLS_ERROR_READ_FAILURE);
```
graebm, line 575: we don't need to shutdown the channel, the function calling us will do that since we're returning an error
```suggestion
            return aws_raise_error(AWS_IO_TLS_ERROR_READ_FAILURE);
```

source/s2n/s2n_tls_channel_handler.c
```diff
@@ -558,9 +558,24 @@ static int s_s2n_handler_process_read_message(
...
         };
 
+        // if read > 0
```
graebm, line 578: utterly trivial: consistency with other comments in this file
```suggestion
        /* if read > 0 */
```

graebm (approved):

source/s2n/s2n_tls_channel_handler.c
waahm7: Thanks, discussed offline that we do need to shutdown the channel but we can not return an error because that will result in a double free.

PR #594, opened 2023-08-18
log as warning for socket connect and client bootstrap

TingDaoK:
*Issue #, if available:*

- Our client bootstrap will try to connect for multiple address resolved by the host resolver. And it may have failure for some address, as long as any one succeed, we established the connection and ignore the error came from other address. Especially, as we do happy eyeball, it will try to use ipv6 and ipv4 at the same time. Most of the case, one of the connection will fail
- It's very annoy when an error level log shows up, but it's kind expected.

*Description of changes:*
- downgrade the level we log the error from socket connect

### DIFF
--- a/source/channel_bootstrap.c
+++ b/source/channel_bootstrap.c
@@ -595,11 +595,18 @@ task_cancelled:
     if (task_data->args->failed_count == task_data->args->addresses_count) {
         AWS_LOGF_ERROR(
             AWS_LS_IO_CHANNEL_BOOTSTRAP,
-            "id=%p: failed to create socket with error %d",
+            "id=%p: Last attempt failed to create socket with error %d",
             (void *)task_data->args->bootstrap,
             err_code);
         s_connection_args_setup_callback(task_data->args, err_code, NULL);
+    } else {
+        AWS_LOGF_DEBUG(
+            AWS_LS_IO_CHANNEL_BOOTSTRAP,
+            "id=%p: failed to create socket with error %d. More attempts ongoing.",
+            (void *)task_data->args->bootstrap,
+            err_code);
     }
+
     s_client_connection_args_release(task_data->args);
 
 cleanup_task:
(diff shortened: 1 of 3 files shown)


source/posix/socket.c
```diff
@@ -674,7 +674,7 @@ int aws_socket_connect(
 
     if (pton_err != 1) {
```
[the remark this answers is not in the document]
graebm: And for IOCP, it would be here:
https://github.com/awslabs/aws-c-io/blob/c4b661f44497b18201b56ffd200cc478441f6434/source/windows/iocp/socket.c#L692-L702

and here: https://github.com/awslabs/aws-c-io/blob/c4b661f44497b18201b56ffd200cc478441f6434/source/windows/iocp/socket.c#L753-L763

source/channel_bootstrap.c
```diff
@@ -592,17 +588,16 @@ static void s_attempt_connection(struct aws_task *task, void *arg, enum aws_task
...
-        AWS_LS_IO_CHANNEL_BOOTSTRAP,
-        "id=%p: failed to create socket with error %d",
-        (void *)task_data->args->bootstrap,
-        err_code);
 task_cancelled:
+    err_code = aws_last_error();
     task_data->args->failed_count++;
     /* if this is the last attempted connection and it failed, notify the user */
     if (task_data->args->failed_count == task_data->args->addresses_count) {
+        AWS_LOGF_ERROR(
```
graebm, line 596: probably still want to log (at lesser level) when failed_count != addresses_count

source/channel_bootstrap.c
```diff
@@ -595,11 +595,18 @@ static void s_attempt_connection(struct aws_task *task, void *arg, enum aws_task
...
+            "id=%p: Last attempt failed to create socket with error %d",
             (void *)task_data->args->bootstrap,
             err_code);
         s_connection_args_setup_callback(task_data->args, err_code, NULL);
+    } else {
+        AWS_LOGF_DEBUG(
+            AWS_LS_IO_CHANNEL_BOOTSTRAP,
+            "id=%p: failed to create socket with error %d. More attempts ongoing.",
+            (void *)task_data->args->bootstrap,
+            err_code);
```
graebm, line 607: trivial: `id=%p: Socket connect attempt %d/%d failed with error %d. More attempts ongoing...`

graebm (approved):

PR #602, opened 2023-09-16
Enable pkcs11 in sanitizer jobs

sfod:


sfod:
To whoever comes across this PR and wonders why it silently sits here.
The solution provided in the PR works just fine. But it's kinda ugly, and cryptic, and probably brittle. So I'm going to park it for some time, maybe I'll find a cleaner solution.

PR #605, opened 2023-10-05
refactor pem

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*


.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
+    "files.associations": {
+        "*.h": "c",
+        "*.c": "c",
+        "iosfwd": "c"
+    }
+}
```
TingDaoK, line 7: get this out of here, lol.

We can add `.vscode` to the .gitignore, I added it to my system one, but we did add it to some of our repos.

DmitriyMusatkin: added to gitignore. and removed the file

include/aws/io/pem.h
```diff
@@ -0,0 +1,95 @@
...
+ * Describes PEM object decoded from file.
+ * data points to raw data bytes of object (decoding will do additional base 64
+ * decoding for each object).
+ * type will be set to object type or to AWS_PEM_TYPE_UNKNOWN if it could not
+ * figure out type.
+ * type_buf are the types bytes, i.e. the string between -----BEGIN and -----
+ */
+struct aws_pem_object {
+    enum aws_pem_object_type type;
+    struct aws_byte_buf type_buf;
```
graebm, line 52: trivial: It's a string, so why not save it as a string
```suggestion
    struct aws_string *type_str;
```

include/aws/io/pem.h
```diff
@@ -0,0 +1,95 @@
...
+
+/*
+ * Describes PEM object decoded from file.
+ * data points to raw data bytes of object (decoding will do additional base 64
+ * decoding for each object).
+ * type will be set to object type or to AWS_PEM_TYPE_UNKNOWN if it could not
+ * figure out type.
+ * type_buf are the types bytes, i.e. the string between -----BEGIN and -----
+ */
+struct aws_pem_object {
```
graebm, line 50: ughhh: putting a struct definition in a public header means it's a breaking change to ever modify it. But putting it in a private header forces an additional allocation when someone could have stored it by-value.

I'm OK putting this here. Name/value pair seems like a pretty stable thing. I just had to vent my frustration with C programming

include/aws/io/pem.h
```diff
@@ -0,0 +1,95 @@
...
+ * If it comes across an object it cannot read, list of all object read until
+ * that point is returned.
+ * If no objects can be read PEM or objects could not be base 64 decoded,
+ * AWS_ERROR_PEM_MALFORMED_OBJECT is raised.
+ * out_pem_objects stores aws_pem_object struct by value.
+ * Caller must initialize out_pem_objects before calling the function.
+ * This code is slow, and it allocates, so please try
+ * not to call this in the middle of something that needs to be fast or resource sensitive.
+ */
+AWS_IO_API int aws_read_and_decode_pem_file_to_object_list(
```
graebm, line 89: debatable: would it be simpler if these functions were alternate array-list initializers? I can't imagine this function ever being used without the user needing to initialize the list first. Like:

`int aws_array_list_init_from_pem_file_contents(struct aws_array_list *list, struct aws_allocator, struct aws_byte_cursor pem_contents);`

`int aws_array_list_init_from_pem_file_path(struct aws_array_list *list, struct aws_allocator, const char *pem_path);`

Or call it "pem_objects" instead of "array_list" (I like this idea best)?

`int aws_pem_objects_init_from_file_contents(struct aws_array_list *pem_objects, struct aws_allocator, struct aws_byte_cursor pem_contents);`

`int aws_pem_objects_init_from_file_path(struct aws_array_list *pem_objects, struct aws_allocator, const char *pem_path);`

include/aws/io/pem.h
```diff
@@ -0,0 +1,95 @@
...
+    enum aws_pem_object_type type;
+    struct aws_byte_buf type_buf;
+    struct aws_byte_buf data;
+};
+
+/**
+ * Cleans up and securely zeroes out the outputs of 'aws_decode_pem_to_object_list()'
+ * and 'aws_read_and_decode_pem_file_to_object_list()'
+ */
+AWS_IO_API void aws_pem_objects_clean_up(struct aws_array_list *pem_objects);
```
graebm, line 60: I'm iffy on this thing's name. It does "clean_up" the inner contents, but it doesn't "clean_up" the datastructure itself.

If we turn the "decode" functions into "init" functions, then it would make sense for this to clean_up the array-list as well.

But if we keep the "decode" functions as-is, I vote we name this like "clear" instead of "clean_up" so it's obvious the user still needs to call aws_array_list_clean_up() afterwards

graebm (commented):
reviewed the public header

include/aws/io/io.h
```diff
@@ -252,6 +252,8 @@ enum aws_io_errors {
 
     AWS_IO_TLS_ERROR_READ_FAILURE,
 
+    AWS_ERROR_PEM_MALFORMED_OBJECT,
```
graebm, line 255: Do you expect other systems to use a different error if the objects themselves are all valid, but the "document" itself is malformed due to missing required objects?

If we fear people will re-use this error-code for bad documents, give it a more generic name like `AWS_ERROR_PEM_MALFORMED`. But if you want distinct error-codes to be created for malformed documents of a specific type, then this is good

include/aws/io/pem.h
DmitriyMusatkin: yeah, it was a conscious decision to put it into header, since it should be stable and i dont envision it ever change (famous last words?)

include/aws/io/pem.h
```diff
@@ -0,0 +1,95 @@
...
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+#include <aws/io/io.h>
+
+AWS_EXTERN_C_BEGIN
+
+enum aws_pem_object_type {
+    AWS_PEM_TYPE_UNKNOWN = 0,
+    AWS_PEM_TYPE_X509_OLD,
+    AWS_PEM_TYPE_X509,
```
graebm, line 15: is there a more specific name than "old"? or is this purposefully vague, lumping together several types we no longer support? Add comment if this is something worth explaining

include/aws/io/io.h
DmitriyMusatkin: let me just make it generic AWS_ERROR_PEM_MALFORMED. i dont think its worth distinguishing that its individual objects that are malformed

include/aws/io/pem.h
DmitriyMusatkin: its trying to stay close to openssl naming for pem types. i can add a comment on the expected type string for each in file

include/aws/io/pem.h
DmitriyMusatkin: Current names were mostly based on existing names for internal function with a modification to refer to elements as pem objects instead of cert_or_key. 
I like your suggestion of just naming it pem_objects_init. next commit switches to that.

I think original code wanted array to be initialized outside of the call because the caller might have better context on how many elements are there in the pem file. But in practice the answer to that is either probably just one or probably many, so initializing outside was not super helpful. Moved array init inside of the function.

include/aws/io/pem.h
```diff
@@ -0,0 +1,100 @@
...
+    struct aws_string *type_string;
+    struct aws_byte_buf data;
+};
+
+/**
+ * Cleans up elements of pem_objects list 'aws_pem_objects_init_from_file_contents()'
+ * and 'aws_pem_objects_init_from_file_path()'.
+ * Does not clean_up list itself.
+ */
+AWS_IO_API void aws_pem_objects_clear(struct aws_array_list *pem_objects);
```
graebm, line 65: well, now that we have aws_pem_objects_init(), it makes a lot of sense to have a complementary aws_pem_objects_clean_up() 😬

But I'm fine either way, this is a pretty niche API

include/aws/io/pem.h
```diff
@@ -0,0 +1,100 @@
...
+ * AWS_ERROR_PEM_MALFORMED is raised.
+ * out_pem_objects stores aws_pem_object struct by value.
+ * Function will initialize pem_objects list.
+ * This code is slow, and it allocates, so please try
+ * not to call this in the middle of something that needs to be fast or resource sensitive.
+ */
+AWS_IO_API int aws_pem_objects_init_from_file_contents(
+    struct aws_allocator *alloc,
+    struct aws_byte_cursor pem_cursor,
+    struct aws_array_list *out_pem_objects);
```
graebm, line 81: the traditional order in an init() function is: object-to-init, allocator, args...
```suggestion
AWS_IO_API int aws_pem_objects_init_from_file_contents(
    struct aws_array_list *pem_objects,
    struct aws_allocator *alloc,
    struct aws_byte_cursor pem_cursor)
```

include/aws/io/pem.h
```diff
@@ -0,0 +1,100 @@
...
+ * AWS_ERROR_PEM_MALFORMED is raised.
+ * out_pem_objects stores aws_pem_object struct by value.
+ * Function will initialize pem_objects list.
+ * This code is slow, and it allocates, so please try
+ * not to call this in the middle of something that needs to be fast or resource sensitive.
+ */
+AWS_IO_API int aws_pem_objects_init_from_file_path(
+    struct aws_allocator *allocator,
+    const char *filename,
+    struct aws_array_list *out_pem_objects);
```
graebm, line 97:
```suggestion
AWS_IO_API int aws_pem_objects_init_from_file_path(
    struct aws_array_list *pem_objects,
    struct aws_allocator *allocator,
    const char *filename);
```

graebm (approved):

PR #606, opened 2023-10-09
add format unknown

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*
SecItemImport overrides import format after the call, so we need to reset it to unknown before every call
https://developer.apple.com/documentation/security/1395728-secitemimport?language=objc

### DIFF
--- a/source/darwin/darwin_pki_utils.c
+++ b/source/darwin/darwin_pki_utils.c
@@ -66,7 +66,6 @@ int aws_import_ecc_key_into_keychain(
         AWS_ZERO_STRUCT(import_params);
         import_params.version = SEC_KEY_IMPORT_EXPORT_PARAMS_VERSION;
         import_params.passphrase = CFSTR("");
-        format = kSecFormatUnknown;
 
         OSStatus key_status =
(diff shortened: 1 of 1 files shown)


source/darwin/darwin_pki_utils.c
```diff
@@ -66,7 +66,6 @@ int aws_import_ecc_key_into_keychain(
         AWS_ZERO_STRUCT(import_params);
         import_params.version = SEC_KEY_IMPORT_EXPORT_PARAMS_VERSION;
         import_params.passphrase = CFSTR("");
-        format = kSecFormatUnknown;
```
graebm, line 69: A few lines up this is initialized as `SecExternalFormat format = kSecFormatOpenSSL;` each time we loop.

That's on purpose, right? 

It seems OK to me, but the PR description "we need to reset it to unknown before every call", but maybe the PR description is describing the NEXT change?

graebm (approved):

source/darwin/darwin_pki_utils.c
DmitriyMusatkin: yes, that was an error in my previous pr. we dont need unknown format here, since it already specifies a format couple lines above.

PR #619, opened 2024-01-24
Cache the Error Before Any Side Effects

waahm7:
*Description of changes:*
- Cache the error before doing any cleanup to avoid any potential side effects on the error code. We found that in some cases, the exisiting approach was leading to a bogus error, `AWS_ERROR_PRIORITY_QUEUE_EMPTY`, when we failed in `aws_event_loop_run` while doing `pthread_create`.

### DIFF
--- a/source/event_loop.c
+++ b/source/event_loop.c
@@ -165,12 +165,16 @@ static struct aws_event_loop_group *s_event_loop_group_new(
 
     return el_group;
 
-on_error:
+on_error:;
+    /* cache the error code before potential side effects */
+    int cached_error_code = aws_last_error();
 
     aws_mem_release(alloc, usable_cpus);
     s_aws_event_loop_group_shutdown_sync(el_group);
     s_event_loop_group_thread_exit(el_group);
 
+    /* raise the cached error code */
+    aws_raise_error(cached_error_code);
     return NULL;
 }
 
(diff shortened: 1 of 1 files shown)


source/event_loop.c
```diff
@@ -165,12 +165,16 @@ static struct aws_event_loop_group *s_event_loop_group_new(
 
     return el_group;
 
-on_error:
+on_error:;
+    /* cache the error code to prevent any potential side effects */
+    int cached_error_code = aws_last_error();
 
     aws_mem_release(alloc, usable_cpus);
     s_aws_event_loop_group_shutdown_sync(el_group);
```
DmitriyMusatkin, line 173: is one of those throwing some new error because it cant clean up correctly? should we fix that as well?

source/event_loop.c
```diff
@@ -165,12 +165,16 @@ static struct aws_event_loop_group *s_event_loop_group_new(
 
     return el_group;
 
-on_error:
+on_error:;
```
graebm, line 168: `; _ ;`

graebm (approved):

source/event_loop.c
graebm: There's code in here that loops through a priority list until it raises the AWS_ERROR_PRIORITY_QUEUE_EMPTY, indicating it's empty.

So it's not an error-as-fail, it's an error-code-as-flow-control

PR #631, opened 2024-03-08
Fix sizeof channel negotiation task

quinnj:
Noticed while running a client program that invoked `aws_channel_setup_client_tls` and subsequently had heap corruption errors. When running with `ENABLE_SANITIZERS=ON`, it produced a report that the `negotiation_task` was only being allocated with 64 bytes where as `aws_channel_task_init` called `AWS_ZERO_STRUCT` where it was trying to set 104 bytes to 0, thus over-running the original memory region.


waahm7 (approved):

waahm7:
Released in https://github.com/awslabs/aws-c-io/releases/tag/v0.14.6.

Issue #521, opened 2022-10-18
Calling setenv during init is unsafe for multithreaded programs


johnwheffner:
We have an application that runs on Linux and uses the AWS C++ SDK.  It's a multithreaded application, and for reasons it initializes the AWS SDK (and transitively the aws-c-io library) after there are already a number of threads executing.  Some of these threads may at times call `getenv`.  We get crashes in these `getenv` calls if they occur concurrently with aws-c-io init, which calls `setenv` in `aws_tls_init_static_state`.  According to the [glibc docs](https://www.gnu.org/software/libc/manual/html_node/Environment-Access.html), "Modifications of environment variables are not allowed in multi-threaded programs."

It would be better to avoid the `setenv` call.

jmklix:
Closing. This was fixed in this PR: https://github.com/awslabs/aws-c-io/pull/634

PR #634, opened 2024-03-25
Remove setenv

DmitriyMusatkin:
*Issue #:* https://github.com/awslabs/aws-c-io/issues/521


source/s2n/s2n_tls_channel_handler.c
```diff
@@ -196,7 +229,8 @@ void aws_tls_init_static_state(struct aws_allocator *alloc) {
     }
 
     if (!s_s2n_initialized_externally) {
-        setenv("S2N_DONT_MLOCK", "1", 1);
+        //setenv("S2N_DONT_MLOCK", "1", 1);
```
graebm, line 232: trivial: don't commit commented-out code

source/s2n/s2n_tls_channel_handler.c
```diff
@@ -196,7 +229,8 @@ void aws_tls_init_static_state(struct aws_allocator *alloc) {
...
+        s_override_s2n_mem_functions(alloc);
```
graebm, line 233: debatable: why bother making this a helper function?

source/s2n/s2n_tls_channel_handler.c
```diff
@@ -174,6 +174,39 @@ AWS_IO_API const char *aws_determine_default_pki_ca_file(void) {
...
+    (void)size;
+    aws_mem_release(s_library_allocator, ptr);
+    return S2N_SUCCESS;
+}
+
+static void s_override_s2n_mem_functions(struct aws_allocator *alloc) {
+    if (alloc) {
+        s_library_allocator = alloc;
+    } else {
+        s_library_allocator = aws_default_allocator();
```
graebm, line 204: I'm pretty sure we don't allow NULL to be passed into library-init functions. Just ASSERT on it, don't try to handle the error

source/s2n/s2n_tls_channel_handler.c
```diff
@@ -174,6 +174,39 @@ AWS_IO_API const char *aws_determine_default_pki_ca_file(void) {
...
+    }
+
+    s2n_mem_set_callbacks(s_s2n_mem_init, s_s2n_mem_cleanup, s_s2n_mem_malloc, s_s2n_mem_free);
```
graebm, line 207: this returns an error code, we should check it

source/s2n/s2n_tls_channel_handler.c
DmitriyMusatkin: both s3 and auth have their init written this way, so just copying them here. happy to switch to assert if we disallow NULL

graebm: aws-c-http would crash, aws-c-cal (on linux) would crash,

waahm7 (approved):

PR #641, opened 2024-05-22
Implement TLS1.3 for windows client side

alfred2g:


source/windows/secure_channel_tls_handler.c
```diff
@@ -1078,7 +1123,7 @@ static int s_do_application_data_decrypt(struct aws_channel_handler *handler) {
 
         SECURITY_STATUS status = DecryptMessage(&sc_handler->sec_handle, &buffer_desc, 0, NULL);
 
-        if (status == SEC_E_OK) {
+        if (status == SEC_E_OK || status == SEC_I_RENEGOTIATE || status == SEC_I_CONTEXT_EXPIRED) {
```
bretambrose, line 1126: I don't know enough to validate this change but given that we have an existing else-if clause for  content expired below here that is unchanged (and now skipped), something feels really off.

source/windows/secure_channel_tls_handler.c
```diff
@@ -1142,9 +1192,95 @@ static int s_do_application_data_decrypt(struct aws_channel_handler *handler) {
             struct aws_channel_slot *slot = handler->slot;
             aws_channel_shutdown(slot->channel, AWS_OP_SUCCESS);
             error = AWS_OP_SUCCESS;
+        }
+        if (status == SEC_I_RENEGOTIATE) {
```
bretambrose, line 1196: Similar to the existing statues, can we get a brief explanation of what this means?

In fact the whole giant block desperately needs explanatory comments.

source/windows/secure_channel_tls_handler.c
```diff
@@ -1142,9 +1192,95 @@ static int s_do_application_data_decrypt(struct aws_channel_handler *handler) {
             struct aws_channel_slot *slot = handler->slot;
             aws_channel_shutdown(slot->channel, AWS_OP_SUCCESS);
             error = AWS_OP_SUCCESS;
+        }
+        if (status == SEC_I_RENEGOTIATE) {
+            AWS_LOGF_TRACE(
+                AWS_LS_IO_TLS, "id=%p: Renegotiation received. SECURITY_STATUS is %d.", (void *)handler, (int)status);
+            /* if we are the client */
```
bretambrose, line 1199: Is this an accurate comment?

source/windows/secure_channel_tls_handler.c
```diff
@@ -1142,9 +1192,95 @@ static int s_do_application_data_decrypt(struct aws_channel_handler *handler) {
...
+                sc_handler->ctx_req,
+                0,
+                0,
+                &input_bufs_desc,
+                0,
+                NULL,
+                &output_buffers_desc,
+                &sc_handler->ctx_ret_flags,
+                NULL);
+            error = status;
```
bretambrose, line 1260: This isn't a semantically valid assignment.

source/windows/secure_channel_tls_handler.c
```diff
@@ -1142,9 +1192,95 @@ static int s_do_application_data_decrypt(struct aws_channel_handler *handler) {
...
+            AWS_LOGF_ERROR(AWS_LS_IO_TLS, "id=%p: Error renegotiation happened. status %lu", (void *)handler, status);
```
bretambrose, line 1261: This seems incorrect, we havent checked the status code yet

source/windows/secure_channel_tls_handler.c
```diff
@@ -1812,20 +1916,199 @@ static struct aws_channel_handler *s_tls_handler_new(
...
+                break;
+            case AWS_IO_TLS_VER_SYS_DEFAULTS:
+                grbitEnabledProtocols = 0;
+                break;
+        }
+    }
+    return grbitEnabledProtocols;
+}
+
+static struct aws_channel_handler *s_tls_handler_new_win10_plus(
```
bretambrose, line 1968: I'd rather these were named based on what we're checking for semantically, ie the win32 feature and not just a windows version.

source/windows/secure_channel_tls_handler.c
```diff
@@ -1812,20 +1916,199 @@ static struct aws_channel_handler *s_tls_handler_new(
...
+
+    TLS_PARAMETERS tls_params = {0};
+    if (sc_ctx->schannel_creds.enabledProtocols == 0) {
+        tls_params.grbitDisabledProtocols = 0;
+    } else {
+        tls_params.grbitDisabledProtocols = ~(sc_ctx->schannel_creds.enabledProtocols);
+    }
+    credentials.pTlsParameters = &tls_params;
+    credentials.cTlsParameters = 1;
+    credentials.dwSessionLifespan = 0; /* default 10 hours */
```
bretambrose, line 1993: Is this a bound on 1.3 session resumption?  Hopefully its not a bound on connection lifetime.

source/windows/secure_channel_tls_handler.c
```diff
@@ -1960,25 +2201,24 @@ struct aws_tls_ctx *s_ctx_new(
...
+                   SCH_CRED_NO_SERVERNAME_CHECK | SCH_CRED_MANUAL_CRED_VALIDATION;
     } else if (is_client_mode) {
-        secure_channel_ctx->credentials.dwFlags |= SCH_CRED_REVOCATION_CHECK_CHAIN | SCH_CRED_IGNORE_REVOCATION_OFFLINE;
+        dwFlags |= SCH_CRED_REVOCATION_CHECK_CHAIN | SCH_CRED_IGNORE_REVOCATION_OFFLINE;
     }
 
     /* if someone wants to use broken algorithms like rc4/md5/des they'll need to ask for a special control */
-    secure_channel_ctx->credentials.dwFlags |= SCH_USE_STRONG_CRYPTO;
+    dwFlags |= SCH_USE_STRONG_CRYPTO;
+    dwFlags |= SCH_CRED_NO_DEFAULT_CREDS;
```
bretambrose, line 2221: This is new.  What's the reasoning behind it?

bretambrose (commented):
Also, there aren't any new tests which seems surprising.

source/windows/secure_channel_tls_handler.c
alfred2g: The number of milliseconds that Schannel keeps the session in its session cache. After this time has passed, any new connections between the client and the server require a new Schannel session. Set the value of this member to zero to use the default value of 36000000 milliseconds (ten hours).

alfred2g: it is similar to what we were using before.. with schannel_cred

source/windows/secure_channel_tls_handler.c
alfred2g: Prevent Schannel from attempting to automatically supply a certificate chain for client authentication.

i remember it didn't work without it at the beginning i will try again without it and see how it goes

alfred2g: ok it works, i updated it

source/windows/secure_channel_tls_handler.c
```diff
@@ -1094,24 +1144,29 @@ static int s_do_application_data_decrypt(struct aws_channel_handler *handler) {
                 int append_failed = aws_byte_buf_append(&sc_handler->buffered_read_out_data_buf, &to_append);
                 AWS_ASSERT(!append_failed);
                 (void)append_failed;
-
                 /* if we have extra we have to move the pointer and do another Decrypt operation. */
```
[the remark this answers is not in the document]
alfred2g: moved it below

source/windows/secure_channel_tls_handler.c
```diff
@@ -1142,9 +1197,101 @@ static int s_do_application_data_decrypt(struct aws_channel_handler *handler) {
...
+                NULL);
+            if (status == SEC_E_OK) {
+                error = AWS_OP_SUCCESS;
+                if (input_buffers2[1].BufferType == SECBUFFER_EXTRA && input_buffers2[1].cbBuffer > 0 &&
+                    sc_handler->buffered_read_in_data_buf.len > input_buffers2[1].cbBuffer) {
+                    memmove(
+                        sc_handler->buffered_read_in_data_buf.buffer,
+                        (sc_handler->buffered_read_in_data_buf.buffer + sc_handler->buffered_read_in_data_buf.len) -
+                            input_buffers2[1].cbBuffer,
+                        input_buffers2[1].cbBuffer);
```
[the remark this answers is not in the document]
alfred2g: we pass it again through decrypt to get sec_ok

source/windows/secure_channel_tls_handler.c
```diff
@@ -1142,9 +1197,101 @@ static int s_do_application_data_decrypt(struct aws_channel_handler *handler) {
...
+                &input_bufs_desc,
+                0,
+                NULL,
+                &output_buffers_desc,
+                &sc_handler->ctx_ret_flags,
+                NULL);
+            if (status == SEC_E_OK) {
+                error = AWS_OP_SUCCESS;
+                if (input_buffers2[1].BufferType == SECBUFFER_EXTRA && input_buffers2[1].cbBuffer > 0 &&
+                    sc_handler->buffered_read_in_data_buf.len > input_buffers2[1].cbBuffer) {
```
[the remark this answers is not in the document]
alfred2g: if this condition is false, the data is not needed, i've seen that case happen, it is fine

source/windows/secure_channel_tls_handler.c
```diff
@@ -1135,16 +1190,111 @@ static int s_do_application_data_decrypt(struct aws_channel_handler *handler) {
...
+                    memmove(
+                        sc_handler->buffered_read_in_data_buf.buffer,
+                        (sc_handler->buffered_read_in_data_buf.buffer + sc_handler->buffered_read_in_data_buf.len) -
+                            input_buffers2[1].cbBuffer,
+                        input_buffers2[1].cbBuffer);
+                    sc_handler->buffered_read_in_data_buf.len = input_buffers2[1].cbBuffer;
+                    sc_handler->read_extra = input_buffers2[1].cbBuffer;
+                    continue;
+                }
+                break;
```
sfod, line 1282: I can't figure out why we don't want to try to decrypt extra data when `sc_handler->buffered_read_in_data_buf.len == input_buffers2[1].cbBuffer`. I think, a comment explaining this would be useful.

source/windows/secure_channel_tls_handler.c
```diff
@@ -1812,20 +1933,223 @@ static struct aws_channel_handler *s_tls_handler_new(
...
+
+    return NULL;
+}
+
+static bool s_is_testing_deprecated_schannel_creds_defined() {
+    DWORD ret;
+    char buffer[10];
+
+    /* Used for testing: if defined to any value, we run the deprecarted SCHANNEL_CREDS on newer windows versions */
+    ret = GetEnvironmentVariable("TEST_DEPRECATED_SCHANNEL_CREDS", buffer, 10);
```
sfod, line 2116: Reading environment variable in production might not meet the security requirements. I think this moment should be verified.
But even apart the security concerns, the approach with checking environment variable in production code to determine what to do doesn't feel right.

source/windows/secure_channel_tls_handler.c
```diff
@@ -120,6 +132,42 @@ static size_t s_message_overhead(struct aws_channel_handler *handler) {
     return sc_handler->stream_sizes.cbTrailer + sc_handler->stream_sizes.cbHeader;
 }
 
+static bool s_is_windows_equal_or_above_10(void) {
```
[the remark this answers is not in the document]
sfod: Did you decide to keep the old name? Asking because you were going to rename the function.

source/windows/secure_channel_tls_handler.c
bretambrose: A better approach might be to read it in test code and use a private API to set a flag that changes behavior.

source/windows/secure_channel_tls_handler.c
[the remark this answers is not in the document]
sbSteveK: This function is still not doing a functionality check but is checking the version of windows being used. I have changed its name to the specific version being checked but don't see any reason to change the internal execution.

source/windows/secure_channel_tls_handler.c
```diff
@@ -1078,7 +1130,7 @@ static int s_do_application_data_decrypt(struct aws_channel_handler *handler) {
 
         SECURITY_STATUS status = DecryptMessage(&sc_handler->sec_handle, &buffer_desc, 0, NULL);
 
-        if (status == SEC_E_OK) {
+        if (status == SEC_E_OK || status == SEC_I_RENEGOTIATE) {
             error = AWS_OP_SUCCESS;
             /* if SECBUFFER_DATA is the buffer type of the second buffer, we have decrypted data to process.
                If SECBUFFER_DATA is the type for the fourth buffer we need to keep track of it so we can shift
```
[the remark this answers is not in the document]
sbSteveK: Yup. Changing the comment.

sfod:
Implemented in https://github.com/awslabs/aws-c-io/pull/732

PR #642, opened 2024-05-30
Fix bug where last few bytes on socket go unread

graebm:
**Issue:**

aws-c-s3 is occasionally seeing errors when the server sends an HTTP response with a `Connection: close` header (meaning it intends to close the connection after the response is sent). The server is sending the full response, then immediately hanging up. But the last few bytes of the response never make it to the HTTP client.

**Diagnosis:**

If a peer closed their socket immediately after the last few bytes were sent, our socket code wasn't always reading those last few bytes.

On Linux, if a socket has unread data AND the peer has closed their side, the event from epoll has 2 flags set: `EPOLLIN|EPOLLRDHUP`. This means "the socket is has data to read AND the other side is closed (or half-closed) and won't be sending any more data".

Our [socket handler code](https://github.com/awslabs/aws-c-io/blob/e762fd250589dfa98a9cce9c3ffca3414d99fdda/source/socket_channel_handler.c#L217-L225) *kinda* did the right thing by "attempting" to read from the socket before shutting down the channel. But if the downstream read window reached 0, that "attempt" wouldn't read all the data.

**Description of changes:**

The socket handler no longer shuts down the channel in response to an error event. Instead, the error event queues more reads to happen. And only when `read()` reports that the socket is finished (due to error or EOF), will the socket handler shut down the channel.


source/socket_channel_handler.c
```diff
@@ -212,17 +219,23 @@ static void s_on_readable_notification(struct aws_socket *socket, int error_code
...
+        AWS_LS_IO_SOCKET_HANDLER,
+        "id=%p: socket on-readable with error code %d(%s)",
+        (void *)socket_handler->slot->handler,
+        error_code,
+        aws_error_name(error_code));
+
+    /* Regardless of error code call read() until it reports error or EOF,
+     * so we can pick up data that was sent prior to the close.
+     * For example, if peer closes the socket immediately after sending the last
+     * bytes of data, the READABLE and HANGUP events arrive simultaneously.
```
TingDaoK, line 232: should we mention "as an example, the window reaches to zero"?

source/socket_channel_handler.c
```diff
@@ -212,17 +219,23 @@ static void s_on_readable_notification(struct aws_socket *socket, int error_code
...
+     * Another example, peer sends a TLS ALERT then immediately closes the socket.
+     * On some platforms, we'll never see the readable flag. So we want to make
      * sure we read the ALERT, otherwise, we'll end up telling the user that the channel shutdown because of a socket
      * closure, when in reality it was a TLS error */
+    (void)error_code;
     s_do_read(socket_handler);
-
-    if (error_code && !socket_handler->shutdown_in_progress) {
-        aws_channel_shutdown(socket_handler->slot->channel, error_code);
-    }
```
graebm, line 225: This is the only line that truly HAD to change to fix this bug

TingDaoK (approved):
LGTM

DmitriyMusatkin (approved):

PR #51, opened 2018-11-07
Channel refcount

graebm:
Allow users to delay deletion of a channel and its handlers by acquiring a hold on the channel. Deletion continues when the last hold is released.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #647, opened 2024-06-25
Socket option for binding to a network interface by name

waahm7:
*Description of changes:*
- Adds a socket option to bind to a particular network interface by name. This feature is currently only supported on Linux and MacOS. For Windows, we will need to add another option to bind to an ip address but we can support Windows later. 


include/aws/io/socket.h
```diff
@@ -43,6 +50,15 @@ struct aws_socket_options {
...
     bool keepalive;
+
+    /**
+     * (Optional)
+     * This property is used to bind the socket to a particular network interface by name, such as eth0 and ens32.
+     * If this is empty, the socket will not be bound to any interface and will use OS defaults. If the provided name
+     * is invalid, `aws_socket_init()` will error out with AWS_IO_SOCKET_INVALID_OPTIONS. This option is only
+     * supported on Linux and MacOS and will be ignored on other platforms.
+     */
+    char network_interface_name[AWS_NETWORK_INTERFACE_MAX_LEN];
```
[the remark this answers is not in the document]
waahm7: We copy socket_options by value. See https://github.com/awslabs/aws-c-io/pull/467#issuecomment-1888064444

tests/socket_test.c
```diff
@@ -804,17 +854,30 @@ static int s_test_outgoing_tcp_sock_error(struct aws_allocator *allocator, void
     struct aws_socket outgoing;
     ASSERT_SUCCESS(aws_socket_init(&outgoing, allocator, &options));
     /* tcp connect is non-blocking, it should return success, but the error callback will be invoked. */
-    ASSERT_SUCCESS(aws_socket_connect(&outgoing, &endpoint, event_loop, s_null_sock_connection, &args));
```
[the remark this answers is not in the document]
waahm7: I don't think so. If `aws_socket_connect` fails with an error, our contract should be that callback will not be triggered.

include/aws/io/socket.h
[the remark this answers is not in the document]
DmitriyMusatkin: hmm, is this a copy by value issue or more of a lifetime issue? where we use this struct way past the init function scope?

tests/socket_test.c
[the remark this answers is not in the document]
DmitriyMusatkin: im not necessarily talking about triggering a callback here. 
but if on freebsd we can get AWS_IO_SOCKET_CONNECTION_REFUSED right away instead of callback, should we document it and explicitly check here that last error was AWS_IO_SOCKET_CONNECTION_REFUSED?

include/aws/io/socket.h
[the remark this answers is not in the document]
waahm7: I haven't looked deeply, but off the top of my head, one example is that the connection manager needs this struct whenever it acquires a connection, so we make a copy at https://github.com/awslabs/aws-c-http/blob/main/source/connection_manager.c#L868. We will need to look through all the usages and make a deep copy/destroy if we want to put a bytecursor here. I felt it was not worth the effort considering IFNAMSIZ value is 16.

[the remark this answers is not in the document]
DmitriyMusatkin: agree. its diminishing returns.
having users do a strcpy into our buffer seems awkward, but whatever not a huge hill im willing to die on.

tests/socket_test.c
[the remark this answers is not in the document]
waahm7: On looking more deeply, I think this is a problem with the [O_NONBLOCK](https://github.com/awslabs/aws-c-io/blob/main/source/posix/socket.c#L154) flag rather than the connect timeout. FreeBSD doesn't seem to respect it. I have reverted the change here for now and will try to create another PR after looking more into it, or maybe trying SOCK_NONBLOCK instead of O_NONBLOCK.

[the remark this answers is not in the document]
waahm7: On FreeBSD, the connect is failing sync with error  51/202 Test  #51: outgoing_tcp_sock_error 
```
***FAILURE*** [DEBUG] [2024-06-26T17:55:17Z] [00000b9aa7412000] [Unknown] - id=0x8209262b0 fd=7: connect failed with error code 61.
```. error code 61 is ECONNREFUSED.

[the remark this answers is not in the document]
Comment (waahm7) [reply tests/socket_test.c]: Created https://github.com/awslabs/aws-c-io/pull/648.

File: source/posix/socket.c [hunk 3]
```diff
@@ -1254,6 +1254,69 @@ int aws_socket_set_options(struct aws_socket *socket, const struct aws_socket_op
...
+        AWS_LOGF_ERROR(
+            AWS_LS_IO_SOCKET,
+            "id=%p fd=%d: network_interface_name max length must be %d length and NULL terminated",
+            (void *)socket,
+            socket->io_handle.data.fd,
+            AWS_NETWORK_INTERFACE_MAX_LEN);
+        return aws_raise_error(AWS_IO_SOCKET_INVALID_OPTIONS);
+    }
+    if (network_interface_length != 0) {
+#if defined(__APPLE__)
```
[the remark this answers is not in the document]
waahm7: Thanks, updated.

DmitriyMusatkin (approved):

include/aws/io/socket.h
```diff
@@ -43,6 +50,15 @@ struct aws_socket_options {
      * lost. If zero OS defaults are used. On Windows, this option is meaningless until Windows 10 1703.*/
     uint16_t keep_alive_max_failed_probes;
     bool keepalive;
+
+    /**
+     * (Optional)
+     * This property is used to bind the socket to a particular network interface by name, such as eth0 and ens32.
+     * If this is empty, the socket will not be bound to any interface and will use OS defaults. If the provided name
+     * is invalid, `aws_socket_init()` will error out with AWS_IO_SOCKET_INVALID_OPTIONS. This option is only
+     * supported on Linux and MacOS and will be ignored on other platforms.
```
graebm, line 59: maybe AWS_ERROR_PLATFORM_NOT_SUPPORTED or AWS_IO_SOCKET_INVALID_OPTIONS instead of "ignored on other platforms"

seems better to give an error if someone is trying to use it on windows. Otherwise they'll think it works, but wonder why nothing is different, and spend a bunch of time chasing down what's going wrong

include/aws/io/socket.h
```diff
@@ -30,6 +30,13 @@ enum aws_socket_type {
     AWS_SOCKET_DGRAM,
 };
 
+#ifdef AWS_OS_WINDOWS
+#    define AWS_NETWORK_INTERFACE_MAX_LEN 16
+#else
+#    include <net/if.h>
+#    define AWS_NETWORK_INTERFACE_MAX_LEN IFNAMSIZ
```
graebm, line 37: `AWS_NETWORK_INTERFACE_NAME_MAX` instead of ~AWS_NETWORK_INTERFACE_MAX_LEN~. It's confusing to use "LEN" to mean a string size that includes the `\0`, when `strlen()` doesn't include the `\0` in it.

This aligns with some [POSIX](https://pubs.opengroup.org/onlinepubs/009695399/basedefs/limits.h.html) names like `PATH_MAX` and `NAME_MAX`

I'm sure there are examples of `MAX_LEN` in our codebase, but 🤷‍♀️

source/posix/socket.c
```diff
@@ -1254,6 +1254,69 @@ int aws_socket_set_options(struct aws_socket *socket, const struct aws_socket_op
...
+        return aws_raise_error(AWS_IO_SOCKET_INVALID_OPTIONS);
+    }
+    if (network_interface_length != 0) {
+#if defined(AWS_OS_APPLE)
+        uint network_interface_index = if_nametoindex(options->network_interface_name);
+        if (network_interface_index == 0) {
+            int errno_value = errno; /* Always cache errno before potential side-effect */
+            AWS_LOGF_ERROR(
+                AWS_LS_IO_SOCKET,
+                "id=%p fd=%d: network_interface_name (%s) was not found with errno %d.",
```
graebm, line 1274: trivial/debatable: If someone is going to do the homework to figure out what this errno value is, they'd probably like to know the function that failed too.
```suggestion
                "id=%p fd=%d: network_interface_name "%s" not found. if_nametoindex() failed with errno %d.",
```

source/posix/socket.c
```diff
@@ -1254,6 +1254,69 @@ int aws_socket_set_options(struct aws_socket *socket, const struct aws_socket_op
...
+            AWS_LS_IO_SOCKET,
+            "id=%p fd=%d: network_interface_name max length must be %d length and NULL terminated",
+            (void *)socket,
+            socket->io_handle.data.fd,
+            AWS_NETWORK_INTERFACE_MAX_LEN);
+        return aws_raise_error(AWS_IO_SOCKET_INVALID_OPTIONS);
+    }
+    if (network_interface_length != 0) {
+#if defined(AWS_OS_APPLE)
+        uint network_interface_index = if_nametoindex(options->network_interface_name);
```
graebm, line 1269: add comment explaining whey we use if_nametoindex() on Apple, but not Linux. I'm curious, and future generations will probably also be curious

source/posix/socket.c
```diff
@@ -1254,6 +1254,69 @@ int aws_socket_set_options(struct aws_socket *socket, const struct aws_socket_op
...
+        if (setsockopt(
+                socket->io_handle.data.fd,
+                SOL_SOCKET,
+                SO_BINDTODEVICE,
+                options->network_interface_name,
+                network_interface_length)) {
+            int errno_value = errno; /* Always cache errno before potential side-effect */
+            AWS_LOGF_ERROR(
+                AWS_LS_IO_SOCKET,
+                "id=%p fd=%d: setsockopt() for SO_BINDTODEVICE failed with errno %d.",
```
graebm, line 1306: trivial: print the interface name that failed
```suggestion
                "id=%p fd=%d: setsockopt() with SO_BINDTODEVICE for "%s" failed with errno %d.",
```

source/posix/socket.c
```diff
@@ -1254,6 +1254,69 @@ int aws_socket_set_options(struct aws_socket *socket, const struct aws_socket_op
...
+        if (setsockopt(
+                socket->io_handle.data.fd,
+                IPPROTO_IP,
+                IP_BOUND_IF,
+                &network_interface_index,
+                sizeof(network_interface_index))) {
+            int errno_value = errno; /* Always cache errno before potential side-effect */
+            AWS_LOGF_ERROR(
+                AWS_LS_IO_SOCKET,
+                "id=%p fd=%d: setsockopt() for IPROTO_IP failed with errno %d.",
```
graebm, line 1290: trivial: probably worth printing the interface name, to give a hint what it's about. Also, it's IPPROTO_IP not ~IPPROTO_IP~. But maybe better to print the more specific IP_BOUND_IF?
```suggestion
                "id=%p fd=%d: setsockopt() with IP_BOUND_IF for "%s" failed with errno %d.",
```

include/aws/io/socket.h
waahm7: I thought about it and thought it's better to keep it cross-platform. Otherwise customers will have to do something like this:
```
#ifdef __linux__
socket_options.interface_name = ...
#endif
```
With the ignoring it approach, they don't need to ifdef.

graebm: in what world is there a specific name in the user's cross-platform code? Any name is going to be extremely specific

include/aws/io/socket.h
```diff
@@ -30,6 +30,13 @@ enum aws_socket_type {
     AWS_SOCKET_DGRAM,
 };
 
+#ifdef AWS_OS_WINDOWS
+#    define AWS_NETWORK_INTERFACE_MAX_LEN 16
+#else
+#    include <net/if.h>
```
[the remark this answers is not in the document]
waahm7: > Just have some large-enough-for-all-platforms value here
We are only using IFNAMSIZ for all platforms and its value is 16. Maybe we can hard code 16 here, but then if they increase it, we will have to manually bump it up. We were also including a header at https://github.com/awslabs/aws-c-io/blob/main/include/aws/io/socket.h#L96. I think net/if.h is a standard header and should be fine, but I can hard code it if that's a more preferred approach.

include/aws/io/socket.h
waahm7: Thanks, updated.

include/aws/io/socket.h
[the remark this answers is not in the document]
graebm: I realize it's subjective, but in my experience a simple hard-coded value has less risks than including external headers from our public headers

source/posix/socket.c
```diff
@@ -1254,7 +1254,94 @@ int aws_socket_set_options(struct aws_socket *socket, const struct aws_socket_op
...
+                AWS_LS_IO_SOCKET,
+                "id=%p fd=%d: setsockopt() with IP_BOUND_IF for \"%s\" failed with errno %d.",
+                (void *)socket,
+                socket->io_handle.data.fd,
+                options->network_interface_name,
+                errno_value);
+            return aws_raise_error(AWS_IO_SOCKET_INVALID_OPTIONS);
+        }
+#elif defined(AWS_OS_LINUX)
+        if (setsockopt(
```
graebm, line 1321: do we need to worry about IPv4 vs IPv6 here?
SOL_SOCKET,SO_BINDTODEVICE works on both?

source/posix/socket.c
```diff
@@ -1254,7 +1254,94 @@ int aws_socket_set_options(struct aws_socket *socket, const struct aws_socket_op
...
+        AWS_LOGF_ERROR(
+            AWS_LS_IO_SOCKET,
+            "id=%p fd=%d: network_interface_name max length must be %d length and NULL terminated",
+            (void *)socket,
+            socket->io_handle.data.fd,
+            AWS_NETWORK_INTERFACE_NAME_MAX);
+        return aws_raise_error(AWS_IO_SOCKET_INVALID_OPTIONS);
+    }
+    if (network_interface_length != 0) {
+#if defined(AWS_OS_APPLE)
```
graebm, line 1268: instead of this:
```
if APPLE: use IP_BOUND_IF
elif: LINUX SO_BINDTODEVICE`
else: bork
```

seems more flexible (in a world with 10,000 Unix variants) to do something like this:
```
if defined(SO_BINDTODEVICE): use SO_BINDTODEVICE
elif defined(IP_BOUND_IF): use IP_BOUND_IF
else: bork
```

tests/socket_test.c
```diff
@@ -432,6 +432,56 @@ static int s_test_tcp_socket_communication(struct aws_allocator *allocator, void
 
 AWS_TEST_CASE(tcp_socket_communication, s_test_tcp_socket_communication)
 
+static int s_test_socket_with_bind_to_interface(struct aws_allocator *allocator, void *ctx) {
+    (void)ctx;
+#if !defined(AWS_OS_APPLE) && !defined(AWS_OS_LINUX)
```
graebm, line 437: since the ifdefs might get crazy, maybe just go through with the test and skip if it fails with PLATFORM_NOT_SUPPORTED

source/posix/socket.c
```diff
@@ -1254,7 +1254,94 @@ int aws_socket_set_options(struct aws_socket *socket, const struct aws_socket_op
...
+        /*
+         * Apple does not support SO_BINDTODEVICE and the alternative is IP_BOUND_IF which requires an index instead
+         * of name.
```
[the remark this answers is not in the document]
waahm7: Thanks, updated.

source/posix/socket.c
```diff
@@ -1254,7 +1254,94 @@ int aws_socket_set_options(struct aws_socket *socket, const struct aws_socket_op
...
+            AWS_LOGF_ERROR(
+                AWS_LS_IO_SOCKET,
+                "id=%p fd=%d: network_interface_name \"%s\" not found. if_nametoindex() failed with errno %d.",
+                (void *)socket,
+                socket->io_handle.data.fd,
+                options->network_interface_name,
+                errno_value);
+            return aws_raise_error(AWS_IO_SOCKET_INVALID_OPTIONS);
+        }
+        if (options->domain == AWS_SOCKET_IPV6) {
```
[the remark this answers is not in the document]
waahm7: VSOCK is Linux-only, so it doesn't matter. For local, the else branch fails with an error message. We can replace it with our own error, but I think it was cleaner to just let the OS handle it.

source/posix/socket.c
waahm7: Yeah, it's more flexible, but I think it's better to be exact about where we want to support this in a world of 10,000 Unix variants and this feature requiring Linux kernel >= 5.7 or root permissions, etc. If we make it more flexible and try to support all Unix variants instead of just Linux and macOS, it will increase our support scope, which I am not sure if we want to do unless someone asks for it.

source/posix/socket.c
waahm7: I have added a test for IPv6. The docs for SO_BINDTODEVICE confusingly mention, "Note that this only works for some socket types, particularly AF_INET sockets. It is not supported for packet sockets." AF_INET does mean IPv4, but I think it also includes IPv6 in this particular case, which is AF_INET6.

waahm7: I have also verified that the IPv6 test fails on Mac without an explicit branch.

tests/socket_test.c
waahm7: > since the ifdefs might get crazy
We can discuss the `ifdefs` in the office tomorrow.

source/posix/socket.c
waahm7: I have updated the ifdef's

source/posix/socket.c
```diff
@@ -1254,7 +1254,95 @@ int aws_socket_set_options(struct aws_socket *socket, const struct aws_socket_op
...
+                "id=%p fd=%d: setsockopt() with SO_BINDTODEVICE for \"%s\" failed with errno %d.",
+                (void *)socket,
+                socket->io_handle.data.fd,
+                options->network_interface_name,
+                errno_value);
+            return aws_raise_error(AWS_IO_SOCKET_INVALID_OPTIONS);
+        }
+#elif defined(IP_BOUND_IF)
+        /*
+         * Apple does not support SO_BINDTODEVICE and the alternative is IP_BOUND_IF which requires an index instead
```
graebm, line 1287:
```suggestion
         * If SO_BINDTODEVICE is not supported, the alternative is IP_BOUND_IF which requires an index instead
```

graebm (approved):

PR #648, opened 2024-06-27
Fix Freebsd CI

waahm7:


tests/socket_test.c
```diff
@@ -803,20 +801,33 @@ static int s_test_outgoing_tcp_sock_error(struct aws_allocator *allocator, void
...
     ASSERT_SUCCESS(aws_socket_init(&outgoing, allocator, &options));
-    /* tcp connect is non-blocking, it should return success, but the error callback will be invoked. */
-    ASSERT_SUCCESS(aws_socket_connect(&outgoing, &endpoint, event_loop, s_null_sock_connection, &args));
+    int result = aws_socket_connect(&outgoing, &endpoint, event_loop, s_null_sock_connection, &args);
+#ifdef __FreeBSD__
+    /**
+     * FreeBSD doesn't seem to respect the O_NONBLOCK or SOCK_NONBLOCK flag. It fails immediately when trying to
+     * connect to a socket which is not listening. Since this test does not aim to test for that, skip it in that
+     * case.
+     */
```
waahm7, line 810:
```suggestion
   /**
 * FreeBSD doesn't seem to respect the O_NONBLOCK or SOCK_NONBLOCK flag. It fails immediately when trying to
 * connect to a socket which is not listening. This is flaky and works sometimes, but we don't know why. Since this test does not aim to test for that, skip it in that
 * case.
 */
```

DmitriyMusatkin (approved):

PR #650, opened 2024-06-28
TLS deliver buffer data during shutdown

TingDaoK:
*Issue #, if available:*

- Very similar to https://github.com/awslabs/aws-c-io/pull/642
- In this case: 
  - the downstream window is limiting the data to be delivered from TLS, so TLS handler will cache the socket data.
  - While the window update task is pending for channel to pick up, socket reads EOF and start to schedule the shutdown task. So the shutdown task will run immediately after the window update happens.
  - The window update task opens the window again and schedule another read task like: https://github.com/awslabs/aws-c-io/blob/9072f86b8826f69e2698fb68eb3ae71a6960f0b4/source/s2n/s2n_tls_channel_handler.c#L1072-L1082, which will happen after the shutdown task
  - So, the cached data will be delivered to downstream after the shutdown happens and result in lost of data.

*Description of changes:*

- TLS handler now waits for buffered data to be delivered before it completes shutdown
- Allow the window update happens during shut down
- fix a bug in windows that we may endup overflow buffer when we try to read the whole frame for TLS to decrypt, while the downstream has no window left.

Notes:
- With this fix, if downstream of TLS has 0 window and channel shuts down, it may result in hanging.
  - If the shutdown in not initialized by the downstream, it's okay, since it's expected behavior for downstream when the window reaches to 0.
  - If the shutdown is initialized by downstream, and it wants to shutdown without caring the window size. It should increase the window after shutdown starts. With CRT libs, we only used TLS handler with the channel flow-control window from HTTP/1. https://github.com/awslabs/aws-c-http/pull/474


source/darwin/secure_transport_tls_channel_handler.c
```diff
@@ -567,6 +588,51 @@ static int s_handle_shutdown(
             "id=%p: shutting down read direction with error %d. Flushing queues.",
             (void *)handler,
             error_code);
+        if (!abort_immediately && secure_transport_handler->negotiation_finished &&
+            !aws_linked_list_empty(&secure_transport_handler->input_queue) && slot->adj_right) {
```
graebm, line 592: don't think you need to check `adj_right`?

`s_process_read_message()` deals with it just fine. If it's NULL, it uses SIZE_MAX as the window, and discards the aws_io_message after invoking the TLS on_data_read() callback

source/darwin/secure_transport_tls_channel_handler.c
```diff
@@ -669,14 +743,26 @@ static int s_process_read_message(
...
 
     return AWS_OP_SUCCESS;
+
+shutdown_channel:
+    if (secure_transport_handler->read_delayed_shutdown_task) {
+        if (shutdown_error_code != AWS_ERROR_SUCCESS) {
+            secure_transport_handler->read_delayed_shutdown_task->error = shutdown_error_code;
+        }
+        /* Schedule the task to continue the shutdown process. */
+        aws_channel_schedule_task_now(slot->channel, &secure_transport_handler->read_delayed_shutdown_task->task);
```
graebm, line 753: I don't understand why we're scheduling a task to call aws_channel_slot_on_handler_shutdown_complete()? Why not just call it synchronously? We've already sent any remaining messages at this point

TingDaoK (commented):
I only updated the mac os, if we agreed on the change, I'll propagate it to the other handlers

source/darwin/secure_transport_tls_channel_handler.c
```diff
@@ -669,14 +743,26 @@ static int s_process_read_message(
...
+    } else {
+        /* Starts the shutdown process */
+        aws_channel_shutdown(slot->channel, shutdown_error_code);
+    }
+    return AWS_OP_SUCCESS;
 }
 
 static void s_run_read(struct aws_channel_task *task, void *arg, aws_task_status status) {
     (void)task;
     if (status == AWS_TASK_STATUS_RUN_READY) {
```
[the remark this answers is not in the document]
TingDaoK: If we abort_immediately, there won't be any incoming message or pending data (we flushed those pending data during shutdown if `abort_immediately`), so, the read will still result in a no-op.

source/darwin/secure_transport_tls_channel_handler.c
```diff
@@ -556,22 +594,29 @@ static int s_handle_shutdown(
...
+        }
+        /* Flushing queues */
         while (!aws_linked_list_empty(&secure_transport_handler->input_queue)) {
             struct aws_linked_list_node *node = aws_linked_list_pop_front(&secure_transport_handler->input_queue);
             struct aws_io_message *message = AWS_CONTAINER_OF(node, struct aws_io_message, queueing_handle);
             aws_mem_release(message->allocator, message);
         }
+
+    } else {
+        /* Shutdown in write direction */
```
graebm, line 615: I ❤️ comments like this: explaining the `else`, after a very long `if`

source/darwin/secure_transport_tls_channel_handler.c
```diff
@@ -556,22 +594,29 @@ static int s_handle_shutdown(
...
+            AWS_LS_IO_TLS, "id=%p: shutting down read direction with error %d.", (void *)handler, error_code);
+        if (!abort_immediately && secure_transport_handler->negotiation_finished &&
+            !aws_linked_list_empty(&secure_transport_handler->input_queue) && slot->adj_right) {
+            s_initialize_read_delay_shutdown(handler, slot, error_code);
+            /* Early out, not complete the shutdown process for the handler until the handler processes the pending
+             * data. */
+            return AWS_OP_SUCCESS;
+        }
+        /* Flushing queues */
         while (!aws_linked_list_empty(&secure_transport_handler->input_queue)) {
```
graebm, line 608: There's a possible memory leak right now:
- if we're doing delayed shutdown
- and we end up calling `aws_channel_slot_on_handler_shutdown_complete()` with some data remaining in `input_queue` (maybe there was an error from `SSLRead()`, or maybe there's some garbage data leftover after the graceful close frame)

Anway, maybe move this back where it was, in the common code at the end of the function, so it gets hit on immediate-read-shutdown OR when we swing back around in the write direction.

graebm (commented):
Darwin code looks pretty good. Reviewing S2N-handler now...

source/darwin/secure_transport_tls_channel_handler.c
```diff
@@ -111,6 +111,9 @@ struct secure_transport_handler {
     bool negotiation_finished;
     bool verify_peer;
     bool read_task_pending;
+    bool delay_shutdown_scheduled;
+    bool delay_shutdown_error_code;
```
graebm, line 115: 😬 error_code should be an int

include/aws/io/private/tls_channel_handler_shared.h
```diff
@@ -19,6 +19,12 @@ struct aws_tls_channel_handler_shared {
     struct aws_crt_statistics_tls stats;
 };
 
+struct aws_tls_delayed_shutdown_task {
```
graebm, line 22: all the new tasks you added are simply calling `aws_channel_slot_on_handler_shutdown_complete()` which you could have just called synchronously

you could probably just revert this, and have a simpler diff overall

source/s2n/s2n_tls_channel_handler.c
```diff
@@ -1019,6 +1076,50 @@ static int s_s2n_handler_shutdown(
...
+                     * Nothing will kick off read in that case. */
+                    aws_channel_task_init(
+                        &s2n_handler->read_task, s_run_read, handler, "s2n_channel_handler_read_on_window_increment");
+                    aws_channel_schedule_task_now(slot->channel, &s2n_handler->read_task);
+                }
+            }
+            return AWS_OP_SUCCESS;
+        }
+
         while (!aws_linked_list_empty(&s2n_handler->input_queue)) {
```
graebm, line 1123: this has the same memory leak the darwin code currently has, where input_queue might not be cleaned if we delay read-shutdown, but then something goes wrong and we don't process everything in the queue

source/s2n/s2n_tls_channel_handler.c
```diff
@@ -1019,6 +1076,50 @@ static int s_s2n_handler_shutdown(
             s2n_handler->state = NEGOTIATION_FAILED;
         }
 
+        if (!abort_immediately && s2n_handler->state == NEGOTIATION_SUCCEEDED &&
+            !aws_linked_list_empty(&s2n_handler->input_queue) && slot->adj_right) {
+            /**
+             * In case of if we have any queued data in the handler after negotiation and we start to shutdown,
+             * make sure we pass those data down the pipeline before we complete the shutdown.
+             */
+            size_t downstream_size = aws_channel_slot_downstream_read_window(slot);
```
graebm, line 1085: trivial: I liked how you moved this to a little utility function in the darwin code, to keep the if/else's simpler

source/s2n/s2n_tls_channel_handler.c
```diff
@@ -622,6 +623,19 @@ static int s_s2n_handler_process_read_message(
         (unsigned long long)downstream_window - processed);
 
     return AWS_OP_SUCCESS;
+
+shutdown_channel:
+    if (s2n_handler->read_delayed_shutdown_task) {
+        if (shutdown_error_code != AWS_ERROR_SUCCESS) {
+            s2n_handler->read_delayed_shutdown_task->error = shutdown_error_code;
```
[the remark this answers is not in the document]
graebm: whatever we end up doing, make sure we do the same in darwin/windows

source/darwin/secure_transport_tls_channel_handler.c
```diff
@@ -111,6 +111,9 @@ struct secure_transport_handler {
...
+    bool shutdown_in_progress;
```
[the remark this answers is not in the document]
graebm: and error_code should be an int

source/s2n/s2n_tls_channel_handler.c
TingDaoK: yeah, I haven't made the change we agreed for drawin to s2n yet... should notify you earlier😣

source/s2n/s2n_tls_channel_handler.c
```diff
@@ -577,15 +575,18 @@ static int s_s2n_handler_process_read_message(
...
         }
 
         if (read < 0) {
             aws_mem_release(outgoing_read_message->allocator, outgoing_read_message);
 
             /* the socket blocked so exit from the loop */
             if (s2n_error_get_type(s2n_errno) == S2N_ERR_T_BLOCKED) {
+                if (s2n_handler->read_delayed_shutdown_task) {
+                    /* Propagate the shutdown as we blocked now. */
+                    goto shutdown_channel;
```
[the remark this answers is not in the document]
TingDaoK: I thought it would just be `AWS_IO_SOCKET_CLOSED`, but changed my mind to pass whatever was passed-in during shutdown handler callback.

source/channel.c
```diff
@@ -828,7 +828,7 @@ static void s_window_update_task(struct aws_channel_task *channel_task, void *ar
 
     channel->window_update_scheduled = false;
 
-    if (status == AWS_TASK_STATUS_RUN_READY && channel->channel_state < AWS_CHANNEL_SHUTTING_DOWN) {
+    if (status == AWS_TASK_STATUS_RUN_READY && channel->channel_state < AWS_CHANNEL_SHUT_DOWN) {
```
graebm, line 831: this change needs to happen, but it opens up a can of worms where a handler might get a window-increment callback after it's finished shutting down in the read-dir. So it might naively try to read again and send more read-messages, which is a crime after shutdown.

Can you look into this?

ughghghgh it makes me want to rewrite the channel API so that things like this are impossible. Where the channel tracks basic state information about each handler and won't allow stuff like this (don't alert a handler about window-update if it's already completed shutdown-in-read-dir). So every handler doesn't need to be so paranoid and defensive to be "correct". We don't have time for now right now, this fix needs to get out. But if there are followup issues...

source/s2n/s2n_tls_channel_handler.c
```diff
@@ -622,6 +631,15 @@ static int s_s2n_handler_process_read_message(
         (unsigned long long)downstream_window - processed);
 
     return AWS_OP_SUCCESS;
+
+shutdown_channel:
+    if (s2n_handler->delay_shutdown_scheduled) {
+        aws_channel_slot_on_handler_shutdown_complete(slot, AWS_CHANNEL_DIR_READ, shutdown_error_code, false);
```
graebm, line 637: if the channel is shutting down with a non-zero error-code, but s2n_recv() returns 0, we'll end up passing along 0 as the error-code.

seems like, if `s2n_handler->delayed_shutdown_error_code != 0`, we should prefer to pass along THAT error-code

source/windows/secure_channel_tls_handler.c
```diff
@@ -1222,14 +1222,21 @@ static int s_process_pending_output_messages(struct aws_channel_handler *handler
...
+        return ret;
+    }
+
+done:
+    if (sc_handler->read_state == AWS_TLS_HANDLER_READ_SHUTTING_DOWN) {
+        sc_handler->read_state = AWS_TLS_HANDLER_READ_SHUT_DOWN_COMPLETE;
         /* Continue the shutdown process delayed before. */
         aws_channel_slot_on_handler_shutdown_complete(
-            sc_handler->slot, AWS_CHANNEL_DIR_READ, sc_handler->delay_shutdown_error_code, false);
+            sc_handler->slot, AWS_CHANNEL_DIR_READ, sc_handler->shutdown_error_code, false);
```
graebm, line 1236: if a local error happens, and `sc_handler->shutdown_error_code == 0`, we should pass along the local error code instead

maybe do a local `int shutdown_error_code = 0;` like in s2n/darwin

source/windows/secure_channel_tls_handler.c
```diff
@@ -1252,7 +1259,10 @@ static int s_process_read_message(
     struct aws_io_message *message) {
 
     struct secure_channel_handler *sc_handler = handler->impl;
-    if (sc_handler->read_shutdown_completed) {
+    if (sc_handler->read_state == AWS_TLS_HANDLER_READ_SHUT_DOWN_COMPLETE) {
+        if (message) {
+            aws_mem_release(message->allocator, message);
+        }
         return AWS_OP_SUCCESS;
     }
```
graebm, line 1268: down below, on line 1305, which is like:
```
/* throw this one as a protocol error. */
aws_raise_error(AWS_IO_TLS_ERROR_WRITE_FAILURE);
```

I think there should be a `break;` afterwards?

source/windows/secure_channel_tls_handler.c
```diff
@@ -1283,15 +1310,23 @@ static int s_process_read_message(
...
+                    if (downstream_window > 0) {
+                        /* prevent a deadlock due to downstream handlers wanting more data, but we have an incomplete
+                           record, and the amount they're requesting is less than the size of a tls record. */
+                        size_t window_size = slot->window_size;
+                        if (!window_size &&
+                            aws_channel_slot_increment_read_window(slot, sc_handler->estimated_incomplete_size)) {
+                            err = AWS_OP_ERR;
+                        } else {
+                            sc_handler->estimated_incomplete_size = 0;
+                            err = AWS_OP_SUCCESS;
```
graebm, line 1328: by moving this line under a new `if (downstream_window > 0)` branch, the `err` doesn't get reset to 0

OMG THIS LOOP IS BRAIN MELTING

graebm (commented):
@TingDaoK I pushed my own take on the Windows error-handling. Can you take a look at it?

source/windows/secure_channel_tls_handler.c
```diff
@@ -1264,131 +1264,92 @@ static int s_process_read_message(
...
+                break;
+            }
+        }
 
-            size_t available_buffer_space =
-                sc_handler->buffered_read_in_data_buf.capacity - sc_handler->buffered_read_in_data_buf.len;
-            size_t available_message_len = message_cursor.len;
-            size_t amount_to_move_to_buffer =
-                available_buffer_space > available_message_len ? available_message_len : available_buffer_space;
+        /* if any data was decrypted, try to send it upstream */
```
TingDaoK, line 1305:
```suggestion
        /* if any data was decrypted, try to send it downstream */
```
Right?

source/windows/secure_channel_tls_handler.c
```diff
@@ -1264,131 +1264,92 @@ static int s_process_read_message(
...
-
-        aws_channel_shutdown(slot->channel, aws_last_error());
-        return AWS_OP_ERR;
     }
 
-    if (sc_handler->buffered_read_out_data_buf.len) {
-        if (s_process_pending_output_messages(handler)) {
-            return AWS_OP_ERR;
-        }
+    if (!err) {
```
TingDaoK, line 1346: trivial: maybe do 
```
if(err) {
   // handle error
}
```
will be more straightforward.

TingDaoK (commented):
looks good to me, thanks!

source/windows/secure_channel_tls_handler.c
graebm: I was just keeping the old way, but whatever I've already changed so much

graebm (approved):

PR #651, opened 2024-07-01
Update secure_channel_tls_handler.c

normanade:
Fix memory leak.

*Issue from [aws-sdk-cpp](https://github.com/aws/aws-sdk-cpp/issues/3021) if available: S3Client Uploading Leaks Memory on Windows caused by InitializeSecurityContext misuse*

*Description of changes: add free for InitializeSecurityContext*

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #652, opened 2024-07-03
Fix memory leaks in Windows TLS code

graebm:
**Issue:**

A C++ SDK user found a memory leak in the Windows TLS channel-handler code and submitted the following PR: https://github.com/awslabs/aws-c-io/pull/651

In evaluating this 2-line fix, I noticed that it sat in the middle of some (preexisting) sloppy error-handling.

**Description of changes:**
- Fix memory leak, as shown in @normanade's PR, by calling `FreeContextBuffer()`.
    - Change it so we **always** call `FreeContextBuffer()` after calling to `AcceptSecurityContext()`/`InitializeSecurityContextA()`. Previously we'd only free the buffer in certain circumstances, and were probably leaking.
- Remove (often sloppy) error-handling that tries to deal with allocation failures of `aws_io_message`. Officially declare that message allocations can't fail.
    - NOTE: Ever since [this change in 2021](https://github.com/awslabs/aws-c-common/pull/830), our allocators will never return `NULL`. They deal with OOM by immediately terminating the program. This was done because error-handling code for allocation failure makes everything so much more complex, and it's seldom tested. We tried for years to handle it. It's not worth the effort.
- Fix some error-handling that forgot to delete the `aws_io_message` after checking its capacity
- In the Windows TLS code, replace some `AWS_ASSERT()` about buffers being large enough with `AWS_FATAL_ASSERT()`. Similarly, add an `AWS_FATAL_ASSERT()` in one place that never checked or asserted before doing the copy.
    - I don't know enough about TLS or Windows APIs to say whether or not these deserve real error-handling. So I'm just preserving the status quo (but crashing instead of doing undefined behavior).

Co-authored-by: @normanade

### DIFF
--- a/source/windows/secure_channel_tls_handler.c
+++ b/source/windows/secure_channel_tls_handler.c
@@ -476,11 +476,6 @@ static int s_fillin_alpn_data(
     return AWS_OP_SUCCESS;
 }
 
-static int s_process_connection_state(struct aws_channel_handler *handler) {
-    struct secure_channel_handler *sc_handler = handler->impl;
-    return sc_handler->s_connection_state_fn(handler);
-}
-
 static int s_do_application_data_decrypt(struct aws_channel_handler *handler);
 
 static int s_do_server_side_negotiation_step_2(struct aws_channel_handler *handler);
@@ -492,6 +487,9 @@ static int s_do_server_side_negotiation_step_1(struct aws_channel_handler *handl
     AWS_LOGF_TRACE(AWS_LS_IO_TLS, "id=%p: server starting negotiation", (void *)handler);
 
     aws_on_drive_tls_negotiation(&sc_handler->shared_state);
+
+    /* set this, and goto cleanup, if function encounters an error */
+    int aws_error = 0;
 
     unsigned char alpn_buffer_data[128] = {0};
     SecBuffer input_bufs[] = {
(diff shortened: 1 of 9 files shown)


source/windows/secure_channel_tls_handler.c
```diff
@@ -1585,11 +1558,17 @@ static int s_handler_shutdown(
...
+                    aws_mem_release(outgoing_message->allocator, outgoing_message);
+                    FreeContextBuffer(output_buffer.pvBuffer);
+                    if (error_code == 0) {
+                        error_code = AWS_IO_TLS_ERROR_WRITE_FAILURE;
+                    }
+                    return aws_channel_slot_on_handler_shutdown_complete(slot, dir, error_code, true);
                 }
                 memcpy(outgoing_message->message_data.buffer, output_buffer.pvBuffer, output_buffer.cbBuffer);
                 outgoing_message->message_data.len = output_buffer.cbBuffer;
+                FreeContextBuffer(output_buffer.pvBuffer);
```
TingDaoK, line 1571: why we only free the buffer when `status == SEC_E_OK || status == SEC_I_CONTEXT_EXPIRED`?
The [doc](https://learn.microsoft.com/en-us/windows/win32/api/sspi/nf-sspi-initializesecuritycontexta) doesn't seem to be very clear about failure, but it does mention that it may contain alert information with failure, if i understand correctly
```
On output, if an alert is generated, this buffer contains information about that alert, and the function fails.
```

normanade: Should be a typo by @graebm . In my original commit #651 , FreeContextBuffer is called before return of function, no matter success or failure.

graebm: Thanks for pointing that out, both of you.

The existing code had several functions where `FreeContextBuffer()` wasn't being called in call cases after a [AcceptSecurityContext()](https://learn.microsoft.com/en-us/windows/win32/api/sspi/nf-sspi-acceptsecuritycontext) or [InitializeSecurityContextA()](https://learn.microsoft.com/en-us/windows/win32/api/sspi/nf-sspi-initializesecuritycontexta) call.

I changed the error-handling styles for these functions, so `FreeContextBuffer()` is **always** called, regardless of what gets returned from `AcceptSecurityContext()`/`InitializeSecurityContextA()`.

TingDaoK (approved):

Issue #655, opened 2024-07-15
[regression] Bulding socket.c fails on Amazon Linux 2023 (and other OSs)


gorloffslava:
### Describe the bug

Looks like we've already had this bug before: 
1. https://github.com/awslabs/aws-c-io/pull/506
2. https://github.com/awslabs/aws-c-io/pull/514

And looks like this commit reintroduces this issue: https://github.com/awslabs/aws-c-io/commit/d04508d113851f1bc15630d93490b2aa09676137#diff-bbb6031b15894efd8c3b6ae10be38d407f72ac6c844513199348725cf044782c

It relies on that `inttypes.h` is included transitvely via `net/if.h`, which isn't the case on many OSs, including Amazon Linux 2023 amon others.

### Expected Behavior

Compilation works

### Current Behavior

Compilation fails with an error:
```
0.539 CMake Deprecation Warning at CMakeLists.txt:2 (cmake_minimum_required):
0.539   Compatibility with CMake < 3.5 will be removed from a future version of
0.539   CMake.
0.539 
0.539   Update the VERSION argument <min> value or use a ...<max> suffix to tell
0.539   CMake that the project does not need compatibility with older versions.
0.539 
0.539 
0.625 -- The C compiler identification is Clang 19.0.0
0.664 -- Detecting C compiler ABI info
0.844 -- Detecting C compiler ABI info - done
0.858 -- Check for working C compiler: /opt/llvm_toolchain/bin/clang - skipped
0.858 -- Detecting C compile features
0.858 -- Detecting C compile features - done
0.868 -- Performing Test CMAKE_HAVE_LIBC_PTHREAD
1.026 -- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
1.027 -- Found Threads: TRUE
1.032 -- Found crypto: /usr/local/lib/libcrypto.so
1.032 -- LibCrypto Include Dir: /usr/local/include
1.032 -- LibCrypto Shared Lib:  /usr/local/lib/libcrypto.so
1.032 -- LibCrypto Static Lib:  /usr/local/lib/libcrypto.a
1.035 -- Performing Test HAS_FPIC_FLAG
1.186 -- Performing Test HAS_FPIC_FLAG - Success
1.186 -- Performing Test HAS_64BIT_FILE_OFFSET_BY_DEFAULT
1.340 -- Performing Test HAS_64BIT_FILE_OFFSET_BY_DEFAULT - Success
1.340 -- Looking for fseeko
[...]

PR #656, opened 2024-07-15
Fix import of inttype

waahm7:
*Issue #, if available:*
https://github.com/awslabs/aws-c-io/issues/655


TingDaoK (approved):
Can we add a CI for vsock?

waahm7:
> Can we add a CI for vsock?

We already have a test, but we never enabled it. We can look into enabling it in CI as a separate item, as customers are facing this issue.

PR #661, opened 2024-07-29
Grand dispatch queue

sbSteveK:
Integration of Apple's grand dispatch queue with event loop

The CI is failing for dispatch queue because of the apple network socket is not in. The socket related changes is in PR: https://github.com/awslabs/aws-c-io/pull/662


source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,496 @@
...
+
+// returns true if we should execute an iteration, false otherwise
+bool begin_iteration(struct scheduled_service_entry *entry) {
+    bool should_execute_iteration = false;
+    struct dispatch_loop *dispatch_loop = entry->loop->impl_data;
+
+    aws_mutex_lock(&dispatch_loop->synced_data.lock);
+
+    // someone else is already going, do nothing
+    if (dispatch_loop->synced_data.scheduling_state.is_executing_iteration) {
```
sfod, line 300: I think I'm missing something. `is_executing_iteration` is set to true only within `run_iteration`. Does this check mean that multiple `run_iteration` can run in parallel?

include/aws/io/event_loop.h
```diff
@@ -140,6 +144,21 @@ struct aws_event_loop_group {
...
 
+typedef struct aws_event_loop *(aws_new_system_event_loop_fn)(struct aws_allocator *alloc,
+                                                              const struct aws_event_loop_options *options);
+
+struct aws_event_loop_configuration {
+    enum aws_event_loop_style style;
+    aws_new_system_event_loop_fn *event_loop_new_fn;
+    const char *name;
+    bool is_default;
+};
```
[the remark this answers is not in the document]
xiazhvera: Yes, the `is_default` is not functional for now, and is just reserved for a later usage. Currently, we use pre-compile definitions to determine which event-loop we would like to use in aws-c-io. In the future, we would like to allow a runtime configuration to set the event loop, so that the user could make choice theirselves. (Probably after we make sure the dispatch queue is in a stable status.) Once that's there, as we would have multiple event loop implementation enabled, the `is_default` would be used to set the "default" configuration. 
Let me add the comments to explain the details there.

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,496 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+#ifdef AWS_USE_DISPATCH_QUEUE
```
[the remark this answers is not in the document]
xiazhvera: Unfortunitely, the trick part is that we use kqueue for MacOS, and Dispatch Queue for iOS. Therefore, for Apple Platforms, we probably would have both event loop enabled still.

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,496 @@
...
+    .unsubscribe_from_io_events = s_unsubscribe_from_io_events,
+    .free_io_event_resources = s_free_io_event_resources,
+    .is_on_callers_thread = s_is_on_callers_thread,
+};
+
+struct scheduled_service_entry {
+    struct aws_allocator *allocator;
+    uint64_t timestamp;
+    struct aws_linked_list_node node;
+    struct aws_event_loop *loop; // might eventually need to be ref-counted for cleanup?
```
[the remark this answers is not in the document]
xiazhvera: No, removed the comments. The underltying dispatch_loop is ref-counted instead.

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,496 @@
...
+    if (err) {
+        AWS_LOGF_ERROR(AWS_LS_IO_EVENT_LOOP, "id=%p: Initializing task scheduler failed", (void *)loop);
+        goto clean_up_dispatch;
+    }
+
+    aws_linked_list_init(&dispatch_loop->local_cross_thread_tasks);
+    aws_linked_list_init(&dispatch_loop->synced_data.scheduling_state.scheduled_services);
+    aws_linked_list_init(&dispatch_loop->synced_data.cross_thread_tasks);
+
+    dispatch_loop->wakeup_schedule_needed = true;
```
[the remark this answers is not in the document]
xiazhvera: Nice catch. This need to be cleaned up. 
The variable was used to avoid explicitly scheduling a thread iteration. Now we used the function `should_schedule_iteration() ` to determine if we should schedule iterations. This is no long in use.

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,496 @@
...
+    loop->synced_data.scheduling_state.is_executing_iteration = false;
+
+    // if there are any cross-thread tasks, reschedule an iteration for now
+    if (!aws_linked_list_empty(&loop->synced_data.cross_thread_tasks)) {
+        // added during service which means nothing was scheduled because is_executing_iteration was true
+        try_schedule_new_iteration(entry->loop, 0);
+    } else {
+        // no cross thread tasks, so check internal time-based scheduler
+        uint64_t next_task_time = 0;
+        /* we already know it has tasks, we just scheduled one. We just want the next run time. */
```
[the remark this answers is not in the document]
xiazhvera: Hmmm I'm not sure if I understand the question here. The comment was just explain why we called function `aws_task_scheduler_has_tasks`. Here we just tried to retrieve the "next_task_time". 

The iteration will execute all time expired tasks. For example, say we are running the iteration at timestamp 1000, then we will execute all tasks that the `task_run_time` < 1000. 
And at the end of each iteration, we will schedule the next iteration at the `next_task_time`.

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,496 @@
...
+
+      aws_task_scheduler_clean_up(&dispatch_loop->scheduler);
+
+      while (!aws_linked_list_empty(&dispatch_loop->synced_data.cross_thread_tasks)) {
+          struct aws_linked_list_node *node = aws_linked_list_pop_front(&dispatch_loop->synced_data.cross_thread_tasks);
+          struct aws_task *task = AWS_CONTAINER_OF(node, struct aws_task, node);
+          task->fn(task, task->arg, AWS_TASK_STATUS_CANCELED);
+      }
+
+      while (!aws_linked_list_empty(&dispatch_loop->local_cross_thread_tasks)) {
```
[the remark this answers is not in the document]
xiazhvera: `local_cross_thread_tasks` is used to process the `cross_thread_tasks` and the function `void s_schedule_task_common(...)` will insert tasks into `cross_thread_tasks`.

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,500 @@
...
+}
+
+static int s_run(struct aws_event_loop *event_loop) {
+    struct dispatch_loop *dispatch_loop = event_loop->impl_data;
+
+    aws_mutex_lock(&dispatch_loop->synced_data.lock);
+    if (dispatch_loop->synced_data.suspended) {
+        AWS_LOGF_INFO(AWS_LS_IO_EVENT_LOOP, "id=%p: Starting event-loop thread.", (void *)event_loop);
+        dispatch_resume(dispatch_loop->dispatch_queue);
+        dispatch_loop->synced_data.suspended = false;
```
bretambrose, line 267: Shouldn't this schedule a service on the off-chance that there were scheduled tasks at the time of suspension?

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,500 @@
...
+    dispatch_async_f(dispatch_loop->dispatch_queue, entry, run_iteration);
+}
+
+static void s_schedule_task_common(struct aws_event_loop *event_loop, struct aws_task *task, uint64_t run_at_nanos) {
+    struct dispatch_loop *dispatch_loop = event_loop->impl_data;
+
+    aws_mutex_lock(&dispatch_loop->synced_data.lock);
+    bool should_schedule = false;
+
+    bool is_empty = aws_linked_list_empty(&dispatch_loop->synced_data.cross_thread_tasks);
```
bretambrose, line 420: I feel like `was_empty` would be slightly better

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,500 @@
...
+        /* Timestamp 0 is used to denote "now" tasks */
+        if (task->timestamp == 0) {
+            aws_task_scheduler_schedule_now(&dispatch_loop->scheduler, task);
+        } else {
+            aws_task_scheduler_schedule_future(&dispatch_loop->scheduler, task, task->timestamp);
+        }
+    }
+
+    aws_mutex_lock(&dispatch_loop->synced_data.lock);
+    dispatch_loop->synced_data.current_thread_id = aws_thread_current_thread_id();
```
bretambrose, line 377: Why wouldn't this be set in begin_iteration?

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,500 @@
...
+    aws_mutex_unlock(&dispatch_loop->synced_data.lock);
+
+    // run all scheduled tasks
+    uint64_t now_ns = 0;
+    aws_event_loop_current_clock_time(event_loop, &now_ns);
+    aws_task_scheduler_run_all(&dispatch_loop->scheduler, now_ns);
+    aws_event_loop_register_tick_end(event_loop);
+
+    aws_mutex_lock(&dispatch_loop->synced_data.lock);
+    dispatch_loop->synced_data.is_executing = false;
```
bretambrose, line 388: Why not make a part of end_iteration?

source/socket.c
```diff
@@ -156,6 +156,11 @@ void aws_socket_endpoint_init_local_address_for_test(struct aws_socket_endpoint
     struct aws_byte_buf uuid_buf = aws_byte_buf_from_empty_array(uuid_str, sizeof(uuid_str));
     AWS_FATAL_ASSERT(aws_uuid_to_str(&uuid, &uuid_buf) == AWS_OP_SUCCESS);
 
+#if defined(WS_USE_APPLE_NETWORK_FRAMEWORK)
```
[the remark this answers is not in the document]
sbSteveK: Also, I think this should be "AWS_USE_APPLE_NETWORK_FRAMEWORK"

source/darwin/dispatch_queue.h
```diff
@@ -0,0 +1,66 @@
+#ifndef AWS_IO_PRIVATE_DISPATCH_QUEUE_H
```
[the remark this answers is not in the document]
xiazhvera: The data structures here will be required for SecItem related changes, so we shared them in the private header here.

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,500 @@
...
+    loop->impl_data = dispatch_loop;
+    loop->vtable = &s_vtable;
+
+    /** manually increament the thread count, so the library will wait for dispatch queue releasing */
+    aws_thread_increment_unjoined_count();
+
+    return loop;
+
+clean_up:
+    if (dispatch_loop) {
```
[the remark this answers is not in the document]
xiazhvera: I tried to simplify the implementation here, while we still need conditional check here to clean up the underlying apple dispatch queue here, as `s_dispatch_event_loop_destroy` is only responsible for clean up the aws-c-io event loop structure.

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,538 @@
...
+            }
+        }
+    }
+
+    aws_mutex_unlock(&contxt->lock);
+    s_scheduled_service_entry_destroy(entry);
+}
+
+// Iteration function that scheduled and executed by the Dispatch Queue API
+static void s_run_iteration(void *context) {
```
bretambrose, line 379: The locking here seems a little unnecessarily complex.  If you look at a single call to s_run_iteration:

```
lock(context)
unlock(context)
lock(context) // via begin_iteration, also this seems wrong as you're slurping cross-thread tasks which should be protected by lock(loop) instead
unlock(context) // via begin_iteration
lock(loop)
set thread id/execution
unlock(loop)
run all tasks
lock(loop)
executing = false
unlock(loop)
lock(context) // via end_iteration, again this appears wrong since you access lock(loop) protected data here
unlock(context)
```

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,538 @@
...
+    end_iteration(entry);
+}
+
+/**
+ * Checks if a new iteration task needs to be scheduled, given a target timestamp. If so, submits an iteration task to
+ * dispatch queue and registers the pending execution in the event loop's list of scheduled iterations.
+ *
+ * If timestamp==0, the function will always schedule a new iteration as long as the event loop is not suspended.
+ *
+ * The function should be wrapped with dispatch_loop->synced_task_data->lock
```
bretambrose, line 435: This comment does not match the reality.  This function is getting called with the context lock held not the loop lock.

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,538 @@
...
+    if (!dispatch_loop || dispatch_loop->synced_task_data.suspended)
+        return;
+    if (!s_should_schedule_iteration(
+            &dispatch_loop->synced_task_data.context->scheduling_state.scheduled_services, timestamp)) {
+        return;
+    }
+    struct scheduled_service_entry *entry = s_scheduled_service_entry_new(dispatch_loop_context, timestamp);
+    aws_linked_list_push_front(
+        &dispatch_loop->synced_task_data.context->scheduling_state.scheduled_services, &entry->node);
+    dispatch_async_f(dispatch_loop->dispatch_queue, entry, s_run_iteration);
```
bretambrose, line 448: This doesn't seem to take into account that timestamp might be significantly in the future.  I don't get any search hits for `dispatch_after`

source/darwin/dispatch_queue.h
```diff
@@ -0,0 +1,74 @@
...
+    enum aws_tls_versions minimum_version;
+    struct aws_string *alpn_list;
+    bool verify_peer;
+};
+
+struct dispatch_scheduling_state {
+    /**
+     * Let's us skip processing an iteration task if one is already in the middle of executing
+     */
+    bool will_schedule;
```
bretambrose, line 29: It is still not clear to me why this and is_executing are separate.

source/darwin/dispatch_queue_event_loop.c
xiazhvera: Updated to schedule a service on event loop resume.

source/darwin/dispatch_queue.h
```diff
@@ -0,0 +1,74 @@
...
+
+struct dispatch_loop {
+    struct aws_allocator *allocator;
+    dispatch_queue_t dispatch_queue;
+    struct aws_task_scheduler scheduler;
+    struct aws_linked_list local_cross_thread_tasks;
+    struct aws_event_loop *base_loop;
+
+    /* Apple dispatch queue uses the id string to identify the dispatch queue */
+    struct aws_string *dispatch_queue_id;
```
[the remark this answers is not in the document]
xiazhvera: Removed and logged.

source/event_loop.c
```diff
@@ -74,7 +87,7 @@ struct aws_event_loop *aws_event_loop_new_with_epoll(
  * retrieve the default type value.
  * If `aws_event_loop_override_default_type` has been called, return the override default type.
  */
-static enum aws_event_loop_type aws_event_loop_get_default_type(void) {
+enum aws_event_loop_type aws_event_loop_get_default_type(void) {
```
sbSteveK, line 90: This function returns a enum but the `#else` logs an error and the function does not return anything.

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,635 @@
...
+ * If timestamp==0, the function will always schedule a new iteration as long as the event loop is not suspended.
+ *
+ * The function should be wrapped with dispatch_loop->context->lock & dispatch_loop->synced_data.lock
+ */
+static void s_try_schedule_new_iteration(struct dispatch_loop_context *dispatch_loop_context, uint64_t timestamp) {
+    struct dispatch_loop *dispatch_loop = dispatch_loop_context->io_dispatch_loop;
+    if (!dispatch_loop || dispatch_loop->synced_data.suspended) {
+        return;
+    }
+    if (!s_should_schedule_iteration(&dispatch_loop_context->scheduling_state.scheduled_services, timestamp)) {
```
bretambrose, line 517: If we always call should_schedule inside try_schedule, then we can probably replace the 
```
if (should_schedule(...))
    try_schedule(...)
```
call  in end_iteration

source/darwin/dispatch_queue.h
```diff
@@ -0,0 +1,62 @@
...
+    CFArrayRef ca_cert;
+    enum aws_tls_versions minimum_version;
+    struct aws_string *alpn_list;
+    bool verify_peer;
+};
+
+struct dispatch_loop;
+struct dispatch_loop_context;
+
+struct dispatch_loop {
```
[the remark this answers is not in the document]
xiazhvera: We put it here as it will be required for secItem PR.

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,635 @@
...
+    if (!aws_linked_list_empty(&dispatch_loop->synced_data.cross_thread_tasks)) {
+        // added during service which means nothing was scheduled because will_schedule was true
+        s_lock_service_entries(contxt);
+        s_try_schedule_new_iteration(contxt, 0);
+        s_unlock_service_entries(contxt);
+    } else {
+        // no cross thread tasks, so check internal time-based scheduler
+        uint64_t next_task_time = 0;
+        /* we already know it has tasks, we just scheduled one. We just want the next run time. */
+        bool has_task = aws_task_scheduler_has_tasks(&dispatch_loop->scheduler, &next_task_time);
```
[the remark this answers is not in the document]
xiazhvera: I didn't get what you mean here. I assume you are asking what gona happened if `next_task_time` is later than 1 second from "now". The `s_try_schedule_new_iteration` function will handle the clamping if the next_task_time is later than 1 second.

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,635 @@
...
+
+static void s_release_dispatch_loop_context(struct dispatch_loop_context *contxt) {
+    aws_ref_count_release(&contxt->ref_count);
+}
+
+static void s_rlock_dispatch_loop_context(struct dispatch_loop_context *contxt) {
+    aws_rw_lock_rlock(&contxt->lock);
+}
+
+static void s_runlock_dispatch_loop_context(struct dispatch_loop_context *contxt) {
```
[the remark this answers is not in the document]
xiazhvera: I update the functions to return the lock result, while I'm not sure if the lock are going fail in these cases...?

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,686 @@
...
+
+/* Internal ref-counted dispatch loop context to processing Apple Dispatch Queue Resources */
+struct dispatch_loop_context {
+    /**
+     * The conetxt lock is a read-write lock used to protect dispatch_loop.
+     * The write lock will be acquired when we make changes to dispatch_loop. And the read lock will be acquired
+     * when we need verify if the dispatch_loop is alive. This makes sure that the dispatch_loop will not be destroyed
+     * from other thread while we are using it.
+     */
+    struct aws_rw_lock lock;
```
sbSteveK, line 107: Trivial: The comment above refers to this lock as the "conetxt lock". Spelling on "context lock" and also we should clarify by either calling it "The dispatch_loop_context's lock" or renaming the lock member here to "context_lock"

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,686 @@
...
+
+/* Setup a dispatch_queue with a scheduler. */
+struct aws_event_loop *aws_event_loop_new_with_dispatch_queue(
+    struct aws_allocator *alloc,
+    const struct aws_event_loop_options *options) {
+    AWS_PRECONDITION(options);
+    AWS_PRECONDITION(options->clock);
+
+    struct aws_event_loop *loop = aws_mem_calloc(alloc, 1, sizeof(struct aws_event_loop));
+    struct dispatch_loop *dispatch_loop = NULL;
```
sbSteveK, line 291: Trivial: This can be moved below to where `aws_mem_calloc` is assigning memory to it.

sbSteveK: Never mind, dispatch_loop needs to be available for the NULL check in clean_up

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,686 @@
...
+    context->allocator = alloc;
+    aws_mutex_init(&context->scheduling_state.services_lock);
+
+    if (aws_priority_queue_init_dynamic(
+            &context->scheduling_state.scheduled_services,
+            alloc,
+            DEFAULT_QUEUE_SIZE,
+            sizeof(struct scheduled_service_entry *),
+            &s_compare_timestamps)) {
+        goto clean_up;
```
sbSteveK, line 338: If this fails and we goto clean_up, `dispatch_loop_context *context` has not yet been assigned to dispatc_loop->context. This means when `s_dispatch_event_loop_destroy(loop)` checks whether it should decref `context` by doing a NULL check, it fails and leaks memory. We should assign `dispatch_loop->context` with this `context` immediately following its ref_count being initialized so it gets cleaned properly.

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,686 @@
...
+}
+
+static void s_try_schedule_new_iteration(struct dispatch_loop_context *loop, uint64_t timestamp);
+
+static int s_run(struct aws_event_loop *event_loop) {
+    struct dispatch_loop *dispatch_loop = event_loop->impl_data;
+
+    s_lock_cross_thread_data(dispatch_loop);
+    if (dispatch_loop->synced_data.suspended) {
+        AWS_LOGF_INFO(AWS_LS_IO_EVENT_LOOP, "id=%p: Starting event-loop thread.", (void *)event_loop);
```
sbSteveK, line 417: Since dispatch loops are created in a running state when we retrieve it from Apple's framework, we should add this log to the point where the dispatch loop is started in the `aws_event_loop_new_with_dispatch_queue` function. This one will never get called on a newly created and run dispatch loop but we want the log that one was created and Started for debugging purposes.

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,686 @@
...
+}
+
+static int s_stop(struct aws_event_loop *event_loop) {
+    struct dispatch_loop *dispatch_loop = event_loop->impl_data;
+
+    s_lock_cross_thread_data(dispatch_loop);
+    if (!dispatch_loop->synced_data.suspended) {
+        dispatch_loop->synced_data.suspended = true;
+        AWS_LOGF_INFO(AWS_LS_IO_EVENT_LOOP, "id=%p: Stopping event-loop thread.", (void *)event_loop);
+        /* Suspend will increase the dispatch reference count. It is required to call resume before
```
sbSteveK, line 438: Trivial: resume -> run()

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,686 @@
...
+        s_try_schedule_new_iteration(dispatch_loop->context, 0);
+        s_unlock_service_entries(dispatch_loop->context);
+        s_runlock_dispatch_loop_context(dispatch_loop->context);
+    }
+    s_unlock_cross_thread_data(dispatch_loop);
+
+    return AWS_OP_SUCCESS;
+}
+
+static int s_stop(struct aws_event_loop *event_loop) {
```
sbSteveK, line 431: Dispatch loop shouldn't ever be put into a stop state (dispatch_suspend). If this is called, we should either ignore it, or return an error (platform not supported maybe?) and not mess with any complexity we need to add to support a thing that shouldn't be happening.

s_stop and s_run can be emptied, we will no longer need to check that the dispatch loop is in a non-suspended state before attempting to destroy it, and we can clean up the dispatch_loop by removing the bool suspended from synced_data. Let's only add complexity where it's required.

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,686 @@
...
+
+/**
+ * Checks if a new iteration task needs to be scheduled, given a target timestamp. If so, submits an iteration task to
+ * dispatch queue and registers the pending execution in the event loop's list of scheduled_services.
+ *
+ * If timestamp==0, the function will always schedule a new iteration as long as the event loop is not suspended.
+ *
+ * The function should be wrapped with the following locks:
+ *      dispatch_loop->context->lock: To retain the dispatch loop
+ *      dispatch_loop->synced_data.lock : To verify if the dispatch loop is suspended
```
sbSteveK, line 566: If we are removing the ability to suspend a dispatch queue, this is no longer a necessary comment.

source/darwin/dispatch_queue_event_loop_private.h
```diff
@@ -0,0 +1,55 @@
...
+     * Internal ref-counted dispatch loop context to processing Apple Dispatch Queue Resources.
+     * The context keep track of the live status of the dispatch loop. Dispatch queue should be
+     * nulled out in context when it is cleaned up.
+     */
+    struct dispatch_loop_context *context;
+
+    /* Synced data handle cross thread tasks and events, and event loop operations*/
+    struct {
+        /**
+         * The lock is used to protect synced_data across the threads. It should be acquired whenever we touched the
```
sbSteveK, line 33: Trivial: "It should be acquired whenever data in the synced_data struct is accessed or modified." ?

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,698 @@
...
+ *      dispatch_loop_context->scheduling_state->services_lock: To modify the scheduled_services list
+ */
+static void s_try_schedule_new_iteration(struct dispatch_loop_context *dispatch_loop_context, uint64_t timestamp) {
+    struct dispatch_loop *dispatch_loop = dispatch_loop_context->io_dispatch_loop;
+    if (!dispatch_loop || dispatch_loop->synced_data.suspended) {
```
sbSteveK, line 571: Not an issue if we remove suspended but if we are accessing synced_data.suspended, we should probably do so while holding the synced_data lock.

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,698 @@
...
+    s_release_dispatch_loop_context(dispatch_queue_context);
+}
+
+/**
+ * Checks if a new iteration task needs to be scheduled, given a target timestamp. If so, submits an iteration task to
+ * dispatch queue and registers the pending execution in the event loop's list of scheduled_services.
+ *
+ * If timestamp==0, the function will always schedule a new iteration as long as the event loop is not suspended.
+ *
+ * The function should be wrapped with the following locks:
```
sbSteveK, line 564: Trivial: "This function"

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,698 @@
...
+/**
+ * DISPATCH QUEUE
+ *
+ * Event loop is responsible for processing events and tasks by launching an execution loop on a single thread. Each
+ * iteration of this loop performs three primary jobs:
+ * 1. Process I/O events.
+ * 2. Process cross-thread tasks.
+ * 3. Execute all runnable tasks.
+ *
+ * Apple Dispatch queues are FIFO queues to which the application can submit tasks in the form of block objects, and the
```
sbSteveK, line 62: We should state that we create a serial dispatch queue and not that Apple Dispatch queues are FIFO. Concurrent Apple dispatch queues can be created so we want to be clear here what we are doing.

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,698 @@
...
+ * block objects will be executed on a system defined thread pool. Instead of executing the loop on a single thread, we
+ * tried to recurrently run a single iteration of the execution loop as a dispatch queue block object.
```
sbSteveK, line 64: Trivial: remove "tried to". I think we simply are doing it here.

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,698 @@
...
+    entry->dispatch_queue_context = s_acquire_dispatch_loop_context(context);
+    aws_priority_queue_node_init(&entry->priority_queue_node);
+
+    return entry;
+}
+
+/**
+ * The function should be wrapped around scheduling_status->lock
+ */
+static void s_scheduled_service_entry_destroy(
```
sbSteveK, line 193: This is only called as part of cleanup on `s_run_iteration`. Can we remove this function and add its execution code into s_run_iteration. 

If we keep the function, we should pull in the locking and unlocking of the scheduling_status->lock  into this function itself to keep the other end cleaner as well as update the comment to explain what this function is doing in the comment description above.

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,698 @@
...
+}
+
+// Iteration function that scheduled and executed by the Dispatch Queue API
+static void s_run_iteration(void *context) {
+    struct scheduled_service_entry *entry = context;
+    struct dispatch_loop_context *dispatch_queue_context = entry->dispatch_queue_context;
+    s_acquire_dispatch_loop_context(dispatch_queue_context);
+    s_rlock_dispatch_loop_context(dispatch_queue_context);
+
+    if (!begin_iteration(entry)) {
```
sbSteveK, line 513: The `begin_iteration()` function is only used in this one spot. It might be cleaner and easier to follow if we move the internals of that check (it looks to be checking whether the entry->dispatch_queue_context->io_dispatch_loop is NULL).

Also, instead of `goto iteration_done`, since there's only one spot where we use `iteration_done` we can execute the iteration_done code block here directly and return.

include/aws/io/io.h
```diff
@@ -14,12 +14,19 @@ AWS_PUSH_SANE_WARNING_LEVEL
...
 #define AWS_C_IO_PACKAGE_ID 1
 
+struct aws_io_handle;
+typedef void aws_io_set_queue_on_handle_fn(struct aws_io_handle *handle, void *queue);
+typedef void aws_io_clear_queue_on_handle_fn(struct aws_io_handle *handle);
+
 struct aws_io_handle {
     union {
         int fd;
+        /* on Apple systems, handle is of type nw_connection_t. On Windows, it's a SOCKET handle. */
```
sbSteveK, line 24: Clarity needed. I don't think we're using nw_connection_t unless dispatch queue is turned on.

"On Apple systems using dispatch queue event loops handle is of type nw_connection_t"

include/aws/io/private/event_loop_impl.h
```diff
@@ -152,6 +152,16 @@ AWS_IO_API
 struct _OVERLAPPED *aws_overlapped_to_windows_overlapped(struct aws_overlapped *overlapped);
 #endif /* AWS_ENABLE_IO_COMPLETION_PORTS */
 
+/**
+ * @internal - Don't use outside of testing.
+ *
+ * Return the default event loop type. If the return value is `AWS_ELT_PLATFORM_DEFAULT`, the function failed to
```
sbSteveK, line 158: I think you changed this to `AWS_EVENT_LOOP_PLATFORM_DEFAULT` as `AWS_ELT_PLATFORM_DEFAULT` doesn't exist.

Also, it looks like there is nothing returned on a failure to retrieve the default type value. I think its a compile time error?

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,698 @@
...
+    return true;
+}
+
+/**
+ * Clean up the related resource and determine if we should schedule next iteration.
+ * The function should be wrapped with dispatch_loop->context.lock to retain the dispatch loop while running.
+ * */
+static void end_iteration(struct scheduled_service_entry *entry) {
+
+    struct dispatch_loop_context *context = entry->dispatch_queue_context;
```
sbSteveK, line 476: Let's rename this variable to be consistent in its name across various functions for clarity. context -> dispatch_queue_context

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,698 @@
...
+    dispatch_loop = aws_mem_calloc(alloc, 1, sizeof(struct dispatch_loop));
+    dispatch_loop->allocator = alloc;
+    loop->impl_data = dispatch_loop;
+    dispatch_loop->base_loop = loop;
+
+    char dispatch_queue_id[AWS_IO_APPLE_DISPATCH_QUEUE_ID_LENGTH] = {0};
+    s_get_unique_dispatch_queue_id(dispatch_queue_id);
+
+    dispatch_loop->dispatch_queue = dispatch_queue_create(dispatch_queue_id, DISPATCH_QUEUE_SERIAL);
+    if (!dispatch_loop->dispatch_queue) {
```
sbSteveK, line 307: `dispatch_queue_create` Apple API will always return a dispatch_queue_t. If it can't a crash on the OS level is expected on this API call.

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,698 @@
...
+    /* make sure the loop is running so we can schedule a last task. */
+    s_run(event_loop);
+
+    /* cancel outstanding tasks */
+    dispatch_async_and_wait_f(dispatch_loop->dispatch_queue, dispatch_loop, s_dispatch_queue_destroy_task);
+
+    AWS_LOGF_TRACE(AWS_LS_IO_EVENT_LOOP, "id=%p: Releasing Dispatch Queue.", (void *)event_loop);
+}
+
+static int s_wait_for_stop_completion(struct aws_event_loop *event_loop) {
```
xiazhvera, line 412: Suspend will not stop the dispatch queue immediately, but it will finish the current iteration.

source/darwin/dispatch_queue_event_loop.c
sfod: Debatable: locking should happen right around the critical section, not for the whole function. So, instead of current
```
void func()
{
    crit_section_1();
    crit_section_2();
}

lock(m1);
lock(m2);
func();
unlock(m2);
unlock(m1);
```

should be

```
void func()
{
    lock(m1);
    crit_section_1();
    unlock(m1);

    lock(m2);
    crit_section_2();
    unlock(m2);
}
```

It'll be easier to reason about locking. It'll be less locking contention (though it's not a problem here afaics). Less chance to get deadlock, because locks will less overlap with each other.

sfod: A followup: with the current locking order, deadlock can probably happen on shutdown.

Thread1:
```
s_dispatch_event_loop_destroy {
  WLOCK(ctx)                // 3. lock
  WUNLOCK(ctx)
}
```

Thread2:
```
s_run_iteration {
  RLOCK(ctx)                 // 1. acquire
    LOCK(cross-thread)       // 5. deadlock
    UNLOCK(cross-thread)
  RUNLOCK(ctx)
}
```

Thread3:
```
s_run {
  LOCK(cross-thread)         // 2. acquire
    RLOCK(ctx)               // 4. deadlock (because of wlock)
    RUNLOCK(ctx)
  UNLOCK(cross-thread)
}
```

Not 100% sure if it's actually possible that these 3 functions are running in parallel (though it seems it's possible), but there are other functions where these 2 locks are obtained together.

The order of locking should be determined and the same everywhere.

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,716 @@
...
+static void s_dispatch_queue_destroy_task(void *context) {
+    struct aws_dispatch_loop *dispatch_loop = context;
+    AWS_LOGF_TRACE(AWS_LS_IO_EVENT_LOOP, "id=%p: Releasing Dispatch Queue.", (void *)dispatch_loop->base_loop);
+
+    s_lock_synced_data(dispatch_loop);
+    dispatch_loop->synced_data.current_thread_id = aws_thread_current_thread_id();
+    dispatch_loop->synced_data.is_executing = true;
+
+    /*
+     * Because this task was scheudled on the dispatch queue using `dispatch_async_and_wait_t()` we are certain that
```
sfod, line 319: nit: typo dispatch_async_and_wait_t -> dispatch_async_and_wait_f

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,716 @@
...
+
+    s_lock_synced_data(dispatch_loop);
+
+    /*
+     * Check if more cross thread tasks have been added since cancelling existing tasks. If there were, we must run
+     * them with AWS_TASK_STATUS_CANCELED as well before moving on with cleanup and destruction.
+     */
+    if (!aws_linked_list_empty(&dispatch_loop->synced_data.cross_thread_tasks)) {
+        goto populate_local_cross_thread_tasks;
+    }
```
sfod, line 365: I don't think this block works as intended.
It seems, with this block we try to handle a situation when another thread is scheduling a new task. But if this block is the first to lock the lock, then another thread will insert the new task after the block.

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,716 @@
...
+    AWS_LOGF_TRACE(AWS_LS_IO_EVENT_LOOP, "id=%p: cancelling task %p", (void *)event_loop, (void *)task);
+    struct aws_dispatch_loop *dispatch_loop = event_loop->impl_data;
+
+    /* First we move all cross thread tasks into the scheduler in case the task to be cancelled hasn't moved yet. */
+    struct aws_linked_list local_cross_thread_tasks;
+    aws_linked_list_init(&local_cross_thread_tasks);
+    s_lock_synced_data(dispatch_loop);
+    aws_linked_list_swap_contents(&dispatch_loop->synced_data.cross_thread_tasks, &local_cross_thread_tasks);
+    s_unlock_synced_data(dispatch_loop);
+    while (!aws_linked_list_empty(&local_cross_thread_tasks)) {
```
sfod, line 658: A race condition can appear when `s_run_iteration` already took cross threads tasks and currently scheduling them. This function should wait in such a case.

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,716 @@
...
+/* The synced_data_lock is held when any member of `aws_dispatch_loop`'s `synced_data` is accessed or modified */
+static int s_lock_synced_data(struct aws_dispatch_loop *dispatch_loop) {
+    return aws_mutex_lock(&dispatch_loop->synced_data.synced_data_lock);
+}
+
+static int s_unlock_synced_data(struct aws_dispatch_loop *dispatch_loop) {
+    return aws_mutex_unlock(&dispatch_loop->synced_data.synced_data_lock);
+}
+
+// Not sure why use 7 as the default queue size. Just follow what we used in task_scheduler.c
```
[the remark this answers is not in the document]
sbSteveK: When I saw this I dug into it a bit to see what was going on and couldn't find why it was set to 7 elsewhere. It doesn't appear to be for any particular sizing reasons and from what I could tell, could probably be set to pretty much anything. I'm going to lower this to 4 since in my head, that feels like a more reasonable initial limit.

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,716 @@
...
+     */
+    dispatch_loop->dispatch_queue = dispatch_queue_create(dispatch_queue_id, DISPATCH_QUEUE_SERIAL);
+
+    /*
+     * Suspend will increase the dispatch reference count.
+     * A suspended dispatch queue must have dispatch_release() called on it for Apple to release the dispatch queue.
+     * We suspend the newly created Apple dispatch queue here to conform with other event loop types. A new event loop
+     * should start in a non-running state until run() is called.
+     */
+    dispatch_suspend(dispatch_loop->dispatch_queue);
```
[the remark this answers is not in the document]
sbSteveK: I'll clear this comment up. We shouldn't expose (or suggest) anyone decref this using dispatch_release(). During the cleanup, s_run() is ALWAYS called on the dispatch loop to insure it decrefs on Apple's side as well as insure it will run blocks currently scheduled on it (to release the iteration events from memory and cancel tasks) and be able to run the block that is the destroy task.

source/darwin/dispatch_queue_event_loop.c
sbSteveK: Nice catch. I have enclosed the scheduling of cross-thread tasks in run_iteration into the synced data lock there so that  by the time we are here, those tasks are already fully moved into task scheduler.

This should be a non-issue for the destroy task as we are cancelling everything including tasks in the cross_thread_tasks list.

source/darwin/dispatch_queue_event_loop.c
sbSteveK: the purpose of this isn't to capture tasks scheduled by other threads but instead is to cancel tasks scheduled by tasks being cancelled. No other thread should be attempting to schedule tasks on this event loop at this point as we only reach this when the event loop group is destroying all its event loops.

The API we added that allows acquiring a refcount of the parent elg should prevent this from being a thing that ever happens. Unsure whether we need to protect against this scenario but open to discussion.

sfod: If that's the case, it should be fine I guess

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,727 @@
...
+
+        /* Timestamp 0 is used to denote "now" tasks */
+        if (task->timestamp == 0) {
+            aws_task_scheduler_schedule_now(&dispatch_loop->scheduler, task);
+        } else {
+            aws_task_scheduler_schedule_future(&dispatch_loop->scheduler, task, task->timestamp);
+        }
+    }
+
+    s_unlock_synced_data(dispatch_loop);
```
sfod, line 516: This unlock can be returned to its old place above the while loop, because `s_cancel_task` allows to be called only on the event loop thread. So it's impossible to get any race conditions between `s_run_iteration` and `s_cancel_task`.

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,727 @@
...
+/*
+ * Because dispatch queue is async we may need to acquire a refcount of the parent event loop group to prevent
+ * the event loop or dispatch loop from being cleaned out from underneath something that needs it. We expose the
+ * base elg so anything that needs to insure the event loops and dispatch loops don't get prematurely cleaned can
+ * hold a refcount.
+ */
+static void *s_get_base_event_loop_group(struct aws_event_loop *event_loop) {
+    struct aws_dispatch_loop *dispatch_loop = event_loop->impl_data;
+    return dispatch_loop->base_elg;
+}
```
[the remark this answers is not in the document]
sbSteveK: This test was one of the specific instances where we are manually cleaning up the elg before the socket and THEN calling socket cleanup. My current understanding is that the work Vera is doing in socket should cover these specific cases (probably through separate testing for dispatch queue for these types of tests). From the perspective of dispatch queue and apple nw_connections, we NEVER want to be in a scenario where the elg/event loops/dispatch loops are cleaned up before the socket.

bretambrose (commented):
This was left over from a few weeks ago.  Unsure if its still relevant.

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -0,0 +1,727 @@
...
+#include <aws/common/clock.h>
+#include <aws/common/mutex.h>
+#include <aws/common/task_scheduler.h>
+#include <aws/common/uuid.h>
+
+#include <aws/io/logging.h>
+
+#include <unistd.h>
+
+#include "./dispatch_queue_event_loop_private.h" // private header
```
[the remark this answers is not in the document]
xiazhvera: The file will be used by other source files in the future secitem changes. We will just make it a private file in this PR first.

PR #662, opened 2024-07-29
Apple Network Framework Socket Changes

sbSteveK:
Apple Network Framework socket integration

### DIFF
--- a/README.md
+++ b/README.md
@@ -646,48 +646,11 @@ All exported functions, simply shim into the v-table and return.
 
 We include a cross-platform API for sockets. We support TCP and UDP using IPv4 and IPv6, and Unix Domain sockets. On Windows,
 we use Named Pipes to support the functionality of Unix Domain sockets. On Windows, this is implemented with winsock2, and on
-all unix platforms we use the posix API. Then we will use Apple Network Framework on Apple platforms. 
+all unix platforms we use the posix API. We also provides options to use Apple Network Framework on Apple.
 
 Upon a connection being established, the new socket (either as the result of a `connect()` or `start_accept()` call)
 will not be attached to any event loops. It is your responsibility to register it with an event loop to begin receiving
 notifications.
-
-
-#### V-Table
-
-    struct aws_socket_vtable {
-        void (*socket_cleanup_fn)(struct aws_socket *socket);
-        int (*socket_connect_fn)(
-            struct aws_socket *socket,
-            const struct aws_socket_endpoint *remote_endpoint,
-            struct aws_event_loop *event_loop,
-            aws_socket_on_connection_result_fn *on_connection_result,
-            void *user_data);
-        int (*socket_bind_fn)(struct aws_socket *socket, const struct aws_socket_endpoint *local_endpoint);
-        int (*socket_listen_fn)(struct aws_socket *socket, int backlog_size);
-        int (*socket_start_accept_fn)(
-            struct aws_socket *socket,
-            struct aws_event_loop *accept_loop,
-            aws_socket_on_accept_result_fn *on_accept_result,
-            void *user_data);
-        int (*socket_stop_accept_fn)(struct aws_socket *socket);
-        int (*socket_close_fn)(struct aws_socket *socket);
-        int (*socket_shutdown_dir_fn)(struct aws_socket *socket, enum aws_channel_direction dir);
-        int (*socket_set_options_fn)(struct aws_socket *socket, const struct aws_socket_options *options);
-        int (*socket_assign_to_event_loop_fn)(struct aws_socket *socket, struct aws_event_loop *event_loop);
-        int (*socket_subscribe_to_readable_events_fn)(
-            struct aws_socket *socket,
-            aws_socket_on_readable_fn *on_readable,
-            void *user_data);
-        int (*socket_read_fn)(struct aws_socket *socket, struct aws_byte_buf *buffer, size_t *amount_read);
-        int (*socket_write_fn)(
-            struct aws_socket *socket,
-            const struct aws_byte_cursor *cursor,
-            aws_socket_on_write_completed_fn *written_fn,
-            void *user_data);
-        int (*socket_get_error_fn)(struct aws_socket *socket);
-        bool (*socket_is_open_fn)(struct aws_socket *socket);
-    };
 
 #### API
     typedef enum aws_socket_domain {
@@ -753,10 +716,13 @@ upon completion of asynchronous operations. If you are using UDP or LOCAL, `conn
 
 Shuts down any pending operations on the socket, and cleans up state. The socket object can be re initialized after this operation.
 
-    int aws_socket_connect(struct aws_socket *socket, struct aws_socket_endpoint *remote_endpoint);
+    int aws_socket_set_cleanup_complete_callback(struct aws_socket *socket, aws_socket_on_shutdown_complete_fn fn, void *user_data);
 
-Connects to a remote endpoint. In TCP and all Apple Network Framework connections (regardless it is UDP, TCP or LOCAL), when the connection succeed, you still must wait on  the `on_connection_established()` callback to 
-be invoked before using the socket.
+Sets the clean up completion callback. The callback will be invoked if `aws_socket_clean_up()` finish to clean up the socket resources. It is safe to release the socket memory after this callback is invoked.
+
+    int aws_socket_connect(struct aws_socket *socket, const struct aws_socket_endpoint *remote_endpoint, struct aws_event_loop *event_loop, aws_socket_on_connection_result_fn *on_connection_result, void *user_data);
+
+Connects to a remote endpoint. In TCP and all Apple Network Framework connections (regardless it is UDP, TCP or LOCAL), when the connection succeed, you still must wait on  the `on_connection_result()` callback to be invoked before using the socket.
 
 In UDP, this simply binds the socket to a remote address for use with `aws_socket_write()`, and if the operation is successful, 
 the socket can immediately be used for write operations.
@@ -773,10 +739,9 @@ modes or if you are using Apple Network Framework (regardless it is UDP or TCP),
 
 TCP, LOCAL, and Apple Network Framework only. Sets up the socket to listen on the address bound to in `aws_socket_bind()`.
 
-    int aws_socket_start_accept(struct aws_socket *socket);
+    int aws_socket_start_accept(struct aws_socket *socket, struct aws_event_loop *accept_loop, struct aws_socket_listener_options options);
 
-TCP, LOCAL, and Apple Network Framework only. The socket will begin accepting new connections. This is an asynchronous operation. New 
-connections will arrive via the `on_incoming_connection()` callback.
+TCP, LOCAL, and Apple Network Framework only. The socket will begin accepting new connections. This is an asynchronous operation. `on_accept_start()` will be invoked when the listener is ready to accept new connection. New connections will arrive via the `on_accept_result()` callback. 
 
     int aws_socket_stop_accept(struct aws_socket *socket);
 
@@ -787,14 +752,17 @@ again after this operation.
 
 Calls `close()` on the socket and unregisters all io operations from the event loop.
 
+    int aws_socket_set_close_complete_callback(struct aws_socket *socket, aws_socket_on_shutdown_complete_fn fn, void *user_data);
+
+Sets the close completion callback. The callback will be invoked if `aws_socket_close()` finish to process all the I/O events and close the socket.
+
     struct aws_io_handle *aws_socket_get_io_handle(struct aws_socket *socket);
 
 Fetches the underlying io handle for use in event loop registrations and channel handlers.
 
     int aws_socket_set_options(struct aws_socket *socket, struct aws_socket_options *options);
 
-Sets new socket options on the underlying socket. This is mainly useful in context of accepting a new connection via:
-`on_incoming_connection()`.
+Sets new socket options on the underlying socket.
 
     int aws_socket_read(struct aws_socket *socket, struct aws_byte_buf *buffer, size_t *amount_read);
 
--- a/include/aws/io/channel_bootstrap.h
+++ b/include/aws/io/channel_bootstrap.h
@@ -298,7 +298,8 @@ AWS_IO_API int aws_server_bootstrap_set_alpn_callback(
  * callbacks are invoked the thread of the event-loop that the listening socket is assigned to
  *
  * `setup_callback`. If set, the callback will be asynchronously invoked when the listener is ready for use. For Apple
- * Network Framework, the listener is not usable until the callback is invoked.
+ * Network Framework, the listener is not usable until the callback is invoked. If the listener creation failed (returned NULL), 
+ * the `setup_callback` will not be invoked.
  *
  * Upon shutdown of your application, you'll want to call `aws_server_bootstrap_destroy_socket_listener` with the return
  * value from this function.
--- a/include/aws/io/channel_bootstrap.h
+++ b/include/aws/io/channel_bootstrap.h
@@ -133,6 +133,14 @@ typedef void(aws_server_bootstrap_on_accept_channel_shutdown_fn)(
     void *user_data);
 
 /**
+ * This function is only used for async listener (Apple Network Framework in this case).
+ * Once the server listener socket is finished setup and starting listening, this fuction
+ * will be invoked.
+ */
+typedef void(
+    aws_server_bootstrap_on_listener_setup_fn)(struct aws_server_bootstrap *bootstrap, int error_code, void *user_data);
+
+/**
  * Once the server listener socket is finished destroying, and all the existing connections are closed, this fuction
  * will be invoked.
  */
@@ -212,6 +220,7 @@ struct aws_server_socket_channel_bootstrap_options {
     const struct aws_tls_connection_options *tls_options;
     aws_server_bootstrap_on_accept_channel_setup_fn *incoming_callback;
     aws_server_bootstrap_on_accept_channel_shutdown_fn *shutdown_callback;
+    aws_server_bootstrap_on_listener_setup_fn *setup_callback;
     aws_server_bootstrap_on_server_listener_destroy_fn *destroy_callback;
     bool enable_read_back_pressure;
     void *user_data;
@@ -294,6 +303,9 @@ AWS_IO_API int aws_server_bootstrap_set_alpn_callback(
  * bootstrap_options is copied.
  */
 AWS_IO_API struct aws_socket *aws_server_bootstrap_new_socket_listener(
+    const struct aws_server_socket_channel_bootstrap_options *bootstrap_options);
+
+AWS_IO_API struct aws_socket *aws_server_bootstrap_new_socket_listener_async(
     const struct aws_server_socket_channel_bootstrap_options *bootstrap_options);
 
 /**
--- a/include/aws/io/channel_bootstrap.h
+++ b/include/aws/io/channel_bootstrap.h
@@ -298,8 +298,8 @@ AWS_IO_API int aws_server_bootstrap_set_alpn_callback(
  * callbacks are invoked the thread of the event-loop that the listening socket is assigned to
  *
  * `setup_callback`. If set, the callback will be asynchronously invoked when the listener is ready for use. For Apple
- * Network Framework, the listener is not usable until the callback is invoked. If the listener creation failed (returned NULL), 
- * the `setup_callback` will not be invoked.
+ * Network Framework, the listener is not usable until the callback is invoked. If the listener creation failed
+ * (return NULL), the `setup_callback` will not be invoked.
  *
  * Upon shutdown of your application, you'll want to call `aws_server_bootstrap_destroy_socket_listener` with the return
  * value from this function.
--- a/include/aws/io/channel_bootstrap.h
+++ b/include/aws/io/channel_bootstrap.h
@@ -218,9 +218,9 @@ struct aws_server_socket_channel_bootstrap_options {
     uint32_t port;
     const struct aws_socket_options *socket_options;
     const struct aws_tls_connection_options *tls_options;
+    aws_server_bootstrap_on_listener_setup_fn *setup_callback;
     aws_server_bootstrap_on_accept_channel_setup_fn *incoming_callback;
     aws_server_bootstrap_on_accept_channel_shutdown_fn *shutdown_callback;
-    aws_server_bootstrap_on_listener_setup_fn *setup_callback;
     aws_server_bootstrap_on_server_listener_destroy_fn *destroy_callback;
     bool enable_read_back_pressure;
     void *user_data;
@@ -297,15 +297,15 @@ AWS_IO_API int aws_server_bootstrap_set_alpn_callback(
  * shutting down. Immediately after the `shutdown_callback` returns, the channel is cleaned up automatically. All
  * callbacks are invoked the thread of the event-loop that the listening socket is assigned to
  *
+ * `setup_callback`. If set, the callback will be asynchronously invoked when the listener is ready for use. For Apple
+ * Network Framework, the listener is not usable until the callback is invoked.
+ *
  * Upon shutdown of your application, you'll want to call `aws_server_bootstrap_destroy_socket_listener` with the return
  * value from this function.
  *
  * bootstrap_options is copied.
  */
 AWS_IO_API struct aws_socket *aws_server_bootstrap_new_socket_listener(
-    const struct aws_server_socket_channel_bootstrap_options *bootstrap_options);
-
-AWS_IO_API struct aws_socket *aws_server_bootstrap_new_socket_listener_async(
     const struct aws_server_socket_channel_bootstrap_options *bootstrap_options);
 
 /**
--- a/include/aws/io/socket.h
+++ b/include/aws/io/socket.h
@@ -121,14 +121,15 @@ typedef void(aws_socket_on_accept_result_fn)(
  * Callback for when the data passed to a call to aws_socket_write() has either completed or failed.
  * On success, error_code will be AWS_ERROR_SUCCESS.
  *
- * `socket` may be NULL in the callback if the socket is released and cleaned up before a callback is triggered.
- * by the system I/O handler,
+ * `socket` may be NULL in the callback if the socket is released and cleaned up before the callback is triggered.
  */
 typedef void(
     aws_socket_on_write_completed_fn)(struct aws_socket *socket, int error_code, size_t bytes_written, void *user_data);
 /**
  * Callback for when socket is either readable (edge-triggered) or when an error has occurred. If the socket is
  * readable, error_code will be AWS_ERROR_SUCCESS.
+ *
+ * `socket` may be NULL in the callback if the socket is released and cleaned up before the callback is triggered.
  */
 typedef void(aws_socket_on_readable_fn)(struct aws_socket *socket, int error_code, void *user_data);
 
--- a/include/aws/io/socket.h
+++ b/include/aws/io/socket.h
@@ -413,7 +413,9 @@ AWS_IO_API void aws_socket_endpoint_init_local_address_for_test(struct aws_socke
 AWS_IO_API bool aws_is_network_interface_name_valid(const char *interface_name);
 
 /**
- * Get default impl type based on the */
+ * Get default impl type based on the platform. 
+ * For user in internal tests only 
+ */
 AWS_IO_API enum aws_socket_impl_type aws_socket_get_default_impl_type(void);
 
 AWS_EXTERN_C_END
--- a/include/aws/io/socket.h
+++ b/include/aws/io/socket.h
@@ -412,6 +412,10 @@ AWS_IO_API void aws_socket_endpoint_init_local_address_for_test(struct aws_socke
  * network_interface_name on Windows */
 AWS_IO_API bool aws_is_network_interface_name_valid(const char *interface_name);
 
+/**
+ * Get default impl type based on the */
+AWS_IO_API enum aws_socket_impl_type aws_socket_get_default_impl_type(void);
+
 AWS_EXTERN_C_END
 AWS_POP_SANE_WARNING_LEVEL
 
--- a/include/aws/io/socket.h
+++ b/include/aws/io/socket.h
@@ -96,6 +96,15 @@ struct aws_event_loop;
 typedef void(aws_socket_on_connection_result_fn)(struct aws_socket *socket, int error_code, void *user_data);
 
 /**
+ * Called by a listening socket when a listener has succeeded or an error has occurred.
+ * If the connection was successful error_code will be AWS_ERROR_SUCCESS and the socket has already been assigned
+ * to the event loop specified in aws_socket_connect().
+ *
+ * If an error occurred error_code will be non-zero.
+ */
+typedef void(aws_socket_on_listen_result_fn)(struct aws_socket *socket, int error_code, void *user_data);
+
+/**
  * Called by a listening socket when either an incoming connection has been received or an error occurred.
  *
  * In the normal use-case, this function will be called multiple times over the lifetime of a single listening socket.
@@ -235,6 +244,24 @@ AWS_IO_API int aws_socket_start_accept(
     struct aws_event_loop *accept_loop,
     aws_socket_on_accept_result_fn *on_accept_result,
     void *user_data);
+
+/**
+ * Apple Network Framework only. The socket will begin accepting new connections. This is an asynchronous operation. The
+ *  `on_listen_result` will be invoked when the listen is setup. Once the `on_listen_result` is invoked with success
+ * state, the socket start to listen.
+ *
+ * New connections or errors will arrive via the `on_accept_result` callback.
+ *
+ * aws_socket_bind() and aws_socket_listen() must be called before calling this function.
+ *
+ */
+AWS_IO_API int aws_socket_start_accept_async(
+    struct aws_socket *socket,
+    struct aws_event_loop *accept_loop,
+    aws_socket_on_accept_result_fn *on_accept_result,
+    void *on_accept_user_data,
+    aws_socket_on_listen_result_fn *on_listen_result,
+    void *on_listen_user_data);
 
 /**
  * TCP, LOCAL and VSOCK only. The listening socket will stop accepting new connections.
--- a/include/aws/io/socket.h
+++ b/include/aws/io/socket.h
@@ -67,9 +67,6 @@ struct aws_socket_options {
     uint16_t keep_alive_max_failed_probes;
     bool keepalive;
 
-    aws_socket_on_shutdown_complete_fn *on_shutdown_complete;
-    void *shutdown_user_data;
-
     /**
      * THIS IS AN EXPERIMENTAL AND UNSTABLE API
      * (Optional)
@@ -281,8 +278,7 @@ AWS_IO_API int aws_socket_close(struct aws_socket *socket);
 AWS_IO_API int aws_socket_shutdown_dir(struct aws_socket *socket, enum aws_channel_direction dir);
 
 /**
- * Sets new socket options on the underlying socket. This is mainly useful in context of accepting a new connection via:
- * `on_incoming_connection()`. options is copied.
+ * Sets new socket options on the underlying socket.
  */
 AWS_IO_API int aws_socket_set_options(struct aws_socket *socket, const struct aws_socket_options *options);
 
--- a/include/aws/io/socket.h
+++ b/include/aws/io/socket.h
@@ -245,10 +245,20 @@ AWS_IO_API int aws_socket_start_accept(
  */
 AWS_IO_API int aws_socket_stop_accept(struct aws_socket *socket);
 
+
+/**
+ * Apple Network Framework only. The callback that will triggered when aws_socket_close() finished. The callback 
+ * will be called from the socket event loop. 
+ */
 AWS_IO_API int aws_socket_set_close_complete_callback(
     struct aws_socket *socket,
     aws_socket_on_shutdown_complete_fn fn,
     void *user_data);
+
+/**
+ * Apple Network Framework only. The callback that will triggered when aws_socket_cleanup() finished. And 
+ * it is only safe to release the socket afterwards. The callback will be called from the socket event loop. 
+ */
 AWS_IO_API int aws_socket_set_cleanup_complete_callback(
     struct aws_socket *socket,
     aws_socket_on_shutdown_complete_fn fn,
--- a/include/aws/io/socket.h
+++ b/include/aws/io/socket.h
@@ -264,24 +264,6 @@ AWS_IO_API int aws_socket_start_accept(
 AWS_IO_API int aws_socket_stop_accept(struct aws_socket *socket);
 
 /**
- * Apple Network Framework only. The callback that will triggered when aws_socket_close() finished. The callback
- * will be called from the socket event loop.
- */
-AWS_IO_API int aws_socket_set_close_complete_callback(
-    struct aws_socket *socket,
-    aws_socket_on_shutdown_complete_fn fn,
-    void *user_data);
-
-/**
- * Apple Network Framework only. The callback that will triggered when aws_socket_cleanup() finished. And
- * it is only safe to release the socket afterwards. The callback will be called from the socket event loop.
- */
-AWS_IO_API int aws_socket_set_cleanup_complete_callback(
-    struct aws_socket *socket,
-    aws_socket_on_shutdown_complete_fn fn,
-    void *user_data);
-
-/**
  * Calls `close()` on the socket and unregisters all io operations from the event loop. This function must be called
  * from the event-loop's thread unless this is a listening socket. If it's a listening socket it can be called from any
  * non-event-loop thread or the event-loop the socket is currently assigned to. If called from outside the event-loop,
@@ -365,6 +347,24 @@ AWS_IO_API int aws_socket_write(
     void *user_data);
 
 /**
+ * Apple Network Framework only. The callback that will triggered when aws_socket_close() finished. The callback
+ * will be called from the socket event loop.
+ */
+AWS_IO_API int aws_socket_set_close_complete_callback(
+    struct aws_socket *socket,
+    aws_socket_on_shutdown_complete_fn fn,
+    void *user_data);
+
+/**
+ * Apple Network Framework only. The callback that will triggered when aws_socket_cleanup() finished. And
+ * it is only safe to release the socket afterwards. The callback will be called from the socket event loop.
+ */
+AWS_IO_API int aws_socket_set_cleanup_complete_callback(
+    struct aws_socket *socket,
+    aws_socket_on_shutdown_complete_fn fn,
+    void *user_data);
+
+/**
  * Gets the latest error from the socket. If no error has occurred AWS_OP_SUCCESS will be returned. This function does
  * not raise any errors to the installed error handlers.
  */
@@ -405,7 +405,7 @@ AWS_IO_API bool aws_is_network_interface_name_valid(const char *interface_name);
 
 /**
  * Get default impl type based on the platform.
- * For user in internal tests only
+ * For user in internal tests only.
  */
 AWS_IO_API enum aws_socket_impl_type aws_socket_get_default_impl_type(void);
 
--- a/include/aws/io/socket.h
+++ b/include/aws/io/socket.h
@@ -96,13 +96,13 @@ struct aws_event_loop;
 typedef void(aws_socket_on_connection_result_fn)(struct aws_socket *socket, int error_code, void *user_data);
 
 /**
- * Called by a listening socket when a listener has succeeded or an error has occurred.
- * If the connection was successful error_code will be AWS_ERROR_SUCCESS and the socket has already been assigned
- * to the event loop specified in aws_socket_connect().
+ * Called by a listening socket when a listener accept has successfully initialized or an error has occurred.
+ * If the listener was successful error_code will be AWS_ERROR_SUCCESS and the socket has already been assigned
+ * to the event loop specified in aws_socket_start_accept().
  *
  * If an error occurred error_code will be non-zero.
  */
-typedef void(aws_socket_on_listen_result_fn)(struct aws_socket *socket, int error_code, void *user_data);
+typedef void(aws_socket_on_accept_started_fn)(struct aws_socket *socket, int error_code, void *user_data);
 
 /**
  * Called by a listening socket when either an incoming connection has been received or an error occurred.
@@ -169,6 +169,16 @@ struct aws_socket {
     aws_socket_on_accept_result_fn *accept_result_fn;
     void *connect_accept_user_data;
     void *impl;
+};
+
+struct aws_socket_listener_options {
+    aws_socket_on_accept_result_fn *on_accept_result;
+    void *on_accept_result_user_data;
+
+    // This callback is invoked when the listener starts accepting incoming connections.
+    // It is only triggered in asynchronous listener APIs while using nw_socket.
+    aws_socket_on_accept_started_fn *on_accept_start_result;
+    void *on_accept_start_user_data;
 };
 
 struct aws_byte_buf;
@@ -242,26 +252,7 @@ AWS_IO_API int aws_socket_listen(struct aws_socket *socket, int backlog_size);
 AWS_IO_API int aws_socket_start_accept(
     struct aws_socket *socket,
     struct aws_event_loop *accept_loop,
-    aws_socket_on_accept_result_fn *on_accept_result,
-    void *user_data);
-
-/**
- * Apple Network Framework only. The socket will begin accepting new connections. This is an asynchronous operation. The
- *  `on_listen_result` will be invoked when the listen is setup. Once the `on_listen_result` is invoked with success
- * state, the socket start to listen.
- *
- * New connections or errors will arrive via the `on_accept_result` callback.
- *
- * aws_socket_bind() and aws_socket_listen() must be called before calling this function.
- *
- */
-AWS_IO_API int aws_socket_start_accept_async(
-    struct aws_socket *socket,
-    struct aws_event_loop *accept_loop,
-    aws_socket_on_accept_result_fn *on_accept_result,
-    void *on_accept_user_data,
-    aws_socket_on_listen_result_fn *on_listen_result,
-    void *on_listen_user_data);
+    struct aws_socket_listener_options options);
 
 /**
  * TCP, LOCAL and VSOCK only. The listening socket will stop accepting new connections.
@@ -413,8 +404,8 @@ AWS_IO_API void aws_socket_endpoint_init_local_address_for_test(struct aws_socke
 AWS_IO_API bool aws_is_network_interface_name_valid(const char *interface_name);
 
 /**
- * Get default impl type based on the platform. 
- * For user in internal tests only 
+ * Get default impl type based on the platform.
+ * For user in internal tests only
  */
 AWS_IO_API enum aws_socket_impl_type aws_socket_get_default_impl_type(void);
 
--- a/include/aws/io/socket.h
+++ b/include/aws/io/socket.h
@@ -176,8 +176,8 @@ struct aws_socket_listener_options {
     void *on_accept_result_user_data;
 
     // This callback is invoked when the listener starts accepting incoming connections.
-    // It is only triggered in asynchronous listener APIs while using nw_socket.
-    aws_socket_on_accept_started_fn *on_accept_start_result;
+    // If the callback set, the socket must not be released before the callback invoked.
+    aws_socket_on_accept_started_fn *on_accept_start;
     void *on_accept_start_user_data;
 };
 
--- a/include/aws/io/socket.h
+++ b/include/aws/io/socket.h
@@ -245,10 +245,9 @@ AWS_IO_API int aws_socket_start_accept(
  */
 AWS_IO_API int aws_socket_stop_accept(struct aws_socket *socket);
 
-
 /**
- * Apple Network Framework only. The callback that will triggered when aws_socket_close() finished. The callback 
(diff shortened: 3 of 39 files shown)


source/darwin/nw_socket.c
```diff
@@ -412,89 +457,142 @@ static int s_socket_connect_fn(
     aws_event_loop_connect_handle_to_completion_port(event_loop, &socket->io_handle);
     socket->event_loop = event_loop;
 
+    nw_socket->on_connection_result_fn = on_connection_result;
+    nw_socket->connect_accept_user_data = user_data;
+
+    struct nw_socket *socket_impl = socket->impl;
+
+    nw_socket->connect_args = aws_mem_calloc(socket->allocator, 1, sizeof(struct nw_socket_connect_args));
+    if (!nw_socket->connect_args) {
```
sbSteveK, line 466: we do not need to check result of an aws_mem_calloc

source/darwin/nw_socket.c
```diff
@@ -73,18 +73,28 @@ enum socket_state {
...
     struct aws_ref_count ref_count;
-    nw_connection_t *nw_connection;
+    nw_connection_t nw_connection;
     nw_parameters_t socket_options_to_params;
     struct aws_linked_list read_queue;
     int last_error;
     aws_socket_on_readable_fn *on_readable;
     void *on_readable_user_data;
     bool setup_run;
+    bool setup_closing;
```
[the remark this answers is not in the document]
sbSteveK: On second thought, changed it to `socket_open`. We can start it as false and only flip it to true on a successful socket connection, and flip it back to false on a socket closure.

source/darwin/nw_socket.c
```diff
@@ -73,18 +73,28 @@ enum socket_state {
     CLOSED,
 };
 
+struct nw_socket_connect_args {
```
sbSteveK, line 76: renaming this to `nw_socket_timeout_args` to better reflect what it's used for.

source/darwin/nw_socket.c
```diff
@@ -73,18 +73,28 @@ enum socket_state {
...
     nw_parameters_t socket_options_to_params;
     struct aws_linked_list read_queue;
     int last_error;
     aws_socket_on_readable_fn *on_readable;
     void *on_readable_user_data;
     bool setup_run;
+    bool setup_closing;
     bool read_queued;
     bool is_listener;
+    struct nw_socket_connect_args *connect_args;
```
sbSteveK, line 95: renaming to `timeout_args` for clarity

README.md
```diff
@@ -717,32 +755,34 @@ Shuts down any pending operations on the socket, and cleans up state. The socket
 
     int aws_socket_connect(struct aws_socket *socket, struct aws_socket_endpoint *remote_endpoint);
 
-Connects to a remote endpoint. In UDP, this simply binds the socket to a remote address for use with `aws_socket_write()`,
-and if the operation is successful, the socket can immediately be used for write operations.
+Connects to a remote endpoint. In TCP and all Apple Network Framework connections (regardless it is UDP, TCP or LOCAL), this will
```
bretambrose, line 758: "this will function will not block." reword

Also, I don't know if this is the right wording.  Saying "will not block" is misleading; none of them "block" in the sense that they're waiting on some async process to complete, it's just sometimes the socket is immediately usable and sometimes it isn't.

README.md
```diff
@@ -717,32 +755,34 @@ Shuts down any pending operations on the socket, and cleans up state. The socket
...
+In UDP, this simply binds the socket to a remote address for use with `aws_socket_write()`, and if the operation is successful, 
+the socket can immediately be used for write operations.
 
 For LOCAL (Unix Domain Sockets or Named Pipes), the socket will be immediately ready for use upon a successful return.
 
     int aws_socket_bind(struct aws_socket *socket, struct aws_socket_endpoint *local_endpoint);
 
-Binds the socket to a local address. In UDP mode, the socket is ready for `aws_socket_read()` operations. In connection oriented
-modes, you still must call `aws_socket_listen()` and `aws_socket_start_accept()` before using the socket.
+Binds the socket to a local address. In UDP mode, the socket is ready for `aws_socket_read()` operations. In connection oriented
```
bretambrose, line 769: This if-else conditional behavior bothers me.  What happens if you call aws_socket_listen on a bound UDP socket?

source/channel.c
```diff
@@ -646,7 +646,12 @@ void aws_channel_schedule_task_future(
 }
 
 bool aws_channel_thread_is_callers_thread(struct aws_channel *channel) {
-    return aws_event_loop_thread_is_callers_thread(channel->loop);
+    // As aws_event_loop is not ref-counted, it is possible that the socket read/write callback get triggered after
+    // the event loop is released, ignore the function call here.
+    if (channel && channel->loop) {
```
bretambrose, line 651: This is alarming.  I would like to know a lot more about how this is possible: how can you have a channel still alive while its event loop is not?

tests/tls_handler_test.c
```diff
@@ -200,6 +200,9 @@ static int s_tls_common_tester_clean_up(struct tls_common_tester *tester) {
 
     aws_condition_variable_clean_up(&tester->condition_variable);
     aws_mutex_clean_up(&tester->mutex);
+    // wait for socket ref count drop and released
+    aws_thread_current_sleep(1000000000);
```
bretambrose, line 204: Doing this (presumably to prevent memory-still-in-use test failures) is the sign of a much larger problem that could easily make life terrible for us: we don't understand, can't reason about, and/or can't tightly control an async cleanup/shutdown process.  We should talk about this with Mike.

bretambrose (commented):
I haven't looked at nw_socket.c yet.  I'm concerned about larger issues implied by the changes outside that file first.

README.md
```diff
@@ -645,12 +646,49 @@ All exported functions, simply shim into the v-table and return.
...
 we use Named Pipes to support the functionality of Unix Domain sockets. On Windows, this is implemented with winsock2, and on
-all unix platforms we use the posix API.
+all unix platforms we use the posix API. Then we will use Apple Network Framework on Apple platforms. 
 
 Upon a connection being established, the new socket (either as the result of a `connect()` or `start_accept()` call)
 will not be attached to any event loops. It is your responsibility to register it with an event loop to begin receiving
 notifications.
 
+
+#### V-Table
```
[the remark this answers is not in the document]
xiazhvera: It would help with anybody who would like to extend the socket implementation. Also, we listed the vtable for other classes as well.

README.md
xiazhvera: aws_socket_listen would fail with `AWS_IO_SOCKET_ILLEGAL_OPERATION_FOR_STATE` on a bounded UDP socket.

tests/socket_test.c
```diff
@@ -237,7 +365,9 @@ static int s_test_socket_ex(
     ASSERT_INT_EQUALS(endpoint->port, bound_endpoint.port);
     ASSERT_STR_EQUALS(endpoint->address, bound_endpoint.address);
 
-    if (options->type == AWS_SOCKET_STREAM) {
+    // The Apple Network Framework always require a "start listener/start connection"
+    // for setup a server socket
+    if (options->type == AWS_SOCKET_STREAM || aws_event_loop_get_default_type() == AWS_EVENT_LOOP_DISPATCH_QUEUE) {
```
sfod, line 370: This condition looks like a leaked abstraction. Maybe we should introduce CONNECTED_UDP socket type instead.
Also, the check should be for Apple Network Framework, not for dispatch queue.

tests/socket_test.c
```diff
@@ -357,35 +468,225 @@ static int s_test_socket_ex(
...
+
+    aws_event_loop_group_release(el_group);
+    aws_io_library_clean_up();
+
+    s_sleep_for_dispatch_queue();
+
+    return 0;
+}
+
+static int s_test_socket_udp_dispatch_queue(
```
sfod, line 521: From my understanding, these is specific to apple network framework, not to dispatch queue. For example, if posix socket is used with dispatch queue, could this function still handle it?

source/channel_bootstrap.c
```diff
@@ -1533,7 +1773,194 @@ struct aws_socket *aws_server_bootstrap_new_socket_listener(
...
+    s_server_connection_args_acquire(server_connection_args);
+
+    if (aws_socket_start_accept_async(
+            &server_connection_args->listener,
+            connection_loop,
+            s_on_server_connection_result,
+            server_connection_args,
+            s_on_listener_connection_established,
+            server_connection_args)) {
+        goto cleanup_listener;
```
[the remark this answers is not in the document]
sfod: Typo: should be `s_server_connection_args_release`

source/darwin/nw_socket.c
```diff
@@ -0,0 +1,2075 @@
...
+                nw_parameters_create_secure_tcp(NW_PARAMETERS_DISABLE_PROTOCOL, ^(nw_protocol_options_t nw_options) {
+                  if (options->connect_timeout_ms) {
+                      /* this value gets set in seconds. */
+                      nw_tcp_options_set_connection_timeout(
+                          nw_options, options->connect_timeout_ms / AWS_TIMESTAMP_MILLIS);
+                  }
+
+                  // Only change default keepalive values if keepalive is true and both interval and timeout
+                  // are not zero.
+                  if (options->keepalive && options->keep_alive_interval_sec != 0 &&
```
bretambrose, line 282: Have we tested (manually with packet trace) that TCP keep alive works as expected?  There were subtle differences between windows and posix and we made some questionable API decisions there.

source/darwin/nw_socket.c
```diff
@@ -0,0 +1,2075 @@
...
+        } else if (options->domain == AWS_SOCKET_LOCAL) {
+            nw_socket->socket_options_to_params =
+                nw_parameters_create_secure_tcp(NW_PARAMETERS_DISABLE_PROTOCOL, NW_PARAMETERS_DEFAULT_CONFIGURATION);
+        }
+    } else if (options->type == AWS_SOCKET_DGRAM) {
+        nw_socket->socket_options_to_params =
+            nw_parameters_create_secure_udp(NW_PARAMETERS_DISABLE_PROTOCOL, NW_PARAMETERS_DEFAULT_CONFIGURATION);
+    }
+
+    if (!nw_socket->socket_options_to_params) {
```
bretambrose, line 306: I feel like we should have different error output whether or not
1. the call to `nw_parameters_*` failed
2. we fell through all the if statements and didn't even try to set up the parameters (for example, if VSOCK was the domain)

source/darwin/nw_socket.c
```diff
@@ -0,0 +1,2075 @@
...
+}
+
+struct read_queue_node {
+    struct aws_allocator *allocator;
+    dispatch_data_t received_data;
+    struct aws_linked_list_node node;
+    size_t current_offset;
+};
+
+static void s_clean_up_read_queue_node(struct read_queue_node *node) {
```
bretambrose, line 421: By convention we use `clean_up` when it's a function that cleans up struct members but not the struct itself (ie, it can be stack allocated) and `destroy` for things we know are on the heap, like this.

source/darwin/nw_socket.c
```diff
@@ -0,0 +1,2075 @@
...
+
+static void s_close_complete_callback(struct aws_task *task, void *arg, enum aws_task_status status) {
+    (void)status;
+    struct socket_close_complete_args *task_arg = arg;
+    struct aws_allocator *allocator = task_arg->allocator;
+    if (task_arg->shutdown_complete_fn) {
+        task_arg->shutdown_complete_fn(task_arg->user_data);
+    }
+    aws_ref_count_release(&task_arg->nw_socket->ref_count);
+    aws_mem_release(allocator, task_arg);
```
bretambrose, line 442: Given that the task and the args are tightly coupled, why not just have the task be a member of socket_close_complete_args?

source/darwin/nw_socket.c
```diff
@@ -0,0 +1,2075 @@
...
+    if (on_cleanup_complete) {
+        on_cleanup_complete(cleanup_user_data);
+    }
+}
+
+static void s_socket_internal_destroy(void *sock_ptr) {
+    struct nw_socket *nw_socket = sock_ptr;
+    AWS_LOGF_DEBUG(AWS_LS_IO_SOCKET, "id=%p : start s_socket_internal_destroy", (void *)sock_ptr);
+
+    aws_mutex_lock(&nw_socket->synced_data.lock);
```
bretambrose, line 480: Why is this lock needed?  synced_data protects base_socket which does not seem to be referenced here?

source/darwin/nw_socket.c
```diff
@@ -0,0 +1,2075 @@
...
+    const struct aws_socket_options *options) {
+    AWS_ASSERT(options);
+    AWS_ZERO_STRUCT(*socket);
+
+    // Network Interface is not supported with Apple Network Framework yet
+    size_t network_interface_length = 0;
+    if (aws_secure_strlen(options->network_interface_name, AWS_NETWORK_INTERFACE_NAME_MAX, &network_interface_length)) {
+        AWS_LOGF_ERROR(
+            AWS_LS_IO_SOCKET,
+            "id=%p fd=%d: network_interface_name max length must be %d length and NULL terminated",
```
bretambrose, line 518: "less than or equal to"?

source/darwin/nw_socket.c
```diff
@@ -0,0 +1,2075 @@
...
+
+    socket->allocator = alloc;
+    socket->options = *options;
+    socket->impl = nw_socket;
+    socket->vtable = &s_vtable;
+
+    aws_mutex_init(&nw_socket->synced_data.lock);
+    aws_mutex_init(&nw_socket->synced_state.lock);
+
+    aws_mutex_lock(&nw_socket->synced_data.lock);
```
bretambrose, line 544: Why do we need to lock when no one can see the newly initialized mutexes yet?

source/darwin/nw_socket.c
```diff
@@ -0,0 +1,2075 @@
...
+    nw_socket->synced_data.base_socket = socket;
+    aws_mutex_unlock(&nw_socket->synced_data.lock);
+
+    s_set_socket_state(nw_socket, socket, INIT);
+
+    if (s_setup_socket_params(nw_socket, options)) {
```
bretambrose, line 550: This should be moved rigth after the nw_socket is created (after the allocator set).  Otherwise a failure here leaves multiple mutexes un-cleaned-up.

source/darwin/nw_socket.c
```diff
@@ -0,0 +1,2075 @@
...
+        aws_mem_release(alloc, nw_socket);
+        return AWS_OP_ERR;
+    }
+    aws_ref_count_init(&nw_socket->ref_count, nw_socket, s_socket_impl_destroy);
+    aws_ref_count_init(&nw_socket->internal_ref_count, nw_socket, s_socket_internal_destroy);
+    aws_ref_count_acquire(&nw_socket->ref_count);
```
bretambrose, line 556: Why do we start off with a ref count of 2?

A thorough comment on the structure members explaining everything that affects the ref count would be nice.

source/darwin/nw_socket.c
```diff
@@ -0,0 +1,2075 @@
...
+            (void *)socket,
+            (void *)nw_socket->nw_connection);
+
+        int error_code = AWS_IO_SOCKET_TIMEOUT;
+
+        if (status != AWS_TASK_STATUS_RUN_READY) {
+            error_code = AWS_IO_EVENT_LOOP_SHUTDOWN;
+            // nw_socket->event_loop = NULL;
+        }
+        aws_raise_error(error_code);
```
bretambrose, line 593: Is this necessary?  We don't raise error (with the callback error code) in callbacks that apply the error code or pass it on.

source/darwin/nw_socket.c
```diff
@@ -0,0 +1,2075 @@
...
+        // Must set timeout_args to NULL to avoid double cancel. Clean up the timeout task
+        aws_mem_release(nw_socket->allocator, nw_socket->timeout_args);
+        nw_socket->timeout_args = NULL;
+        aws_socket_close(socket);
+        nw_socket->on_connection_result_fn(socket, error_code, nw_socket->connect_result_user_data);
+    } else { // else we simply clean up the timeout args
+        aws_mem_release(nw_socket->allocator, nw_socket->timeout_args);
```
bretambrose, line 600: I would like to see a comment explaining when this can happen

bretambrose (commented):
Partial initial review.  Will do more tomorrow.

include/aws/io/socket.h
```diff
@@ -358,6 +412,10 @@ AWS_IO_API void aws_socket_endpoint_init_local_address_for_test(struct aws_socke
  * network_interface_name on Windows */
 AWS_IO_API bool aws_is_network_interface_name_valid(const char *interface_name);
 
+/**
+ * Get default impl type based on the */
+AWS_IO_API enum aws_socket_impl_type aws_socket_get_default_impl_type(void);
```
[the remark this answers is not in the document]
xiazhvera: The `aws_socket_get_default_impl_type` is also used in downstream c-http library to determine if we should wait for async callbacks. Even with a generalize API, we might still need handle socket differently..

source/darwin/nw_socket.c
```diff
@@ -0,0 +1,2075 @@
...
+     * aws_socket_assign_event_loop()
+     */
+    struct aws_event_loop *event_loop;
+
+    /* Indicate the connection result is updated. Set to true if
+     *      1. nw_connection returned with state=`nw_connection_state_ready`, indicating the connection succeed
+     *      2. nw_connection returned with state=`nw_connection_state_failed`, indicating the connection failed
+     *      3. directly set to true for the incoming socket, as the incoming socket is already connected
+     */
+    bool connection_setup;
```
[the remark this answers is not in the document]
xiazhvera: Updated the comments. The argument is actually used to determine if we should cancel the timeout task. We don't really care if the connection succeed or failed, as long as we get a result back, we could cancel the timeout task from then.

source/darwin/nw_socket.c
```diff
@@ -0,0 +1,2075 @@
...
+        struct aws_mutex lock;
+        struct aws_socket *base_socket;
+    } synced_data;
+
+    /* The nw_socket_state. aws_socket also has a field `state` which should be represent the same parameter,
+    however, as
+     * it is possible that the aws_socket object is released while nw_socket is still alive, we will use
+     * nw_socket->state instead of socket->state to verify the socket_state. */
+    struct {
+        int state;
```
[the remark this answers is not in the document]
xiazhvera: The state was defined as int in the aws_socket, the nw_socket matches it. Besides, we also to bitwise operations on the state. 
https://github.com/awslabs/aws-c-io/blob/5fcecfc621059e254f2dc0dcac46265fcba0bf3a/include/aws/io/socket.h#L150

source/darwin/nw_socket.c
```diff
@@ -0,0 +1,2075 @@
...
+
+static inline void s_set_socket_state(struct nw_socket *nw_socket, struct aws_socket *socket, int state) {
+    s_lock_socket_state(nw_socket);
+    // The state can only go increasing, except for LISTENING and STOPPED. They can switch between each other.
+    if (nw_socket->synced_state.state < state || (state == LISTENING && nw_socket->synced_state.state == STOPPED)) {
+        nw_socket->synced_state.state = state;
+        if (socket) {
+            socket->state = state;
+        }
+    }
```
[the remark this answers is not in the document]
xiazhvera: Not really. User is allowed to clean up the aws_socket memory, while the nw_socket is still alive and processing events...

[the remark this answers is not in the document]
bretambrose: I'm not following.  We put an if check around what is claimed to be valid state transitions but then do nothing if that expectation is violated.

source/darwin/nw_socket.c
xiazhvera: I didnt see why that's necessary. The task is accessible from the task function call regardless.

include/aws/io/socket.h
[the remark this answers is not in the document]
bretambrose: Sure, but I'm talking about the entire public API for sockets.  How much of it **needs** to be public?

source/darwin/nw_socket.c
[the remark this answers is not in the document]
bretambrose: Let's use the enum as the type wherever possible.

As a followup, does the `aws_socket` type have to be public?

source/darwin/nw_socket.c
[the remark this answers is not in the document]
bretambrose: I'm talking about the outer if, not the inner

source/darwin/nw_socket.c
bretambrose: It's strictly simpler to allocate and free one thing than two.  As examples: 

* https://github.com/awslabs/aws-c-mqtt/blob/main/source/v5/mqtt5_client.c#L852-L858
* https://github.com/awslabs/aws-c-mqtt/blob/main/source/v5/mqtt5_client.c#L2101-L2107
* https://github.com/awslabs/aws-c-mqtt/blob/main/source/v5/mqtt5_to_mqtt3_adapter.c#L57-L61

source/darwin/nw_socket.c
```diff
@@ -0,0 +1,2075 @@
...
+    however, as
+     * it is possible that the aws_socket object is released while nw_socket is still alive, we will use
+     * nw_socket->state instead of socket->state to verify the socket_state. */
+    struct {
+        int state;
+        struct aws_mutex lock;
+    } synced_state;
+};
+
+struct socket_address {
```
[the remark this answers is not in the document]
xiazhvera: The struct should be shared between nw_socket and posix. Moved to `socket_impl.h`.

source/darwin/nw_socket.c
```diff
@@ -0,0 +1,2058 @@
...
+
+static void s_process_listener_success_task(struct aws_task *task, void *args, enum aws_task_status status) {
+    (void)status;
+    struct nw_listener_connection_args *task_args = args;
+    struct aws_allocator *allocator = task_args->allocator;
+    struct nw_socket *listener_nw_socket = task_args->nw_socket;
+
+    if (status == AWS_TASK_STATUS_RUN_READY) {
+
+        if (listener_nw_socket) {
```
bretambrose, line 1006: When would this be NULL?  If never, let's fatal assert, if possible, let's log or do something in the fall through case (currently there's nothing).

source/darwin/nw_socket.c
```diff
@@ -0,0 +1,2058 @@
...
+            struct aws_socket *listener = listener_nw_socket->synced_data.base_socket;
+
+            struct aws_socket *new_socket = aws_mem_calloc(allocator, 1, sizeof(struct aws_socket));
+            struct aws_socket_options options = listener->options;
+            int error = aws_socket_init(new_socket, allocator, &options);
+            if (error) {
+                aws_mem_release(allocator, new_socket);
+                nw_release(task_args->new_connection);
+                if (listener->accept_result_fn) {
+                    listener->accept_result_fn(listener, task_args->error_code, NULL, task_args->user_data);
```
bretambrose, line 1017: Why wouldn't we use `get_last_error()` as the error code here? Presumably task_args->error_code is 0 at this point?

source/darwin/nw_socket.c
```diff
@@ -0,0 +1,2058 @@
...
+
+                nw_endpoint_t endpoint = nw_connection_copy_endpoint(task_args->new_connection);
+                const char *hostname = nw_endpoint_get_hostname(endpoint);
+                uint16_t port = nw_endpoint_get_port(endpoint);
+
+                if (hostname != NULL) {
+                    size_t hostname_len = strlen(hostname);
+                    size_t buffer_size = AWS_ARRAY_SIZE(new_socket->remote_endpoint.address);
+                    size_t to_copy = aws_min_size(hostname_len, buffer_size);
+                    memcpy(new_socket->remote_endpoint.address, hostname, to_copy);
```
bretambrose, line 1032: This does not seem safe.  It would appear to allow a host name that is not null-terminated if it's exactly the size of the array.  Are we doing this anywhere else?

source/darwin/nw_socket.c
```diff
@@ -0,0 +1,2058 @@
...
+                    "id=%p handle=%p: incoming connection connected to %s:%d, the incoming handle is %p",
+                    (void *)listener,
+                    listener->io_handle.data.handle,
+                    new_socket->remote_endpoint.address,
+                    new_socket->remote_endpoint.port,
+                    new_socket->io_handle.data.handle);
+
+                if (listener->accept_result_fn) {
+                    listener->accept_result_fn(listener, task_args->error_code, new_socket, task_args->user_data);
+                } else // The connection is not sent to user, clean it up. The nw_connection should be released in
```
bretambrose, line 1064: How would this happen?  Why would we do all this work and then give up at the last instant because of something we could have checked at the beginning?

source/darwin/nw_socket.c
```diff
@@ -0,0 +1,2058 @@
...
+        nw_release(task_args->new_connection);
+    }
+
+    s_socket_release_internal_ref(listener_nw_socket);
+
+    aws_mem_release(task_args->allocator, task);
+    aws_mem_release(task_args->allocator, task_args);
+}
+
+static void s_schedule_on_listener_success(
```
bretambrose, line 1083: `s_handle_on_listener_success` maybe

source/darwin/nw_socket.c
```diff
@@ -0,0 +1,2058 @@
...
+                nw_connection_set_state_changed_handler(
+                    new_socket->io_handle.data.handle, ^(nw_connection_state_t state, nw_error_t error) {
+                      s_schedule_connection_state_changed_fn(
+                          new_socket, new_nw_socket, new_nw_socket->os_handle.nw_connection, state, error);
+                    });
+
+                // released when the connection state changed to nw_connection_state_cancelled
+                s_socket_acquire_internal_ref(new_nw_socket);
+
+                AWS_LOGF_TRACE(
```
bretambrose, line 1053: I feel like server connection establishment deserves more than Trace logging.  Also, it would be nice if we logged the fact this function was invoked at all.  Right now, we're only logging the success path.

source/darwin/nw_socket.c
```diff
@@ -0,0 +1,2058 @@
...
+    aws_mutex_lock(&nw_socket->synced_data.lock);
+    if (nw_socket->synced_data.base_socket && s_validate_event_loop(nw_socket->event_loop)) {
+        struct aws_task *task = aws_mem_calloc(nw_socket->allocator, 1, sizeof(struct aws_task));
+
+        struct nw_listener_connection_args *args =
+            aws_mem_calloc(nw_socket->allocator, 1, sizeof(struct nw_listener_connection_args));
+
+        args->nw_socket = nw_socket;
+        args->allocator = nw_socket->allocator;
+        args->error_code = error_code;
```
bretambrose, line 1098: Can we convert all Apple error codes to CRT error codes before sticking them in an integer?  Maybe log them first before converting?  This would apply everywhere we stick an Apple code into an int currently.  

If I see error_code in a struct, my expectation is that it's a CRT error code not an OS error code.

source/darwin/nw_socket.c
```diff
@@ -0,0 +1,2058 @@
...
+}
+
+static void s_schedule_write_fn(
+    struct nw_socket *nw_socket,
+    int error_code,
+    size_t bytes_written,
+    void *user_data,
+    aws_socket_on_write_completed_fn *written_fn) {
+    aws_mutex_lock(&nw_socket->synced_data.lock);
+    if (s_validate_event_loop(nw_socket->event_loop)) {
```
bretambrose, line 1139: Is it possible to have completed a write to a socket that does not have an event loop?  By the time we're reading/writing hasn't everything been set?

bretambrose (commented):
Another checkpoint on the overall review.  Beyond the comments, I'd love to see regular trace-level logging for every callback function if it does not already have it.  We want to be able to follow the functional flow in the logs.

include/aws/io/socket.h
```diff
@@ -239,12 +272,33 @@ AWS_IO_API int aws_socket_start_accept(
  */
 AWS_IO_API int aws_socket_stop_accept(struct aws_socket *socket);
 
+/**
+ * Apple Network Framework only. The callback that will triggered when aws_socket_close() finished. The callback
+ * will be called from the socket event loop.
+ */
+AWS_IO_API int aws_socket_set_close_complete_callback(
```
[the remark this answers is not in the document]
xiazhvera: We discussed this offline and syncing here for the record. Adding callbacks for close/cleanup is tricky because aws_socket_cleanup() might internally call aws_socket_close(), which might cause complicated senarios: 

1. Conflicts in callbacks: It would be a edge case where user provides different completion functions on call  aws_socket_cleanup() and aws_socket_close(). We need to define expected behavior in such cases.
2. It is a conditional call to aws_socket_close() in aws_socket_cleanup(). We’d have to check if aws_socket_close() was called during aws_socket_cleanup() to determine whether to wait for the callback or not. 

This creates messy logic paths, and we need to carefully handle these edge cases if we choose to setup the callbacks in close/cleanup API.

source/darwin/nw_socket.c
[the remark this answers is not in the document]
xiazhvera: To be clarify, the comment was about the following if-statement: 
```
if (nw_socket->synced_state.state < state || (state == LISTENING && nw_socket->synced_state.state == STOPPED)) 
```
In the following corner case the state could go backwards: 
After we called aws_socket_close(), the state is set to "CLOSING". And then when we have a read callback get invoked, it is possible that we got an ERROR back and trying to set the socket state to "ERROR". 
Well, we probably should fatal_assert the rest of the cases though.

tests/socket_test.c
```diff
@@ -357,35 +468,225 @@ static int s_test_socket_ex(
...
+    struct aws_socket outgoing;
+    ASSERT_SUCCESS(aws_socket_init(&outgoing, allocator, options));
+    ASSERT_SUCCESS(aws_socket_connect(&outgoing, endpoint, event_loop, s_local_outgoing_connection, &outgoing_args));
+
+    ASSERT_SUCCESS(aws_mutex_lock(&mutex));
+    ASSERT_SUCCESS(aws_condition_variable_wait_pred(
+        &condition_variable, &mutex, s_connection_completed_predicate, &outgoing_args));
+    ASSERT_SUCCESS(aws_mutex_unlock(&mutex));
+
+    aws_socket_subscribe_to_readable_events(&listener, s_on_readable, NULL);
```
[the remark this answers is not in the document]
xiazhvera: I was not able to reproduce this issue. I tested it by wrap the subscription function with `ASSERT_SUCCESS`. And it seems that `udp_socket_communication` test passed. It possible that it was fixed in previous version. 

```
ASSERT_SUCCESS(aws_socket_subscribe_to_readable_events(&listener, s_on_readable, NULL);)
```

source/darwin/nw_socket.c
```diff
@@ -0,0 +1,2047 @@
...
+
+    nw_connection_send(
+        socket->io_handle.data.handle, data, _nw_content_context_default_message, true, ^(nw_error_t error) {
+          int error_code = !error || nw_error_get_error_code(error) == 0
+                               ? AWS_OP_SUCCESS
+                               : s_determine_socket_error(nw_error_get_error_code(error));
+
+          if (error_code) {
+              nw_socket->last_error = error_code;
+              aws_raise_error(error_code);
```
bretambrose, line 2002: Probably not useful to raise error here in a callback

bretambrose (commented):
Not quite finished, but getting close.

Would suggest doing a pass on log levels and ensuring that they make sense.
Trace - most appropriate for data processing step and extremely common/spammy events
Debug/Info - appropriate for connection-level events, state changes, etc...

source/darwin/nw_socket.c
```diff
@@ -0,0 +1,2058 @@
...
+        s_schedule_on_connection_result(nw_socket, AWS_OP_SUCCESS);
+        s_schedule_next_read(nw_socket);
+        aws_ref_count_release(&nw_socket->external_ref_count);
+    } else if (state == nw_connection_state_waiting) {
+        AWS_LOGF_DEBUG(
+            AWS_LS_IO_SOCKET,
+            "id=%p handle=%p: socket connection is waiting for a usable network before re-attempting. ErrorCode: %d.",
+            (void *)nw_socket,
+            (void *)nw_socket->os_handle.nw_connection,
+            connection_args->error);
```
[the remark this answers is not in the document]
xiazhvera: So far I haven't see error code in preparing, while there are cases that we got error in waiting state (for example: timeout, invalid file path).

source/channel_bootstrap.c
```diff
@@ -1402,12 +1606,20 @@ void s_on_server_connection_result(
 
 error_cleanup:
     /* no channel is created */
-    connection_args->incoming_callback(connection_args->bootstrap, aws_last_error(), NULL, connection_args->user_data);
-
+    (void)socket; // to avoid expression error after a label
```
bretambrose, line 1609: can just use a semicolon

source/channel_bootstrap.c
```diff
@@ -1402,12 +1606,20 @@ void s_on_server_connection_result(
...
     struct aws_allocator *allocator = new_socket->allocator;
+
+    struct socket_shutdown_server_connection_result_args *close_args =
```
bretambrose, line 1612: I see this pattern everywhere.  Can we create some helper functions and use them instead?

bretambrose (commented):
Looked through channel bootstrap.

Would really like to see _new, _destroy functions for all the task-arg situations across the whole PR.

source/darwin/nw_socket.c
```diff
@@ -0,0 +1,2052 @@
...
+}
+
+struct read_queue_node {
+    struct aws_allocator *allocator;
+    dispatch_data_t received_data;
+    struct aws_linked_list_node node;
+    size_t current_offset;
+};
+
+static void s_destroy_read_queue_node(struct read_queue_node *node) {
```
sbSteveK, line 477: Trivial: naming convention on destroy functions. This should be renamed to `s_read_queue_node_destroy()`. Also, if we have a destroy function for `read_queue_node` we should probably set up its pair of `s_read_queue_node_new()` that is called when we are creating something that needs to be destroyed. It looks like there's only one place where this node is created but it feels right to have them both.

source/darwin/nw_socket.c
```diff
@@ -0,0 +1,2052 @@
...
+        struct aws_socket *socket = nw_socket->synced_data.base_socket;
+        if (socket) {
+            if (hostname != NULL) {
+                size_t hostname_len = strlen(hostname);
+                size_t buffer_size = AWS_ARRAY_SIZE(socket->local_endpoint.address);
+                size_t to_copy = aws_min_size(hostname_len, buffer_size);
+                memcpy(socket->local_endpoint.address, hostname, to_copy);
+                socket->local_endpoint.port = port;
+            }
+            nw_release(local_endpoint);
```
sbSteveK, line 892: The local_endpoint doesn't appear to be released if there's no socket. We also appear to not use the hostname and port members if there's no socket. We could probably move hostname and port into this if block and move the nw_release of local_endpoint outside of it unless it's being taken care of elsewhere.

source/darwin/nw_socket.c
```diff
@@ -0,0 +1,2052 @@
...
+        AWS_LOGF_ERROR(
+            AWS_LS_IO_SOCKET,
+            "id=%p handle=%p: socket connection get apple os error code: %d",
+            (void *)nw_socket,
+            (void *)nw_socket->os_handle.nw_connection,
+            connection_args->error);
+
+        int error_code = s_determine_socket_error(connection_args->error);
+        nw_socket->last_error = error_code;
+        aws_mutex_lock(&nw_socket->synced_data.lock);
```
sbSteveK, line 949: Trivial: There are a lot of places we use `aws_mutex_lock(&nw_socket->synced_data.lock)` despite having the helper functions `s_lock_socket_synced_data` and `s_unlock_socket_synced_data`.  We should either always use the helper functions or get rid of them entirely. A quick search shows 16 manual  calls on synced_data.lock and 8 of the helper function.

source/darwin/nw_socket.c
```diff
@@ -0,0 +1,2047 @@
...
+        "id=%p handle=%p: closing state %d",
+        (void *)socket,
+        socket->io_handle.data.handle,
+        socket->state);
+
+    if (nw_socket->synced_state.state < CLOSING) {
+        s_unlock_socket_state(nw_socket);
+        // We would like to keep CONNECTED_READ so that we could continue processing any received data until the we got
+        // the system callback indicates that the system connection has been closed in the receiving direction.
+        s_set_socket_state(nw_socket, nw_socket->synced_data.base_socket, CLOSING | CONNECTED_READ);
```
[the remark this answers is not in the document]
xiazhvera: Any example of a breaking scenario?

source/darwin/nw_socket.c
```diff
@@ -0,0 +1,2047 @@
...
+    }
+    s_unlock_socket_state(nw_socket);
+
+    /* fill in posix sock addr, and then let Network framework sort it out. */
+    size_t address_strlen;
+    if (aws_secure_strlen(remote_endpoint->address, AWS_ADDRESS_MAX_LEN, &address_strlen)) {
+        return AWS_OP_ERR;
+    }
+
+    struct socket_address address;
```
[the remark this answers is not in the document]
xiazhvera: I updated the union structure by adding the `sockaddr` as a member of the union. Would this help to eliminate the concern? 
```
    struct socket_address {
    union sock_addr_types {
        struct sockaddr_in addr_in;
        struct sockaddr_in6 addr_in6;
        struct sockaddr_un un_addr;
        struct sockaddr addr_base;
#    ifdef USE_VSOCK
        struct sockaddr_vm vm_addr;
#    endif
    } sock_addr_types;
};
```

source/darwin/nw_socket.c
```diff
@@ -0,0 +1,2047 @@
...
+
+    // Acquire nw_socket as we called nw_connection_receive, and the ref will be released when the handler is
+    // called.
+    s_socket_acquire_internal_ref(nw_socket);
+
+    /* read and let me know when you've done it. */
+    nw_connection_receive(
+        socket->io_handle.data.handle,
+        1,
+        UINT32_MAX,
```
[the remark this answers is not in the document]
xiazhvera: Any suggestion on a proper size? The API takes in uint32_t, therefore a value between [0, UINT32_MAX] would be valid.

include/aws/io/channel_bootstrap.h
```diff
@@ -296,6 +305,9 @@ AWS_IO_API int aws_server_bootstrap_set_alpn_callback(
 AWS_IO_API struct aws_socket *aws_server_bootstrap_new_socket_listener(
     const struct aws_server_socket_channel_bootstrap_options *bootstrap_options);
 
+AWS_IO_API struct aws_socket *aws_server_bootstrap_new_socket_listener_async(
```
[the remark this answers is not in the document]
xiazhvera: We could, but I dont think that's a good idea. For nw_socket, we had to keep event loop alive during the socket life time so that we could properly fire the async callback on shutdown. We would have to make the same change to the other socket to handle the async callbacks properly.

source/darwin/nw_socket.c
```diff
@@ -0,0 +1,2232 @@
...
+        AWS_LOGF_DEBUG(AWS_LS_IO_SOCKET, "id=%p: s_release_event_loop: socket has not event loop.", (void *)nw_socket);
+        return;
+    }
+    aws_event_loop_group_release(get_base_event_loop_group(nw_socket->event_loop));
+    AWS_LOGF_DEBUG(
+        AWS_LS_IO_SOCKET, "id=%p: s_release_event_loop: socket release event loop group.", (void *)nw_socket);
+    nw_socket->event_loop = NULL;
+}
+
+static void s_set_socket_state(struct nw_socket *nw_socket, struct aws_socket *socket, enum aws_nw_socket_state state) {
```
sbSteveK, line 315: Add comment to function that this needs to be wrapped in a synced data lock.

source/darwin/nw_socket.c
```diff
@@ -0,0 +1,2232 @@
...
+    if (s_setup_socket_params(nw_socket, options)) {
+        aws_mem_release(alloc, nw_socket);
+        return AWS_OP_ERR;
+    }
+
+    aws_mutex_init(&nw_socket->synced_data.lock);
+    aws_mutex_init(&nw_socket->base_socket_synced_data.lock);
+    nw_socket->base_socket_synced_data.base_socket = socket;
+
+    s_set_socket_state(nw_socket, socket, INIT);
```
sbSteveK, line 700: Since this is init(), we can safely set the socket state without calling this function.

source/darwin/nw_socket.c
```diff
@@ -0,0 +1,2232 @@
...
+
+        new_socket->io_handle.data.handle = task_args->new_connection;
+        new_socket->io_handle.set_queue = s_client_set_dispatch_queue;
+
+        struct nw_socket *new_nw_socket = new_socket->impl;
+        new_nw_socket->os_handle.nw_connection = task_args->new_connection;
+        new_nw_socket->connection_setup = true;
+
+        // Setup socket state to start read/write operations.
+        s_set_socket_state(new_nw_socket, new_socket, CONNECTED_READ | CONNECTED_WRITE);
```
sbSteveK, line 1122: Unsure if it's necessary but we are not holding the synced data lock at the point this is called.

source/darwin/nw_socket.c
```diff
@@ -0,0 +1,2232 @@
...
+    if (aws_secure_strlen(remote_endpoint->address, AWS_ADDRESS_MAX_LEN, &address_strlen)) {
+        s_unlock_socket_synced_data(nw_socket);
+        AWS_LOGF_DEBUG(
+            AWS_LS_IO_SOCKET,
+            "id=%p handle=%p: failed to parse address %s:%d.",
+            (void *)socket,
+            socket->io_handle.data.handle,
+            remote_endpoint->address,
+            (int)remote_endpoint->port);
+        return AWS_OP_ERR;
```
sbSteveK, line 1284: We should probably raise an error here in case anything is checking for last error on a failed operation.

source/channel_bootstrap.c
```diff
@@ -1522,18 +1807,207 @@ struct aws_socket *aws_server_bootstrap_new_socket_listener(
...
+        server_connection_args->setup_callback(
+            server_connection_args->bootstrap, error_code, server_connection_args->user_data);
+    }
+
+    s_server_connection_args_release(server_connection_args);
+
+    return;
+}
+
+struct aws_socket *aws_server_bootstrap_new_socket_listener_async(
```
sbSteveK, line 1877: Trivial: Apart from the requested_event_loop check and an acquire on the server_connection_args, this is identical to the non async function. Is there any way to combine the two and perform those two actions conditionally? We may even be able to do the acquire on the non-async and make sure we release it unless the refcount hitting 0 is important to the shutdown process of the non-async version of the function.

source/darwin/nw_socket.c
xiazhvera: Looking into it, We didn't lock here as we are in initializing process, no other process will touch the socket state. Updated the comment to explain it.

include/aws/io/channel_bootstrap.h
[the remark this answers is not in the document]
xiazhvera: Eventually, removed the aws_server_bootstrap_new_socket_listener_async. Now we added a `setup_callback` in `aws_server_socket_channel_bootstrap_options `. If setup_callback is set, the function will asynchronously fire the setup_callback when the listener is ready for use.

source/channel_bootstrap.c
xiazhvera: Eventually, I removed the aws_server_bootstrap_new_socket_listener_async. Now we have a `setup_callback` in `aws_server_socket_channel_bootstrap_options`. If `setup_callback` is set, it will asynchronously invoke the `setup_callback` when the listener is ready for use.

xiazhvera:
The main changes since last code review: 
1. Improve the locks. Now the socket has 2 locks, (a) lock for `base_socket` (to avoid bad access after `aws_socket_cleanup()` is called)  (b) lock for synced_data including `socket_state` and `read_scheduled`. 
2. Remove `aws_server_bootstrap_new_socket_listener_async` , now `aws_server_bootstrap_new_socket_listener`  will asynchronously invoke the `setup_callback` when the listener is ready for use, if `setup_callback` is set.

source/darwin/nw_socket.c
```diff
@@ -0,0 +1,2047 @@
...
+            return aws_raise_error(AWS_IO_SOCKET_CLOSED);
+        }
+
+        AWS_LOGF_DEBUG(
+            AWS_LS_IO_SOCKET,
+            "id=%p handle=%p: s_schedule_next_read : aws_socket_read() empty queue.",
+            (void *)nw_socket,
+            (void *)nw_socket->os_handle.nw_connection);
+
+        s_schedule_next_read(nw_socket);
```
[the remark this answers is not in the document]
xiazhvera: The initial "read" is triggered at 2 places. 1. user called aws_socket_read() 2. user called aws_socket_subscribe_to_readable_events().

README.md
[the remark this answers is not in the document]
xiazhvera: Updated the README.

source/darwin/nw_socket.c
xiazhvera: Tested on posix and nw_socket, the socket will send the keep alive packet every `keep_alive_interval_sec` seconds, and expecting ACK within `keep_alive_timeout_sec` seconds.

source/darwin/nw_socket.c
```diff
@@ -0,0 +1,2047 @@
...
+        (void *)socket,
+        socket->io_handle.data.handle,
+        (void *)event_loop);
+    return aws_raise_error(AWS_IO_EVENT_LOOP_ALREADY_ASSIGNED);
+}
+
+/* s_schedule_next_read() is called when we find that we've read all of our buffers or we preemptively know we're going
+ * to want more notifications. That read data gets queued into an internal read buffer to then be vended upon a call to
+ * aws_socket_read() */
+static void s_schedule_next_read(struct nw_socket *nw_socket) {
```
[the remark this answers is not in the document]
xiazhvera: Now the read task will be kick off at two places 1. aws_socket_read() 2. aws_socket_subscribe_to_readable_events(). And then read will be rescheduled at the nw_connection_receive completion handler callback when read completed. 
As the read task inited at two places, I still keep a `read_scheduled` flag to avoid duplicated read task.

source/darwin/nw_socket.c
```diff
@@ -1039,82 +1035,121 @@ static void s_process_listener_success_task(struct aws_task *task, void *args, e
     struct nw_listener_connection_args *task_args = args;
     struct aws_allocator *allocator = task_args->allocator;
     struct nw_socket *listener_nw_socket = task_args->nw_socket;
+    int error = task_args->error_code;
```
bretambrose, line 1077: `(void)status` can be removed.

source/darwin/nw_socket.c
```diff
@@ -1039,82 +1035,121 @@ static void s_process_listener_success_task(struct aws_task *task, void *args, e
...
 
-                if (hostname != NULL) {
-                    size_t hostname_len = strlen(hostname);
-                    size_t buffer_size = AWS_ARRAY_SIZE(new_socket->remote_endpoint.address);
-                    size_t to_copy = aws_min_size(hostname_len, buffer_size);
-                    memcpy(new_socket->remote_endpoint.address, hostname, to_copy);
-                    new_socket->remote_endpoint.port = port;
-                }
+        if (hostname != NULL) {
+            size_t address_strlen;
```
bretambrose, line 1109: We almost never want to use libc memory/string functions directly.  Let's just use aws_byte_cursor_from_c_str and make a local aws_byte_buf that refers to the memory and use an appropriate copy function.

source/darwin/nw_socket.c
```diff
@@ -1039,82 +1035,121 @@ static void s_process_listener_success_task(struct aws_task *task, void *args, e
...
-                } else // The connection is not sent to user, clean it up. The nw_connection should be released in
-                       // socket clean up.
-                {
-                    AWS_FATAL_ASSERT("The listener accept_result_fn should not be NULL.");
-                }
-            }
-            s_unlock_base_socket(listener_nw_socket);
+        // this internal ref will be released when the connection canceled ( connection state changed to
+        // nw_connection_state_cancelled)
+        s_socket_acquire_internal_ref(new_nw_socket);
```
bretambrose, line 1138: Can we take this ref before we set the state changed callback?  What would happen if the connection got canceled in between setting the callback and this line?

source/darwin/nw_socket.c
```diff
@@ -1824,9 +1875,17 @@ static void s_schedule_next_read(struct nw_socket *nw_socket) {
           s_unlock_socket_synced_data(nw_socket);
 
           bool complete = is_complete;
-          if (!error || nw_error_get_error_code(error) == 0) {
-              int error = AWS_ERROR_SUCCESS;
+          int nw_error_code = error ? nw_error_get_error_code(error) : 0;
```
bretambrose: Begging for this lambda to become a standalone function.

source/darwin/nw_socket.c
```diff
@@ -2044,19 +2093,26 @@ static int s_socket_write_fn(
 
     nw_connection_send(
         socket->io_handle.data.handle, data, _nw_content_context_default_message, true, ^(nw_error_t error) {
-          int error_code = !error || nw_error_get_error_code(error) == 0
-                               ? AWS_OP_SUCCESS
-                               : s_determine_socket_error(nw_error_get_error_code(error));
+          int nw_error_code = error ? nw_error_get_error_code(error) : 0;
```
bretambrose: Would prefer to see this become a helper function too.

source/darwin/nw_socket.c
```diff
@@ -893,101 +899,109 @@ static void s_process_connection_state_changed_task(struct aws_task *task, void
...
 
-            s_socket_release_internal_ref(nw_socket);
-        } break;
-        case nw_connection_state_ready: {
-            nw_path_t path = nw_connection_copy_current_path(nw_connection);
-            nw_endpoint_t local_endpoint = nw_path_copy_effective_local_endpoint(path);
-            nw_release(path);
-            const char *hostname = nw_endpoint_get_hostname(local_endpoint);
-            uint16_t port = nw_endpoint_get_port(local_endpoint);
+                s_lock_socket_synced_data(nw_socket);
```
bretambrose, line 928: Per discussion, the socket can go away in between pulling it out here and the call to s_set_socket_state.  Also, per discussion, let's try and remove the set to the aws_socket.state completely inside s_set_socket_state

source/darwin/nw_socket.c
```diff
@@ -1560,55 +1562,63 @@ static void s_process_listener_state_changed_task(struct aws_task *task, void *a
...
-            AWS_LS_IO_SOCKET, "id=%p handle=%p: listener cancelled ", (void *)nw_socket, (void *)nw_listener);
+            s_lock_base_socket(nw_socket);
+            s_lock_socket_synced_data(nw_socket);
+            s_set_socket_state(nw_socket, listener_state_changed_args->socket, ERROR);
+            s_unlock_socket_synced_data(nw_socket);
+            struct aws_socket *aws_socket = nw_socket->base_socket_synced_data.base_socket;
+            if (nw_socket->on_accept_started_fn) {
+                nw_socket->on_accept_started_fn(aws_socket, crt_error_code, nw_socket->listen_accept_started_user_data);
+            }
+            s_unlock_base_socket(nw_socket);
```
bretambrose: Is there any chance that something in the accept started callback could invoke something that tries to lock the socket base?

source/darwin/nw_socket.c
```diff
@@ -847,6 +847,8 @@ static void s_process_connection_result_task(struct aws_task *task, void *arg, e
     struct nw_socket_scheduled_task_args *task_args = arg;
     struct nw_socket *nw_socket = task_args->nw_socket;
 
+    AWS_LOGF_TRACE(AWS_LS_IO_SOCKET, "id=%p: start to process connection result task.", (void *)nw_socket);
```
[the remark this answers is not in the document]
xiazhvera: Updated the state_change log to debug. For this specific line, since the function is invoking the connection result callback, I think trace would be a good to track for this callback.

source/darwin/nw_socket.c
xiazhvera: The base socket is only modified(write) at two places:  
(1) socket creation/init : The lock is just inited, and would not hold at this time. 
(2) aws_socket_cleanup(): This is where the base socket get nulled out. We only lock the base socket if the function is not on the event loop to avoid dead lock. Use this callback as an example, if we call aws_socket_cleanup() in `on_accept_started_fn`, as it is on the event loop, we will not lock the base socket here.

source/posix/socket.c
```diff
@@ -1135,21 +1135,56 @@ static void s_socket_accept_event(
         socket->io_handle.data.fd);
 }
 
+struct on_start_accept_result_args {
```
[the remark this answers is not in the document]
xiazhvera: Moved it to the private header,  I'm not sure if that's better either though 🤔

source/darwin/nw_socket.c
xiazhvera: Not really sure what is this mean...? As the dest address is a char array, I don't see how using a local aws_byte_buf would help. Am I missing something?

xiazhvera: Updated to use aws_byte_buf, while I'm not sure if that is necessary... 

```            
            struct aws_byte_buf hostname_buf = aws_byte_buf_from_c_str(hostname);
            struct aws_byte_buf address_buf = aws_byte_buf_from_empty_array(new_socket->remote_endpoint.address, AWS_ADDRESS_MAX_LEN);
            aws_byte_buf_write_from_whole_buffer(&address_buf, hostname_buf);
            aws_byte_buf_clean_up(&address_buf);
            aws_byte_buf_clean_up(&hostname_buf);
```

bretambrose (approved):

PR #664, opened 2024-07-30
Use const pointers in `secure_channel_tls_handler.c`.

teo-tsirpanis:
*Issue #, if available:*

*Description of changes:*

This PR uses `PCCERT_***` types in two cases, which are aliases for const pointers.

Fixes build failures on MinGW.

### DIFF
--- a/source/windows/secure_channel_tls_handler.c
+++ b/source/windows/secure_channel_tls_handler.c
@@ -53,7 +53,7 @@ struct secure_channel_ctx {
     struct aws_tls_ctx ctx;
     struct aws_string *alpn_list;
     SCHANNEL_CRED credentials;
-    PCERT_CONTEXT pcerts;
+    PCCERT_CONTEXT pcerts;
     HCERTSTORE cert_store;
     HCERTSTORE custom_trust_store;
     HCRYPTPROV crypto_provider;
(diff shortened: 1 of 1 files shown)


waahm7:
Unfortunately, at the moment we don't support MinGW. What is your use-case? Is this the only problem you are facing with MinGW, or is this just the first of N problems?

teo-tsirpanis:
I discovered this while working updating dependencies (TileDB-Inc/TileDB#5010), and after applying this patch `aws-c-io` builds in MinGW[^1]. Before making the patch, I tried to build without this warning and without any warnings, but for some reason this warning could not be suppressed and always escalated to an error.

I think this change is good regardless of the MinGW situation, because it matches the constness of the pointers with the API signatures ([example 1](https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-certfreecertificatecontext), [example 2](https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-certgetcertificatechain)).

[^1]: There was another error in `aws-sdk-cpp` which needs more thought on if and how to fix upstream.

waahm7:
Thanks, LGTM!

waahm7:
Released in https://github.com/awslabs/aws-c-io/releases/tag/v0.14.16.

PR #669, opened 2024-08-27
Provide I/O operation status back to event loop

sfod:
To support QNX, the event loop should be implemented based on `ionotify` function. `ionotify` is similar to `epoll` and `kqueue` functions with a couple of differences. One of these differences is `ionotify` needs to know the result of I/O operation performed on the fd (e.g. `EWOULDBLOCK` or `EAGAIN`).

In the current implementation of the event loop, I/O event callbacks (`aws_event_loop_on_event_fn`) do not return results of the I/O operations.

This PR adds feedback functionality to the socket channel.

### DIFF
--- a/include/aws/io/event_loop.h
+++ b/include/aws/io/event_loop.h
@@ -25,6 +25,7 @@ enum aws_io_event_type {
 struct aws_event_loop;
 struct aws_task;
 struct aws_thread_options;
+struct aws_event_loop_io_op_result;
 
 #if AWS_USE_IO_COMPLETION_PORTS
 
@@ -99,6 +100,10 @@ struct aws_event_loop_vtable {
         void *user_data);
 #endif
     int (*unsubscribe_from_io_events)(struct aws_event_loop *event_loop, struct aws_io_handle *handle);
+    void (*feedback_io_result)(
+        struct aws_event_loop *event_loop,
+        struct aws_io_handle *handle,
+        const struct aws_event_loop_io_op_result *io_op_result);
     void (*free_io_event_resources)(void *user_data);
     bool (*is_on_callers_thread)(struct aws_event_loop *event_loop);
 };
@@ -138,6 +143,11 @@ struct aws_event_loop_group {
     struct aws_array_list event_loops;
     struct aws_ref_count ref_count;
     struct aws_shutdown_callback_options shutdown_options;
+};
+
+struct aws_event_loop_io_op_result {
+    size_t read_bytes;
+    int error_code;
 };
 
 AWS_EXTERN_C_BEGIN
@@ -365,6 +375,15 @@ int aws_event_loop_subscribe_to_io_events(
  */
 AWS_IO_API
 int aws_event_loop_unsubscribe_from_io_events(struct aws_event_loop *event_loop, struct aws_io_handle *handle);
+
+/**
+ * Update the I/O operation completion status on the given I/O handle.
+ */
+AWS_IO_API
+void aws_event_loop_feedback_io_op_result(
+    struct aws_event_loop *event_loop,
+    struct aws_io_handle *handle,
+    const struct aws_event_loop_io_op_result *io_op_result);
 
 /**
  * Cleans up resources (user_data) associated with the I/O eventing subsystem for a given handle. This should only
--- a/source/bsd/kqueue_event_loop.c
+++ b/source/bsd/kqueue_event_loop.c
@@ -39,6 +39,10 @@ static int s_subscribe_to_io_events(
     aws_event_loop_on_event_fn *on_event,
     void *user_data);
 static int s_unsubscribe_from_io_events(struct aws_event_loop *event_loop, struct aws_io_handle *handle);
+static void s_feedback_io_result(
+    struct aws_event_loop *event_loop,
+    struct aws_io_handle *handle,
+    const struct aws_event_loop_io_op_result *io_op_result);
 static void s_free_io_event_resources(void *user_data);
 static bool s_is_event_thread(struct aws_event_loop *event_loop);
 
@@ -110,6 +114,8 @@ struct handle_data {
 
     struct aws_task subscribe_task;
     struct aws_task cleanup_task;
+
+    struct aws_event_loop_io_op_result last_io_operation_result;
 };
 
 enum {
@@ -127,6 +133,7 @@ struct aws_event_loop_vtable s_kqueue_vtable = {
     .subscribe_to_io_events = s_subscribe_to_io_events,
     .cancel_task = s_cancel_task,
     .unsubscribe_from_io_events = s_unsubscribe_from_io_events,
+    .feedback_io_result = s_feedback_io_result,
     .free_io_event_resources = s_free_io_event_resources,
     .is_on_callers_thread = s_is_event_thread,
 };
@@ -135,6 +142,7 @@ struct aws_event_loop *aws_event_loop_new_default_with_options(
     struct aws_allocator *alloc,
     const struct aws_event_loop_options *options) {
     AWS_ASSERT(alloc);
+    // FIXME Remove this assert.
     AWS_ASSERT(clock);
     AWS_ASSERT(options);
     AWS_ASSERT(options->clock);
@@ -725,6 +733,23 @@ static int s_unsubscribe_from_io_events(struct aws_event_loop *event_loop, struc
     return AWS_OP_SUCCESS;
 }
 
+static void s_feedback_io_result(
+    struct aws_event_loop *event_loop,
+    struct aws_io_handle *handle,
+    const struct aws_event_loop_io_op_result *io_op_result) {
+    AWS_ASSERT(handle->additional_data);
+    struct handle_data *handle_data = handle->additional_data;
+    AWS_ASSERT(event_loop == handle_data->event_loop);
+    AWS_LOGF_TRACE(
+        AWS_LS_IO_EVENT_LOOP,
+        "id=%p: got feedback on I/O operation for fd %d: status %s",
+        (void *)event_loop,
+        handle->data.fd,
+        aws_error_str(io_op_result->error_code));
+    handle_data->last_io_operation_result.read_bytes = io_op_result->read_bytes;
+    handle_data->last_io_operation_result.error_code = io_op_result->error_code;
+}
+
 static bool s_is_event_thread(struct aws_event_loop *event_loop) {
     struct kqueue_loop *impl = event_loop->impl_data;
 
@@ -931,6 +956,13 @@ static void aws_event_loop_thread(void *user_data) {
                     handle_data->owner->data.fd);
                 handle_data->on_event(
                     event_loop, handle_data->owner, handle_data->events_this_loop, handle_data->on_event_user_data);
+                AWS_LOGF_INFO(
+                    AWS_LS_IO_EVENT_LOOP,
+                    "id=%p: on_event completion status is %d (%s); read %lu bytes",
+                    (void *)event_loop,
+                    handle_data->last_io_operation_result.error_code,
+                    aws_error_str(handle_data->last_io_operation_result.error_code),
+                    handle_data->last_io_operation_result.read_bytes);
             }
 
             handle_data->events_this_loop = 0;
--- a/source/event_loop.c
+++ b/source/event_loop.c
@@ -514,6 +514,15 @@ int aws_event_loop_unsubscribe_from_io_events(struct aws_event_loop *event_loop,
     return event_loop->vtable->unsubscribe_from_io_events(event_loop, handle);
 }
 
+void aws_event_loop_feedback_io_op_result(
+    struct aws_event_loop *event_loop,
+    struct aws_io_handle *handle,
+    const struct aws_event_loop_io_op_result *io_op_result) {
+    if (event_loop->vtable->feedback_io_result) {
+        event_loop->vtable->feedback_io_result(event_loop, handle, io_op_result);
+    }
+}
+
 void aws_event_loop_free_io_event_resources(struct aws_event_loop *event_loop, struct aws_io_handle *handle) {
     AWS_ASSERT(event_loop && event_loop->vtable->free_io_event_resources);
     event_loop->vtable->free_io_event_resources(handle->additional_data);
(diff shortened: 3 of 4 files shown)


bretambrose:
That's a lot of conditional compilation.  Is that intended to be final?

sfod:
Short answer: it's totally possible to completely remove all these compile-time conditions from existing sources.

Now, some details.
- All changes in the `source/posix/` dir can be easily moved to the new `source/qnx/` dir. This way, we'll have two almost identical posix implementations - one with events with feedback, one with no feedback.

- `socket_channel_handler.c` could also be split into two versions easily by moving it into `source/something/` dir.

- As for  `io.h`. It's a public header, so it's trickier. Some cmake magic is definitely can be done to use different headers depending on the platform, but I didn't look into it yet.

To be honest, I want to keep all these checks, because otherwise every fix to posix implementation or socket_channel should be duplicated in the qnx implementation.

Apart from this, do you think it's a viable solution to add feedback functionality to I/O events?

include/aws/io/io.h
```diff
@@ -14,12 +14,36 @@ AWS_PUSH_SANE_WARNING_LEVEL
...
+#if AWS_USE_ON_EVENT_WITH_RESULT
+struct aws_event_loop;
+
+/**
+ * Results of the I/O operation(s) performed on the aws_io_handle.
+ */
+struct aws_io_handle_io_op_result {
+    size_t read_bytes;
+    size_t written_bytes;
+    int read_error_code;
```
bretambrose, line 28: What is the domain here?  crt error codes or os error codes?

source/posix/pipe.c
```diff
@@ -278,6 +278,14 @@ int aws_pipe_read(struct aws_pipe_read_end *read_end, struct aws_byte_buf *dst_b
     if (read_val < 0) {
         int errno_value = errno; /* Always cache errno before potential side-effect */
         if (errno_value == EAGAIN || errno_value == EWOULDBLOCK) {
+#if AWS_USE_ON_EVENT_WITH_RESULT
+            if (read_impl->handle.update_io_result) {
+                struct aws_io_handle_io_op_result io_op_result;
+                memset(&io_op_result, 0, sizeof(struct aws_io_handle_io_op_result));
```
bretambrose, line 284: prefer AWS_ZERO_STRUCT

source/posix/pipe.c
```diff
@@ -454,6 +462,15 @@ static void s_write_end_process_requests(struct aws_pipe_write_end *write_end) {
                 if (errno_value == EAGAIN || errno_value == EWOULDBLOCK) {
                     /* The pipe is no longer writable. Bail out */
                     write_impl->is_writable = false;
+
+#if AWS_USE_ON_EVENT_WITH_RESULT
+                    struct aws_io_handle_io_op_result io_op_result;
```
bretambrose, line 467: Make conditional on non-null `write_impl->handle.update_io_result`?

source/posix/socket.c
```diff
@@ -476,6 +476,15 @@ static void s_socket_connect_event(
                 "id=%p fd=%d: spurious event, waiting for another notification.",
                 (void *)socket_args->socket,
                 handle->data.fd);
+
+#if AWS_USE_ON_EVENT_WITH_RESULT
+            struct aws_io_handle_io_op_result io_op_result;
```
bretambrose, line 481: Make conditional on callback non-null

include/aws/io/io.h
sfod: Yes, this field is a CRT error code (e.g. `AWS_IO_READ_WOULD_BLOCK`). I can rename it to `read_crt_error_code`, wdyt?

bretambrose: Probably fine as is.

bretambrose (changes requested):
Some quick Initial feedback:

The conditional compilation should be removed from the headers and source.  We dont have any backwards compatibility guarantees with the C libraries across different versions.

Additionally, I'm really concerned about the massive duplication between the qnx and posix folders from a maintainability standpoint.

source/qnx/ionotify_event_loop.c
```diff
@@ -0,0 +1,1042 @@
...
+static void s_aws_ionotify_cleanup_aws_lc_thread_local_state(void *user_data) {
+    (void)user_data;
+    aws_cal_thread_clean_up();
+}
+
+static void s_process_pulse(
+    struct aws_event_loop *event_loop,
+    const struct _pulse *pulse,
+    bool *should_process_cross_thread_tasks) {
+    if (pulse->code == CROSS_THREAD_PULSE_SIGEV_CODE) {
```
bretambrose, line 875: This feels like maybe it should be up in the calling code instead, removing the awkward output param.

source/qnx/ionotify_event_loop.c
```diff
@@ -0,0 +1,1042 @@
...
+            (void *)event_loop,
+            handle_id);
+        return;
+    }
+
+    if (!ionotify_event_data->is_subscribed) {
+        return;
+    }
+
+    AWS_LOGF_TRACE(
```
bretambrose, line 907: Let's add the fd to all the logging after this point too.

source/qnx/ionotify_event_loop.c
```diff
@@ -0,0 +1,1042 @@
...
+    }
+    if (pulse->value.sival_int & _NOTIFY_COND_INPUT) {
+        AWS_LOGF_TRACE(AWS_LS_IO_EVENT_LOOP, "id=%p: fd is readable", (void *)event_loop);
+        event_mask |= AWS_IO_EVENT_TYPE_READABLE;
+    }
+    if (pulse->value.sival_int & _NOTIFY_COND_OUTPUT) {
+        AWS_LOGF_TRACE(AWS_LS_IO_EVENT_LOOP, "id=%p: fd is writable", (void *)event_loop);
+        event_mask |= AWS_IO_EVENT_TYPE_WRITABLE;
+    }
+    if (pulse->value.sival_int & _NOTIFY_COND_EXTEN) {
```
bretambrose, line 926: Under what circumstances would this happen and what does it mean?  All we seem to do is log it.

source/qnx/ionotify_event_loop.c
```diff
@@ -0,0 +1,1042 @@
...
+        goto clean_up_thread;
+    }
+    AWS_LOGF_DEBUG(
+        AWS_LS_IO_EVENT_LOOP,
+        "id=%p: Opened QNX channel with ID %d",
+        (void *)event_loop,
+        ionotify_loop->io_events_channel_id);
+
+    /* Open connection over the QNX channel for pulses. */
+    ionotify_loop->pulse_connection_id = ConnectAttach(0, 0, ionotify_loop->io_events_channel_id, _NTO_SIDE_CHANNEL, 0);
```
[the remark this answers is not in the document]
sfod: I couldn't find a simple answer to this in QNX docs, so added `ConnectDetach`.

source/qnx/ionotify_event_loop.c
```diff
@@ -0,0 +1,1042 @@
...
+        (unsigned long long)run_at_nanos);
+    task->timestamp = run_at_nanos;
+
+    aws_mutex_lock(&ionotify_loop->task_pre_queue_mutex);
+    bool is_first_task = aws_linked_list_empty(&ionotify_loop->task_pre_queue);
+    aws_linked_list_push_back(&ionotify_loop->task_pre_queue, &task->node);
+    aws_mutex_unlock(&ionotify_loop->task_pre_queue_mutex);
+
+    /* If the list was not empty, we already sent a cross-thread pulse. No need to send it again. */
+    if (is_first_task) {
```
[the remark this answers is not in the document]
sfod: The original version of this part in the epoll event loop looked like this:
```C
aws_mutex_lock(mutex);
if (aws_linked_list_empty(queue)) {
    write(notifying_handle, ...);
}
aws_linked_list_push_back(queue, task);
aws_mutex_unlock(mutex);
```
After series of changes, `is_first_task` was introduced, but the locking logic wasn't adjusted to it.
I can fix this in epoll event loop (and maybe others?) in a separate PR.

source/qnx/ionotify_event_loop.c
```diff
@@ -0,0 +1,1042 @@
...
+}
+
+/* This callback is called by I/O operations to notify about their results. */
+static void s_update_io_result(
+    struct aws_event_loop *event_loop,
+    struct aws_io_handle *handle,
+    const struct aws_io_handle_io_op_result *io_op_result) {
+
+    if (!s_is_on_callers_thread(event_loop)) {
+        /* Move processing I/O operation results to the epoll thread if the operation is performed in another thread.*/
```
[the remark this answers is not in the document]
sfod: Ah, it's not even a copy-paste error, just a brain fart where I used epoll instead of event loop.

source/qnx/ionotify_event_loop.c
```diff
@@ -0,0 +1,1042 @@
...
+        return AWS_OP_ERR;
+    }
+
+    struct aws_hash_element *elem = NULL;
+    int next_handle_id = ionotify_loop->last_handle_id;
+    int was_created = 0;
+    do {
+        ++next_handle_id;
+        if (next_handle_id > max_handle_id) {
+            next_handle_id = 1;
```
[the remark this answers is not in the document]
sfod: Yes, I decided not to use 0 as a valid `handle_id` because
1. This handle is passed with the events from QNX resource managers, and I want (probably, irrationally) to have at least 1 bit set. In my head, it helps catch a case where a resource manager forgets to return user-provided data.
2. `s_subscribe_task()` is used both for subscribing and resubscribing to the I/O events. `0` helps with distinguishing these two cases. But this point was just a follow-up on the already made decision.

source/qnx/ionotify_event_loop.c
```diff
@@ -0,0 +1,1042 @@
...
+    event_loop->impl_data = ionotify_loop;
+    event_loop->vtable = &s_vtable;
+
+    if (aws_hash_table_init(&ionotify_loop->handles, alloc, 32, aws_hash_ptr, aws_ptr_eq, NULL, NULL)) {
+        goto clean_up_thread;
+    }
+
+    return event_loop;
+
+clean_up_thread:
```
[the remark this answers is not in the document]
sfod: Thanks for the very detailed comment gicing insights about suggested changes. I've implemented a single destroy function, and yeah, it's much nicer this way.

source/qnx/ionotify_event_loop.c
sfod: Oh, this one is tough. I added a comment describing my interaction with this flag. I suggest we keep this log, maybe it'll reveal something.

source/qnx/ionotify_event_loop.c
```diff
@@ -0,0 +1,1042 @@
...
+    /* Setup channel to receive events from resource managers. */
+    ionotify_loop->io_events_channel_id = ChannelCreate(0);
+    int errno_value = errno; /* Always cache errno before potential side-effect */
+    if (ionotify_loop->io_events_channel_id == -1) {
+        AWS_LOGF_ERROR(
+            AWS_LS_IO_EVENT_LOOP,
+            "id=%p: ChannelCreate failed with errno %d (%s)\n",
+            (void *)event_loop,
+            errno_value,
+            strerror(errno_value));
```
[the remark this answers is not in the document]
sfod: You mean convert errno to crt error and then print crt error string representations?
Considering that in most cases when I got errors I had to lookup errno in the docs anyway to understand why it might be returned by a particular function, I think it's not worth to implement a conversion function. I removed `strerror` for now and kept `errno`, maybe I'll revisit this later on.

source/qnx/ionotify_event_loop.c
```diff
@@ -0,0 +1,1042 @@
...
+
+    aws_mem_release(event_loop->alloc, ionotify_loop);
+    aws_event_loop_clean_up_base(event_loop);
+    aws_mem_release(event_loop->alloc, event_loop);
+}
+
+static int s_run(struct aws_event_loop *event_loop) {
+    struct ionotify_loop *ionotify_loop = event_loop->impl_data;
+
+    AWS_LOGF_INFO(AWS_LS_IO_EVENT_LOOP, "id=%p: Starting event-loop thread.", (void *)event_loop);
```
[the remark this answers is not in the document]
sfod: I can actually introduce a new subject, `AWS_LS_IO_EVENT_LOOP_IONOTIFY - "event-loop-ionotify"`. It won't take much effort. Unless you have something else in mind, then I'll leave it as is.

source/qnx/ionotify_event_loop.c
```diff
@@ -0,0 +1,1042 @@
...
+    }
+
+    /* Rearm resource manager. */
+    if (event_types != 0) {
+        AWS_LOGF_TRACE(
+            AWS_LS_IO_EVENT_LOOP, "id=%p: Got EWOULDBLOCK for fd %d, rearming it", (void *)event_loop, handle->data.fd);
+        /* We're on the event loop thread, just schedule subscribing task. */
+        ionotify_event_data->events_subscribed = event_types;
+        struct ionotify_loop *ionotify_loop = event_loop->impl_data;
+        aws_task_scheduler_cancel_task(&ionotify_loop->scheduler, &ionotify_event_data->resubscribe_task);
```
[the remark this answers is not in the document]
sfod: I split subscribing and resubscribing, so it should be slightly cleaner. Cancelling a previously scheduled task is still needed, but events from the canceled task are being added to a newly added task, so this case is handled.

source/qnx/ionotify_event_loop.c
```diff
@@ -0,0 +1,1042 @@
...
+         * https://www.qnx.com/developers/docs/8.0/com.qnx.doc.neutrino.lib_ref/topic/i/ionotify.html
+         * https://www.qnx.com/developers/docs/8.0/com.qnx.doc.neutrino.lib_ref/topic/m/msgregisterevent.html
+         *
+         * It's enough to register an event only once and then reuse it on followup ionotify rearming calls.
+         * NOTE: If you create a new sigevent for the same file descriptor, with the same flags, you HAVE to register
+         * it again. */
+        MsgRegisterEvent(&ionotify_event_data->event, ionotify_event_data->handle->data.fd);
+    } else if (!ionotify_event_data->is_subscribed) {
+        /* This is a resubscribing task, but unsubscribe happened, so ignore it. */
+        return;
```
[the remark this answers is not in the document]
sfod: Moved this logic to a separate task/function.

PR #676, opened 2024-09-25
Tls1.3 win update

bretambrose:
Support TLS 1.3 on Windows when mTLS is used.

Other changes:
- Use persistent key container for RSA keys.
- Implement local TLS server for tests.


tests/tls_handler_test.c
```diff
@@ -1465,9 +1600,35 @@ static int s_tls_client_channel_negotiation_success_ecc384_fn(struct aws_allocat
...
+    s_tls_client_channel_negotiation_success_ecc384_SCHANNEL_CREDS_fn)
+#    endif
+
+AWS_STATIC_STRING_FROM_LITERAL(s_aws_ecc384_host_name, "127.0.0.1");
+static int s_tls_client_channel_negotiation_success_tls1_3_fn(struct aws_allocator *allocator, void *ctx) {
+    (void)ctx;
+    return s_verify_good_host_mqtt_connect(allocator, s_aws_ecc384_host_name, 59443, NULL);
+}
+
+AWS_TEST_CASE(tls_client_channel_negotiation_success_tls1_3, s_tls_client_channel_negotiation_success_tls1_3_fn)
```
sbSteveK, line 1630: Trivial: I think this is testing MTLS so it should probably reflect that in the test name.

source/windows/secure_channel_tls_handler.c
```diff
@@ -124,6 +142,49 @@ static size_t s_message_overhead(struct aws_channel_handler *handler) {
     return sc_handler->stream_sizes.cbTrailer + sc_handler->stream_sizes.cbHeader;
 }
 
+/* Checks whether current system is running Windows 10 version `build_number` or later. This check is used
+   to determin availability of TLS 1.3. This will continue to be a valid check in Windows 11 and later as the
+   build number continues to increment upwards. e.g. Windows 11 starts at version 21H2 (build 22_000) */
+static bool s_is_windows_equal_or_above_version(DWORD build_number) {
```
sbSteveK, line 148: Trivial: If this function is only used to check against this one specific build and in one place, do we need an argument to be provided instead of just checking against WINDOWS_BUILD_1809? 

Also, the log feels a bit misleading. They all imply that the system is windows 10. The fail states should state that we're "Running windows version before build %u" and not "Running earlier windows 10 version before build %u". We can state "Running windows version before windows 10 build %u" if we know we are always checking against WINDOWS_BUILD_1809 which will always be windows 10.

source/windows/secure_channel_tls_handler.c
```diff
@@ -1021,19 +1089,27 @@ static int s_do_client_side_negotiation_step_2(struct aws_channel_handler *handl
...
             status = QueryContextAttributes(&sc_handler->sec_handle, SECPKG_ATTR_APPLICATION_PROTOCOL, &alpn_result);
 
-            if (status == SEC_E_OK && alpn_result.ProtoNegoStatus == SecApplicationProtocolNegotiationStatus_Success) {
-                aws_byte_buf_init(&sc_handler->protocol, handler->alloc, alpn_result.ProtocolIdSize + 1);
-                memset(sc_handler->protocol.buffer, 0, alpn_result.ProtocolIdSize + 1);
-                memcpy(sc_handler->protocol.buffer, alpn_result.ProtocolId, alpn_result.ProtocolIdSize);
-                sc_handler->protocol.len = alpn_result.ProtocolIdSize;
-                AWS_LOGF_DEBUG(
-                    AWS_LS_IO_TLS, "id=%p: Negotiated protocol %s", handler, (char *)sc_handler->protocol.buffer);
+            if (status == SEC_E_OK) {
```
sbSteveK, line 1092: status gets used in different ways throughout this process and at points is reassigned in ways where this status could be the result of different operations. We might want to create separate status variables for the different assignments so we can be explicit about which status result is being processed.

sbSteveK (approved):
Looks good but I think we should probably make the change to status to be clearer about what's going on and what result is being checked. It doesn't appear like we'd run into a fall-through where the result of a previous status is being inappropriately checked but I'm not familiar enough with Windows TLS function to be certain. I'll leave that discretion up to you.

sfod:
I converted this PR to draft, so it's not accidentally merged before it gets an approval from security

source/windows/secure_channel_tls_handler.c
sfod: We have another similar check for ALPN (see a [comment from Bret](https://github.com/awslabs/aws-c-io/pull/641#discussion_r1610568909) in the previous PR). That's where this arg came from.
But ALPN check is quite different, so it's definitely shouldn't be changed here, as this PR is already quite big.

I'll fix the log message, but will keep the arg, in case we'll decide to reuse this function in ALPN check.

source/windows/secure_channel_tls_handler.c
sfod: Good catch, it indeed looks suspicious

PR #680, opened 2024-10-04
Use s2n_cleanup_thread()

graebm:
See: https://github.com/aws/s2n-tls/pull/4584

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #682, opened 2024-10-10
Use s2n_cleanup_thread() and path

TingDaoK:
See: https://github.com/aws/s2n-tls/pull/4584

follow up for https://github.com/awslabs/aws-c-io/pull/680
- Tested on java with insource build [here](https://github.com/awslabs/aws-crt-java/pull/835)
- Tested on Python with prebuild-s2n [here](https://github.com/awslabs/aws-crt-python/pull/602)
- Also tested the nodejs issue locally, no crash.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -155,7 +155,10 @@ if (USE_S2N)
     file(GLOB AWS_IO_TLS_SRC
             "source/s2n/*.c"
             )
-    aws_use_package(s2n)
+    # NOTE: using find_package() (instead of aws_use_package() which may do an IN_SOURCE_BUILD).
+    # We do this because IN_SOURCE_BUILDS of S2N have different paths to the <s2n/unstable/*.h> headers.
+    find_package(s2n REQUIRED)
+    list(APPEND DEP_AWS_LIBS AWS::s2n)
 endif()
 
 file(GLOB IO_HEADERS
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -152,10 +152,22 @@ if (BYO_CRYPTO)
...
+        if (NOT s2n_FOUND)
+                # If s2n was not found, try to do the IN_SOURCE_BUILD.
+                # If not IN_SOURCE_BUILD, s2n will be required to be found and fail from aws_use_package
+                # If in source build, set the flag to use the insource path to the <s2n/unstable/*.h> headers.
+                aws_use_package(s2n)
+                add_definitions(-DAWS_S2N_INSOURCE_PATH)
+        else()
+                # Found s2n, use the install path
+                list(APPEND DEP_AWS_LIBS AWS::s2n)
+        endif()
```
graebm, line 170: trivial: `if true else false` is simpler than: `if not true else true`

```suggestion
        # Prefer find_package() because it's the normal CMake way to do dependencies.
        # But fall back on aws_use_package() because some projects still need to do an IN_SOURCE_BUILD of S2N.
        # (e.g. aws-crt-java until this is resolved: https://github.com/awslabs/aws-crt-java/pull/817)
        if (s2n_FOUND)
            list(APPEND DEP_AWS_LIBS AWS::s2n)
        else()
            # Set flag to use in-source path to  <s2n/unstable/*.h> headers if we do an IN_SOURCE_BUILD.
            aws_use_package(s2n)
            add_definitions(-DAWS_S2N_INSOURCE_PATH)
        endif()
```

graebm (approved):

PR #683, opened 2024-10-11
Handle PKCS#8 private keys in Windows

sfod:
*Issue #, if available:*

Currently, Windows implementation of PKI utils cannot decode private keys stored in the [PKCS#8](https://en.wikipedia.org/wiki/PKCS_8) format.

*Description of changes:*

PKCS#8 can be encrypted and unencrypted. This PR adds support for unencrypted PKCS#8 format only.

### DIFF
--- a/source/windows/windows_pki_utils.c
+++ b/source/windows/windows_pki_utils.c
@@ -678,7 +678,6 @@ int aws_import_key_pair_to_cert_context(
                     &decoded_len)) {
                 cert_type = AWS_CT_X509_RSA;
             }
-            LocalFree(key_wrapper);
         }
 #ifndef AWS_SUPPORT_WIN7
         else if (CryptDecodeObjectEx(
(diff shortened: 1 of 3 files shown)


source/windows/windows_pki_utils.c
```diff
@@ -640,6 +640,8 @@ int aws_import_key_pair_to_cert_context(
 
     struct aws_pem_object *private_key_ptr = NULL;
     DWORD decoded_len = 0;
+    BYTE *key_wrapper = NULL;
```
sbSteveK, line 643: I think we clean up key with LocalFree(key) down below. Do we know whether we need to also free this the same way whether in the event of a success or failure?

sbSteveK (approved):
See comment regarding potential memory leak. Besides that, looks solid.

source/windows/windows_pki_utils.c
sfod: Docs says nothing about this. I checked locally - `key` is NULL in case of failure (which is expected). `LocalFree` does nothing when its parameter is NULL, so it's safe to call it after all attempts to decode the key.
But I decided to move `LocalFree(key_wrapper)` to the clean_up section, so if we add another case with key_wrapper, its destroying will be already handled.

PR #691, opened 2024-11-05
Event loop public api

bretambrose:
* Move unnecessarily public types in event_loop.h to a private header
* Refactor event loop group creation to a single constructor function

Let's see what breaks here.  Downstream is obviously hosed.


include/aws/io/event_loop.h
```diff
@@ -6,286 +6,50 @@
...
-
-struct aws_event_loop_group {
-    struct aws_allocator *allocator;
-    struct aws_array_list event_loops;
-    struct aws_ref_count ref_count;
-    struct aws_shutdown_callback_options shutdown_options;
+    /**
+     * Optional configuration to control how the event loop group's threads bind to CPU groups
+     */
+    struct aws_event_loop_group_pin_options *pin_options;
```
graebm, line 48: trivial: not a fan of the `aws_event_loop_group_pin_options`, "sub-struct" that contains one int. Binding an additional AwsEventLoopGroupPinOptions class to every language seems like unnecessary work, and unnecessary indirection for users.
```suggestion
    const uint16_t *pin_cpu_group;
```

[Here's an example](https://github.com/awslabs/aws-c-io/blob/e247ef89cdf170cadcb4c665fb0146e82cd6e135/include/aws/io/tls_channel_handler.h#L450-L456) where an optional int was passed by pointer to an options struct, and it worked fine in the bindings ([python](https://github.com/awslabs/aws-crt-python/blob/084e141e9b87aea833b53e5b3a9175320218ed57/awscrt/io.py#L396-L397), [java](https://github.com/awslabs/aws-crt-java/blob/d76fdba766f5e8894be64742d26390ab7223f9f8/src/main/java/software/amazon/awssdk/crt/io/TlsContextPkcs11Options.java#L53))

include/aws/io/event_loop.h
```diff
@@ -6,286 +6,50 @@
...
 
-struct aws_event_loop_local_object {
-    const void *key;
-    void *object;
-    aws_event_loop_on_local_object_removed_fn *on_object_removed;
-};
+    /**
+     * Optional callback to invoke when the event loop group finishes destruction.
+     */
+    struct aws_shutdown_callback_options *shutdown_options;
```
graebm, line 43: utterly trivial: const pointers
```suggestion
    const struct aws_shutdown_callback_options *shutdown_options;
```

include/aws/io/event_loop.h
```diff
@@ -387,59 +96,12 @@ AWS_IO_API
...
- * loop count will be the number of available processors in the cpu_group / 2 (to exclude hyper-threads)
- */
-AWS_IO_API
-struct aws_event_loop_group *aws_event_loop_group_new_default_pinned_to_cpu_group(
-    struct aws_allocator *alloc,
-    uint16_t max_threads,
-    uint16_t cpu_group,
-    const struct aws_shutdown_callback_options *shutdown_options);
+    struct aws_allocator *allocator,
+    const struct aws_event_loop_group_options *options);
```
graebm, line 104: This is a breaking change. Probably should keep all the old group_new() functions, but mark them deprecated.

bretambrose: Can't we just document the behavior and update the constructor to only apply the behavior if cpu pinning is enabled?  I do not want to keep the old functions in any way shape or form.

include/aws/io/private/event_loop_impl.h
```diff
@@ -0,0 +1,342 @@
...
+#include <aws/common/atomics.h>
+#include <aws/common/hash_table.h>
+#include <aws/common/ref_count.h>
+
+AWS_PUSH_SANE_WARNING_LEVEL
+
+struct aws_event_loop;
+struct aws_overlapped;
+
+typedef void(aws_event_loop_on_completion_fn)(
```
xiazhvera, line 20: It seems that event loop related members are all moved to private header, are we no longer exposing the event loop API? Would that be a breaking change as well? I did a quick search, we used `aws_event_loop` in multiple c libraries.

bretambrose: What now-private APIs are we using?

xiazhvera: `aws_event_loop_new_default` is used in c-http tests to create an event loop group. I guess we can remove it from c-http and use the new event loop group API anyway.

source/event_loop.c
```diff
@@ -5,6 +5,9 @@
 
 #include <aws/io/event_loop.h>
 
+#include <aws/common/shutdown_types.h>
```
xiazhvera, line 8: Is `shutdown_types.h` a new file? I didn't find it in aws-c-common.

include/aws/io/event_loop.h
graebm: These are functions that MUST be called by any C code written against our libraries. Removing them is going to break any and all code written against aws-c-***. All tests. All samples. All 3rdparty code we don't know exists.

IMO: Marking them deprecated, and just having them call into the new function is going to be far less pain

bretambrose: It would be a lot less work true.  I can do that I suppose.  I'll probably roll back to before I updated all the tests/usage.

graebm (approved):
fix & ship

tests/socket_handler_test.c
```diff
@@ -59,7 +60,8 @@ static int s_socket_common_tester_init(struct aws_allocator *allocator, struct s
     AWS_ZERO_STRUCT(*tester);
     aws_io_library_init(allocator);
 
-    tester->el_group = aws_event_loop_group_new_default(allocator, 0, NULL);
+    struct aws_event_loop_group_options elg_options = {.loop_count = 1};
+    tester->el_group = aws_event_loop_group_new(allocator, &elg_options);
```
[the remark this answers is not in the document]
bretambrose: Probably copy-paste error, will unspecify

PR #692, opened 2024-11-07
Event Loop & Socket Type Multi-Support

xiazhvera:
*Issue #, if available:*

*Description of changes:*
- Allow select from different event loop type
- Allow select from different socket implementation 
- The CI Job was renamed as we introduced new compile options. I will update the required CI once before merge.
macos -> macos(kqueue)
macos-debug -> macos-debug (kqueue)


source/event_loop.c
```diff
@@ -488,6 +540,133 @@ int aws_event_loop_current_clock_time(struct aws_event_loop *event_loop, uint64_
...
+}
+
+/**
+ * Return the default event loop type. If the return value is `AWS_ELT_PLATFORM_DEFAULT`, the function failed to
+ * retrieve the default type value.
+ * If `aws_event_loop_override_default_type` has been called, return the override default type.
+ */
+static enum aws_event_loop_type aws_event_loop_get_default_type(void) {
+#ifdef AWS_USE_APPLE_NETWORK_FRAMEWORK
+    aws_event_loop_override_default_type(AWS_ELT_DISPATCH_QUEUE);
```
graebm, line 563: it's really weird that this getter is calling the setter every time. So if user calls aws_event_loop_override_default_type(), then calls aws_event_loop_get_default_type(), the type flips back ???

maybe less confusing to have this ifdef up where the variable is defined?
```
#ifdef AWS_USE_APPLE_NETWORK_FRAMEWORK
    static enum aws_event_loop_type s_default_event_loop_type_override = AWS_ELT_DISPATCH_QUEUE;
#else
    static enum aws_event_loop_type s_default_event_loop_type_override = AWS_ELT_PLATFORM_DEFAULT;
#endif
```

source/socket.c
```diff
@@ -0,0 +1,267 @@
...
+
+    // 1. get socket type & validate type is avliable the platform
+    enum aws_socket_impl_type type = options->impl_type;
+    if (type == AWS_SIT_PLATFORM_DEFAULT) {
+        type = aws_socket_get_default_impl_type();
+    }
+
+    if (aws_socket_impl_type_validate_platform(type)) {
+        AWS_LOGF_DEBUG(AWS_LS_IO_SOCKET, "Invalid event loop type on the platform.");
+        return AWS_ERROR_PLATFORM_NOT_SUPPORTED;
```
graebm, line 119:
```suggestion
        return aws_raise_error(AWS_ERROR_PLATFORM_NOT_SUPPORTED);
```

graebm (commented):
posting what I got so far...

CMakeLists.txt
```diff
@@ -111,8 +111,6 @@ elseif (APPLE)
             )
 
     file(GLOB AWS_IO_OS_SRC
-            "source/bsd/*.c"
-            "source/posix/*.c"
```
graebm, line 115: won't we need posix/*.c on iOS too? (APPLE but !DARWIN)

include/aws/io/socket.h
```diff
@@ -78,7 +99,7 @@ typedef void(aws_socket_on_connection_result_fn)(struct aws_socket *socket, int
  * A user may want to call aws_socket_set_options() on the new socket if different options are desired.
  *
  * new_socket is not yet assigned to an event-loop. The user should call aws_socket_assign_to_event_loop() before
- * performing IO operations.
+ * performing IO operations. The user is responsible to releasing the socket memory after use.
```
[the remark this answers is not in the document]
xiazhvera: Update: `socket` may be NULL in the callback if the socket is released and cleaned up before a callback is triggered.

include/aws/io/socket.h
```diff
@@ -94,6 +115,8 @@ typedef void(aws_socket_on_accept_result_fn)(
 /**
  * Callback for when the data passed to a call to aws_socket_write() has either completed or failed.
  * On success, error_code will be AWS_ERROR_SUCCESS.
+ *
+ * socket is possible to be a NULL pointer in the callback.
```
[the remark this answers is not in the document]
bretambrose: Maybe sockets need ref-counting to prevent?

include/aws/io/private/event_loop_impl.h
```diff
@@ -172,6 +192,8 @@ struct aws_event_loop *aws_event_loop_new_default(struct aws_allocator *alloc, a
 /**
  * Creates an instance of the default event loop implementation for the current architecture and operating system using
  * extendable options.
+ *
+ * Please note the event loop type defined in the options will be ignored.
  */
 AWS_IO_API
 struct aws_event_loop *aws_event_loop_new_default_with_options(
```
graebm, line 199: Can we just eliminate all these alternate constructors and replace them with a single `aws_event_loop_new(allocator, options)`?

I mean ... they were already in a private header
or is that going to happen in some future PR?

source/exponential_backoff_retry_strategy.c
```diff
@@ -10,8 +10,8 @@
 #include <aws/common/clock.h>
 #include <aws/common/device_random.h>
 #include <aws/common/mutex.h>
-#include <aws/common/shutdown_types.h>
```
graebm, line 13: trivial: change back?

source/socket.c
```diff
@@ -0,0 +1,267 @@
...
+        case AWS_SIT_POSIX:
+            return aws_socket_init_posix(socket, alloc, options);
+            break;
+        case AWS_SIT_WINSOCK:
+            return aws_socket_init_winsock(socket, alloc, options);
+            break;
+
+        case AWS_SIT_APPLE_NETWORK_FRAMEWORK:
+            AWS_ASSERT(false && "Invalid socket implementation on platform.");
+            return aws_socket_init_apple_nw_socket(socket, alloc, options);
```
graebm, line 133: why is this assert(false) here?

source/socket.c
```diff
@@ -0,0 +1,267 @@
...
+    if (aws_socket_impl_type_validate_platform(type)) {
+        AWS_LOGF_DEBUG(AWS_LS_IO_SOCKET, "Invalid event loop type on the platform.");
+        return AWS_ERROR_PLATFORM_NOT_SUPPORTED;
+    }
+
+    // 2. setup vtable based on socket type
+    switch (type) {
+        case AWS_SIT_POSIX:
+            return aws_socket_init_posix(socket, alloc, options);
+            break;
```
graebm, line 126: trivial/debatable: combining `return` and `break` is weird

include/aws/io/socket.h
```diff
@@ -114,7 +137,44 @@ struct aws_socket_endpoint {
     uint32_t port;
 };
 
+struct aws_socket;
+
+struct aws_socket_vtable {
```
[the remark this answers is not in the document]
xiazhvera: Moving the aws_socket into private will affect several downstream libs (mainly c-http). I will hold it for now and create its own PR later.

source/socket.c
xiazhvera: At the point of this PR, apple network framework is not supported yet. Added the assertion to warn developer that we should never use this type so far. 
I will update the comments to explan it.

include/aws/io/private/event_loop_impl.h
xiazhvera: I eventually keep two constructors: `aws_event_loop_new_default` and `aws_event_loop_new`. `aws_event_loop_new_default` was used cross the aws-c-io tests to create default event loop. If we remove it, we probably need create help functions in all those test files, therefore I leave it there...

source/socket.c
```diff
@@ -0,0 +1,263 @@
...
+
+void aws_socket_endpoint_init_local_address_for_test(struct aws_socket_endpoint *endpoint) {
+    (void)endpoint;
+    struct aws_uuid uuid;
+    AWS_FATAL_ASSERT(aws_uuid_init(&uuid) == AWS_OP_SUCCESS);
+    char uuid_str[AWS_UUID_STR_LEN] = {0};
+    struct aws_byte_buf uuid_buf = aws_byte_buf_from_empty_array(uuid_str, sizeof(uuid_str));
+    AWS_FATAL_ASSERT(aws_uuid_to_str(&uuid, &uuid_buf) == AWS_OP_SUCCESS);
+
+#if defined(AWS_ENABLE_KQUEUE) || defined(AWS_ENABLE_EPOLL)
```
bretambrose, line 159: Something feels off about this function using event loop #defines and not default socket type

source/socket.c
```diff
@@ -0,0 +1,263 @@
...
+    type = AWS_SOCKET_IMPL_APPLE_NETWORK_FRAMEWORK;
+#endif // AWS_USE_APPLE_NETWORK_FRAMEWORK
+    if (type != AWS_SOCKET_IMPL_PLATFORM_DEFAULT) {
+        return type;
+    }
+/**
+ * Ideally we should use the platform definition (e.x.: AWS_OS_APPLE) here, however the platform
+ * definition was declared in aws-c-common. We probably do not want to introduce extra dependency here.
+ */
+#if defined(AWS_ENABLE_KQUEUE) || defined(AWS_ENABLE_EPOLL)
```
bretambrose, line 187: elif this sequence, otherwise having multiple types enabled leads to multiple return statements with all but one useless

tests/event_loop_test.c
```diff
@@ -172,7 +172,7 @@ static int s_test_event_loop_canceled_tasks_run_in_el_thread(struct aws_allocato
 
 AWS_TEST_CASE(event_loop_canceled_tasks_run_in_el_thread, s_test_event_loop_canceled_tasks_run_in_el_thread)
 
-#if AWS_USE_IO_COMPLETION_PORTS
+#if AWS_ENABLE_IO_COMPLETION_PORTS
```
bretambrose, line 175: Some tests that verify failure on mismatch between request type (for both sockets and event loops) and platform would be nice.

source/windows/iocp/socket.c
```diff
@@ -239,6 +270,24 @@ static struct socket_vtable vtables[3][2] = {
         },
 };
 
+struct aws_socket_vtable g_winsock_vtable = {
```
bretambrose, line 273: static, -g_ + s_

source/bsd/kqueue_event_loop.c
```diff
@@ -131,7 +131,8 @@ struct aws_event_loop_vtable s_kqueue_vtable = {
     .is_on_callers_thread = s_is_event_thread,
 };
 
-struct aws_event_loop *aws_event_loop_new_default_with_options(
+#ifdef AWS_ENABLE_KQUEUE
```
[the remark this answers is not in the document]
xiazhvera: If not defined AWS_ENABLE_KQUEUE, the following function should be defined. (in source/event_loop.c)
```
#ifndef AWS_ENABLE_KQUEUE
struct aws_event_loop *aws_event_loop_new_with_kqueue(
    struct aws_allocator *alloc,
    const struct aws_event_loop_options *options) {
    (void)alloc;
    (void)options;
    AWS_ASSERT(0);

    AWS_LOGF_DEBUG(AWS_LS_IO_EVENT_LOOP, "Kqueue is not supported on the platform");
    aws_raise_error(AWS_ERROR_PLATFORM_NOT_SUPPORTED);
    return NULL;
}
#endif // AWS_ENABLE_EPOLL
```

source/event_loop.c
```diff
@@ -532,3 +569,131 @@ struct aws_event_loop *aws_event_loop_new_base(
...
+    }
+    return AWS_OP_SUCCESS;
+}
+
+struct aws_event_loop *aws_event_loop_new_dispatch_queue_with_options(
+    struct aws_allocator *alloc,
+    const struct aws_event_loop_options *options) {
+    (void)alloc;
+    (void)options;
+    AWS_ASSERT(0);
```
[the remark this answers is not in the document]
xiazhvera: Ideally, user should never call the function if the event loop is not enabled. I guess it is fine to remove the assert as the function would raise error anyway.

source/event_loop.c
```diff
@@ -532,3 +569,131 @@ struct aws_event_loop *aws_event_loop_new_base(
...
+#ifdef AWS_OS_WINDOWS
+    return AWS_EVENT_LOOP_IOCP;
+#endif
+}
+
+static int aws_event_loop_type_validate_platform(enum aws_event_loop_type type) {
+    switch (type) {
+        case AWS_EVENT_LOOP_EPOLL:
+#ifndef AWS_ENABLE_EPOLL
+            AWS_LOGF_DEBUG(AWS_LS_IO_EVENT_LOOP, "Event loop type EPOLL is not supported on the platform.");
```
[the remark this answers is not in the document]
xiazhvera: I’d like to clarify the distinction between debug and error logs to ensure we're aligned:

For this case, since we're raising and returning an error for unsupported platforms, the caller can handle the returned value and process the result accordingly. I feel this would fall under a debug log.

On the other hand, I think an error log applies to situations that cannot be recovered from. For example, in the `aws_event_loop_get_default_type(void`)` function you mentioned above, if nothing can be returned, that would definitely qualify as an error log.

source/socket.c
```diff
@@ -0,0 +1,259 @@
...
+}
+
+/**
+ * Return the default socket implementation type. If the return value is `AWS_SOCKET_IMPL_PLATFORM_DEFAULT`, the
+ * function failed to retrieve the default type value.
+ */
+static enum aws_socket_impl_type aws_socket_get_default_impl_type(void) {
+    enum aws_socket_impl_type type = AWS_SOCKET_IMPL_PLATFORM_DEFAULT;
+// override default socket
+#ifdef AWS_USE_APPLE_NETWORK_FRAMEWORK
```
bretambrose, line 115: This is pretty convoluted.  Can we just simplify to
```
#ifdef AWS_USE_APPLE_NETWORK_FRAMEWORK
static enum aws_socket_impl_type aws_socket_get_default_impl_type(void) { 
  return AWS_SOCKET_IMPL_APPLE_NETWORK_FRAMEWORK;
}
#else 
<... normal impl>
#endif
```

I don't see how we're not getting dead code warnings when AWS_USE_APPLE_NETWORK is defined.

source/socket.c
```diff
@@ -0,0 +1,259 @@
...
+ * definition was declared in aws-c-common. We probably do not want to introduce extra dependency here.
+ */
+#if defined(AWS_ENABLE_KQUEUE) || defined(AWS_ENABLE_EPOLL)
+    return AWS_SOCKET_IMPL_POSIX;
+#elif AWS_ENABLE_DISPATCH_QUEUE
+    return AWS_SOCKET_IMPL_APPLE_NETWORK_FRAMEWORK;
+#elif AWS_ENABLE_IO_COMPLETION_PORTS
+    return AWS_SOCKET_IMPL_WINSOCK;
+#else
+    return AWS_SOCKET_IMPL_PLATFORM_DEFAULT;
```
bretambrose, line 132: Is this reachable?  Don't we cmake error if no event loop types are enabled?  Maybe add a fatal assert

source/windows/iocp/socket.c
```diff
@@ -357,8 +407,10 @@ static int s_socket_init(
         return AWS_OP_ERR;
     }
 
-    impl->vtable = &vtables[options->domain][options->type];
-    if (!impl->vtable || !impl->vtable->read) {
+    socket->vtable = &s_winsock_vtable;
+
+    impl->winsock_vtable = &s_winsock_vtables[options->domain][options->type];
```
bretambrose, line 412: Can we make the AWS_ASSERT on domain a fatal assert instead?  Same with the assert on type.  Making the asserts fatal means we crash "safely" on invariant violation and not perform undefined behavior by reading from invalid static section memory.

source/bsd/kqueue_event_loop.c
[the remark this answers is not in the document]
xiazhvera: This is a compromise due to a limit on Swift package. 
The Swift package cannot differentiate between Apple platforms when including source files, so the KQUEUE source file gets included for all Apple platforms, as it is required for macOS. While it is possible to define different compile flags for different Apple platforms (e.g., enabling AWS_ENABLE_KQUEUE only for macOS and not for iOS). 
Eventually, on iOS, we end up in a scenario where the kqueue source file is included, but AWS_ENABLE_KQUEUE is disabled.

source/windows/iocp/socket.c
xiazhvera: There is already assertion on domain and type: 
https://github.com/awslabs/aws-c-io/blob/6fcd0198047402687ce892cac07acc6c14493ceb/source/windows/iocp/socket.c#L401-L402

bretambrose: I'm asking for those to become fatal.

bretambrose: because https://github.com/awslabs/aws-c-common/blob/main/include/aws/common/assert.h#L66

xiazhvera:
The CI Job was renamed as we introduced new compile options. I will update the required CI once before merge. 
macos ->  macos(-DAWS_USE_APPLE_NETWORK_FRAMEWORK=OFF)
macos-debug ->  macos-debug (-DAWS_USE_APPLE_NETWORK_FRAMEWORK=OFF)

source/event_loop.c
```diff
@@ -6,20 +6,124 @@
...
+#ifdef AWS_ENABLE_KQUEUE
+    return AWS_EVENT_LOOP_KQUEUE;
+#elif defined(AWS_ENABLE_DISPATCH_QUEUE)
+    return AWS_EVENT_LOOP_DISPATCH_QUEUE;
+#elif defined(AWS_ENABLE_EPOLL)
+    return AWS_EVENT_LOOP_EPOLL;
+#elif defined(AWS_OS_WINDOWS)
+    return AWS_EVENT_LOOP_IOCP;
+#else
+    AWS_LOGF_ERROR(
```
bretambrose, line 94: There's no return value here.  This either needs to crash or return a value.

source/event_loop.c
```diff
@@ -462,10 +566,11 @@ int aws_event_loop_subscribe_to_io_events(
     aws_event_loop_on_event_fn *on_event,
     void *user_data) {
 
-    AWS_ASSERT(event_loop->vtable && event_loop->vtable->subscribe_to_io_events);
-    return event_loop->vtable->subscribe_to_io_events(event_loop, handle, events, on_event, user_data);
+    if (event_loop->vtable && event_loop->vtable->subscribe_to_io_events) {
```
bretambrose, line 569: There's some inconsistency here that feels off.  Most calls we assert the function pointer and invoke it.  But here we have two functions that check it and return an error if it's not set.  Is there something sensible and consistent we could do?

tests/event_loop_test.c
```diff
@@ -971,7 +971,60 @@ static int s_test_event_loop_readable_event_on_2nd_time_readable(struct aws_allo
...
+        aws_event_loop_destroy(event_loop);
+    } else {
+        ASSERT_NULL(event_loop);
+    }
+
+    return AWS_OP_SUCCESS;
+}
+
+/* Verify default event loop type */
+static int s_test_event_loop_all_types_creation(struct aws_allocator *allocator, void *ctx) {
```
bretambrose, line 1001: Something about this bothers me.  I would rather have four tests, one for each type then something that combines them all into one.  Also ORing the results seems wrong both logically (only one has to pass) and semantically (technically these are AWS_OP_* values)

tests/socket_test.c
```diff
@@ -389,6 +389,47 @@ static int s_test_socket_ex(
...
+    int winsock_expected_result = AWS_ERROR_PLATFORM_NOT_SUPPORTED;
+#if defined(AWS_ENABLE_KQUEUE) || defined(AWS_ENABLE_EPOLL)
+    posix_expected_result = AWS_OP_SUCCESS;
+#endif
+#ifdef AWS_ENABLE_IO_COMPLETION_PORTS
+    winsock_expected_result = AWS_OP_SUCCESS;
+#endif
+    // TODO: Apple Network Framework is not implemented yet. Add the related socket test later.
+
+    return s_test_socket_creation(allocator, AWS_SOCKET_IMPL_POSIX, posix_expected_result) ||
```
bretambrose, line 427: Similar comment here: separate tests and don't combine results with logical OR

bretambrose (approved):
Fix n ship.  Check with Mike to see if he wants to give another look.

include/aws/io/socket.h
```diff
@@ -78,7 +98,7 @@ typedef void(aws_socket_on_connection_result_fn)(struct aws_socket *socket, int
  * A user may want to call aws_socket_set_options() on the new socket if different options are desired.
  *
  * new_socket is not yet assigned to an event-loop. The user should call aws_socket_assign_to_event_loop() before
- * performing IO operations.
+ * performing IO operations. The user must call `aws_socket_release()` when they're done with the socket, to free it.
```
graebm, line 101: There is no `aws_socket_release()`? Is this something that existed in some old commit and deleted since?

include/aws/io/socket.h
```diff
@@ -172,10 +185,15 @@ AWS_IO_API void aws_socket_clean_up(struct aws_socket *socket);
  * In TCP, LOCAL and VSOCK this function will not block. If the return value is successful, then you must wait on the
  * `on_connection_result()` callback to be invoked before using the socket.
  *
+ * The function will failed with error if the endpoint is invalid, except for Apple Network Framework. In Apple network
+ * framework, as connect is an async api, we would not know if the local endpoint is valid until we have the connection
+ * state returned in callback. The error will returned in `on_connection_result` callback
+ *
```
graebm, line 191: may just omit these details about endpoint validation?

I mean, the API contract is still the same: if this function returns 0, then the `on_connection_result` will eventually fire and inform you of whether or not it ultimately succeeded

PR #694, opened 2024-11-11
Add an Option to disable retries

waahm7:
*Issue #, if available:*
https://github.com/aws/aws-sdk-cpp/issues/3131

*Description of changes:*

A value of 0 is treated as default, and users have no way to disable retries. Setting the bucket capacity to 1 leads to one retry. This PR adds an explicit option to disable retries if required.


include/aws/io/retry_strategy.h
```diff
@@ -106,14 +106,16 @@ enum aws_exponential_backoff_jitter_mode {
...
+    /* Max retries to allow. The default value is 10 */
     size_t max_retries;
-    /** Scaling factor to add for the backoff. Default is 500ms */
+    /* Scaling factor to add for the backoff. Default is 500ms */
     uint32_t backoff_scale_factor_ms;
-    /** Max retry backoff in seconds. Default is 20 seconds */
+    /* Max retry backoff in seconds. Default is 20 seconds */
     uint32_t max_backoff_secs;
+    /* Disables retries completely. Attempts to acquire_token will fail with `AWS_IO_RETRY_PERMISSION_DENIED` */
+    bool no_retries;
```
TingDaoK, line 118: I am not a fan of the no_retries config hidden inside of the backoff retry options.

I'd suggest to create a new API with `aws_retry_strategy_new_no_retry()` to be more clear.

waahm7: Thanks, I thought it adds to much boiler late for a simple case of no_retries but it makes sense to have it as a separate strategy. I have updated it.

source/no_retry_strategy.c
```diff
@@ -0,0 +1,90 @@
...
+
+struct aws_retry_strategy *aws_retry_strategy_new_no_retry(
+    struct aws_allocator *allocator,
+    const struct aws_no_retry_options *config) {
+    struct aws_retry_strategy_no_retries *strategy =
+        aws_mem_calloc(allocator, 1, sizeof(struct aws_retry_strategy_no_retries));
+
+    if (!strategy) {
+        return NULL;
+    }
```
TingDaoK, line 79: trivial: don't try to handle the allocation failure, just assert on the address. So, that the caller can skip handling the failures from this.

include/aws/io/retry_strategy.h
```diff
@@ -235,6 +243,16 @@ AWS_IO_API struct aws_retry_strategy *aws_retry_strategy_new_standard(
     struct aws_allocator *allocator,
     const struct aws_standard_retry_options *config);
 
+/**
+ * This retry strategy is used to disable retries.
```
TingDaoK, line 247: add document to mention that the config is optional, it can be null. Since normally, the config will be required to be not null.

TingDaoK (approved):
fix&ship

source/no_retry_strategy.c
waahm7: Thanks, updated.

include/aws/io/retry_strategy.h
waahm7: Thanks, updated.

PR #708, opened 2025-02-12
Dispatch Queue Shutdown Polish

bretambrose:
Refactors dispatch queue shutdown so that no loop-associated memory is unreleased at the time event loop group destroy completes.

### DIFF
--- a/source/darwin/dispatch_queue_event_loop.c
+++ b/source/darwin/dispatch_queue_event_loop.c
@@ -19,7 +19,8 @@
 #include <dispatch/dispatch.h>
 #include <dispatch/queue.h>
 
-static void s_destroy(struct aws_event_loop *event_loop);
+static void s_start_destroy(struct aws_event_loop *event_loop);
+static void s_complete_destroy(struct aws_event_loop *event_loop);
 static int s_run(struct aws_event_loop *event_loop);
 static int s_stop(struct aws_event_loop *event_loop);
 static int s_wait_for_stop_completion(struct aws_event_loop *event_loop);
@@ -60,7 +61,8 @@ static void *s_get_base_event_loop_group(struct aws_event_loop *event_loop);
 static bool s_is_on_callers_thread(struct aws_event_loop *event_loop);
 
 static struct aws_event_loop_vtable s_vtable = {
-    .destroy = s_destroy,
+    .start_destroy = s_start_destroy,
+    .complete_destroy = s_complete_destroy,
     .run = s_run,
     .stop = s_stop,
     .wait_for_stop_completion = s_wait_for_stop_completion,
@@ -380,7 +382,11 @@ populate_local_cross_thread_tasks:
     s_dispatch_event_loop_destroy(dispatch_loop->base_loop);
 }
 
-static void s_destroy(struct aws_event_loop *event_loop) {
+static void s_start_destroy(struct aws_event_loop *event_loop) {
+    (void)event_loop;
+}
+
+static void s_complete_destroy(struct aws_event_loop *event_loop) {
     AWS_LOGF_TRACE(AWS_LS_IO_EVENT_LOOP, "id=%p: Destroying Dispatch Queue Event Loop", (void *)event_loop);
     struct aws_dispatch_loop *dispatch_loop = event_loop->impl_data;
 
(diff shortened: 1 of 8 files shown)


source/darwin/dispatch_queue_event_loop.c
```diff
@@ -570,28 +631,30 @@ static void s_run_iteration(void *service_entry) {
  * aws_dispatch_loop->sycned_data
  */
 static void s_try_schedule_new_iteration(struct aws_dispatch_loop *dispatch_loop, uint64_t timestamp) {
-    if (dispatch_loop->synced_data.suspended || dispatch_loop->synced_data.is_executing) {
-        return;
-    }
-
-    if (!s_should_schedule_iteration(&dispatch_loop->synced_data.scheduled_iterations, timestamp)) {
+    if (dispatch_loop->synced_data.execution_state == AWS_DLES_SUSPENDED || dispatch_loop->synced_data.is_executing) {
```
sfod, line 634: The check should be `.execution_state != AWS_DLES_RUNNING`, otherwise `s_run_iteration` and `s_schedule_task_*` can schedule one last iteration of dispatch queue after shutdown initiated.

bretambrose: I think you're right

sfod: I was referring only to the first expression of the if check, I needed to be more clear. Or is `dispatch_loop->synced_data.is_executing` really not needed here anymore?

bretambrose: You're right it should still be there.  If we're in the running state and is_executing is true, we want the iteration end block to do a single reschedule and anything that happens in between to get ignored.

sfod (approved):

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -310,99 +338,110 @@ struct aws_event_loop *aws_event_loop_new_with_dispatch_queue(
...
-     * It is possible that there are scheduled_iterations that are be queued to run s_run_iteration() up to 1 second
-     * AFTER s_dispatch_queue_destroy_task() has executued. Any iteration blocks scheduled to run in the future will
-     * keep Apple's dispatch queue alive until the blocks complete.
+     * This is the release of the initial ref count of 1 that the event loop was created with.
      */
-    dispatch_async_and_wait_f(dispatch_loop->dispatch_queue, dispatch_loop, s_dispatch_queue_destroy_task);
+    s_dispatch_loop_release(dispatch_loop);
+
+    s_lock_synced_data(dispatch_loop);
+    aws_condition_variable_wait_pred(
```
sbSteveK, line 428: Trivial: May be worth leaving a comment here as there's kind of a lot going on in this one function. 

THIS thread is blocked at this point until the dispatch loop hits zero refs.
synced lock is held but is released by `aws_condition_variable_wait_pred` to allow the lock to be held by the dispatch_loop's event_loop.
synced lock is held again before the `aws_condition_variable_wait_pred` continues at which point we unlock after.

Or not. Mainly had to dig around a bit to make sure we weren't gonna deadlock here as underlying `pthread_cond_wait()` doesn't give much info.

bretambrose: Well, we use the condition-variable-and-mutex combo all over the codebase (mostly in tests) so I don't think it's critical to explain a single instance of it.

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -570,28 +631,30 @@ static void s_run_iteration(void *service_entry) {
...
-    aws_priority_queue_push_ref(
-        &dispatch_loop->synced_data.scheduled_iterations, (void *)&entry, &entry->priority_queue_node);
-
     /**
      * Apple dispatch queue uses automatic reference counting (ARC). If an iteration is scheduled to run in the future,
      * the dispatch queue will persist until it is executed. Scheduling a block far into the future will keep the
-     * dispatch queue alive unnecessarily long, even after aws_event_loop and aws_dispatch_loop have been fully
-     * destroyed and cleaned up. To mitigate this, we ensure an iteration is scheduled no longer than 1 second in the
+     * dispatch queue alive unnecessarily long, which blocks event loop group shutdown from completion.
+     * To mitigate this, we ensure an iteration is scheduled no longer than 1 second in the
```
sbSteveK, line 642: Trivial: Maybe change this comment to clamping to "AWS_DISPATCH_QUEUE_MAX_FUTURE_SERVICE_INTERVAL" since it's possible we change the time amount in the future?

sbSteveK (approved):

PR #718, opened 2025-03-27
Fix IP address being labelled "bad" for too long

graebm:
**Issue:**

A server went offline for a while to do some updates. But when it came back online later, it wasn't getting any traffic. The expectation is that traffic to this server would eventually resume.

**Diagnosis:**

The default host resolver has a "bad" list of addresses (connect attempt failed). The idea is: don't use these addresses unless there are NO known good addresses.

Due to quirks of the code, addresses in the "bad' list would not be removed when their TTL expired. Maybe this was an accident? Maybe the author just wanted to avoid some looping?

In any case, this could result in the address being labeled "bad" FOREVER. Even if the server and DNS did the right thing: not reporting the address while the server was offline but reporting it again later when everything was good, the aws_host_resolver cache would still have the address in its "bad" list.

The only way to get off the bad list would be if ALL good addresses failed, and the resolver was forced to pull from the bad list.

But we need to handle a reasonable world, where an address might randomly go bad for a while and get better later.

**Description of changes:**

- Remove addresses in the bad list when their TTL expires
  - TTL is currently [30sec by default](https://github.com/awslabs/aws-c-io/blob/9c7c52cb0d61b6644e6fb1973fcd680d62435ac7/source/host_resolver.c#L23-L24), [5min for aws-c-s3](https://github.com/awslabs/aws-c-s3/blob/169842b7e2f81d71d0719d4a77f9c3e186512f99/source/s3_client.c#L81-L82) (these may change in the future)
- If an address is in the "bad" list, but DNS is still reporting it, don't update its TTL.

So now, if there's a failed connection, the address will remain on the "bad" list until its TTL expires, even if that address appears in future DNS queries. Then, the address will be removed from all caches. If a future DNS query returns that address again, it will go into the "good" list.

### DIFF
--- a/source/host_resolver.c
+++ b/source/host_resolver.c
@@ -710,7 +710,7 @@ static int resolver_record_connection_failure(
  */
 
 struct aws_host_address_cache_entry_lookup {
-    struct aws_host_address_cache_entry *found;
+    struct aws_host_address_cache_entry *entry; /* if lookup succeeds, this is non-NULL */
     bool is_fallback;
 };
 
@@ -729,7 +729,7 @@ static struct aws_host_address_cache_entry_lookup s_find_cached_address_entry_au
         }
     }
 
-    return (struct aws_host_address_cache_entry_lookup){.found = found, .is_fallback = is_fallback};
+    return (struct aws_host_address_cache_entry_lookup){.entry = found, .is_fallback = is_fallback};
 }
 
 /*
@@ -748,7 +748,7 @@ static struct aws_host_address_cache_entry_lookup s_find_cached_address_entry(
             return s_find_cached_address_entry_aux(entry->a_records, entry->failed_connection_a_records, address);
 
         default:
-            return (struct aws_host_address_cache_entry_lookup){.found = NULL};
+            return (struct aws_host_address_cache_entry_lookup){.entry = NULL};
     }
 }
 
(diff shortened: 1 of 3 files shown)


source/host_resolver.c
```diff
@@ -520,16 +520,22 @@ static inline void process_records(
...
-            }
+     * of spotty behavior and upgrade it for reuse. */
+    bool should_promote = (record_count == 0);
+    size_t failed_count = aws_cache_get_element_count(failed_records);
+    for (size_t index = 0; index < failed_count; ++index) {
+        struct aws_host_address_cache_entry *lru_element_entry = aws_lru_cache_use_lru_element(failed_records);
+
+        /* remove expired entries */
+        if (timestamp >= lru_element_entry->address.expiry) {
+            AWS_LOGF_DEBUG(
```
DmitriyMusatkin, line 531: debatable: this might be to spammy for debug level

source/host_resolver.c
```diff
@@ -520,16 +520,22 @@ static inline void process_records(
...
+                AWS_LS_IO_DNS,
+                "static: purging expired bad record %s for %s",
+                lru_element_entry->address.address->bytes,
+                lru_element_entry->address.host->bytes);
+            aws_cache_remove(failed_records, lru_element_entry->address.address);
```
DmitriyMusatkin, line 536: is removing elements while iterating through cache safe here? a bit of naive question cause im not entirely familiar with cache impl

source/host_resolver.c
```diff
@@ -779,19 +794,30 @@ static void s_update_address_cache(
...
 
-        if (address_to_cache_entry) {
-            address_to_cache_entry->address.expiry = new_expiration;
-            AWS_LOGF_TRACE(
-                AWS_LS_IO_DNS,
-                "static: updating expiry for %s for host %s to %llu",
-                address_to_cache_entry->address.address->bytes,
-                host_entry->host_name->bytes,
-                (unsigned long long)new_expiration);
+        if (cache_lookup.found) {
```
DmitriyMusatkin, line 800: nit: found through me for a loop on a first read, since it sounds like it should be boolean. maybe something like element?

DmitriyMusatkin (approved):

source/host_resolver.c
graebm: This is the same level (and similar message) as the AWS_LOGF when a "good" address is removed due to TTL

Agree the host resolver is too spammy, but that's for a separate PR

source/host_resolver.c
graebm: renamed `found -> entry`, which aligns with existing variable names

PR #723, opened 2025-04-08
Revert error code ordering

sbSteveK:
Restore ordering of error codes and move new errors to the end


include/aws/io/io.h
```diff
@@ -271,9 +254,26 @@ enum aws_io_errors {
     AWS_IO_STREAM_GET_LENGTH_FAILED,
     AWS_IO_STREAM_SEEK_UNSUPPORTED,
     AWS_IO_STREAM_GET_LENGTH_UNSUPPORTED,
+    AWS_IO_TLS_ERROR_READ_FAILURE,
 
     AWS_ERROR_PEM_MALFORMED,
 
+    AWS_IO_SOCKET_MISSING_EVENT_LOOP,
```
graebm, line 261: Yup. Confirmed this is how it ought to be. Here's a diff of this branch on the right, and an on the left is an old io.h from commit 46974e96 (right before Dispatch Queue was merged)
image

include/aws/io/io.h
```diff
@@ -271,9 +254,26 @@ enum aws_io_errors {
...
+    AWS_IO_TLS_BAD_CERTIFICATE,
+    AWS_IO_TLS_PEER_CERTIFICATE_EXPIRED,
+    AWS_IO_TLS_BAD_PEER_CERTIFICATE,
+    AWS_IO_TLS_PEER_CERTIFICATE_REVOKED,
+    AWS_IO_TLS_PEER_CERTIFICATE_UNKNOWN,
+    AWS_IO_TLS_INTERNAL_ERROR,
+    AWS_IO_TLS_CLOSED_GRACEFUL,
+    AWS_IO_TLS_CLOSED_ABORT,
+    AWS_IO_TLS_INVALID_CERTIFICATE_CHAIN,
+    AWS_IO_TLS_HOST_NAME_MISSMATCH,
```
graebm, line 275: Not sure if it's too late. But it's misspelled 👗🧨
```suggestion
    AWS_IO_TLS_HOST_NAME_MISMATCH,
```

source/io.c
```diff
@@ -350,7 +302,55 @@ static struct aws_error_info s_errors[] = {
     AWS_DEFINE_ERROR_INFO_IO(
         AWS_IO_STREAM_GET_LENGTH_UNSUPPORTED,
         "Get length is not supported in the underlying I/O source."),
+    AWS_DEFINE_ERROR_INFO_IO(
```
graebm, line 305: You could add some defenses against this happening again by adding a test like:
- in io_lib_test.c
- name something like io_library_error_order
- init library, then assert that a few error codes have their current int value. Like maybe pick the first, last, and a middle error code

that would catch someone naively adding an error code to the middle

graebm (approved):

source/io.c
sbSteveK: io_library_error_order test added

include/aws/io/io.h
sbSteveK: It's a new one and should only be thrown at this point with SecItem. Fixed spelling.

PR #725, opened 2025-04-24
Acquire/Release Event Loop

xiazhvera:
*Issue #, if available:*

*Description of changes:*
- What does this PR do
This PR introduces APIs to acquire and release event loops, by acquiring and releasing the event loop group (ELG) that the event loop belongs to. If the event loop does not belong to any group, the acquire operation will fail. However, this case should never occur in practice, as users are expected to interact with the event loop group API directly.

- Why do we need this
Event loops are not ref-counted. For asynchronous APIs like the Dispatch Queue, it is necessary to retain a reference to the ELG ensures the event loop isn’t destroyed while async work is still pending. Besides, In cases where we need to process tasks on an event loop outside the context of a channel (e.g., request response client), we must ensure the event loop remains alive independently of the channel or channel bootstrap. 
These new APIs help to keep the event loop alive as needed.

Also merged with a previous similar PR: https://github.com/awslabs/aws-c-io/pull/536

Ideally socket should also get updated to acquire event loop (https://github.com/awslabs/aws-c-io/blob/7e75f17400ca2cfdc296865b9c2b10323f49a9c5/include/aws/io/socket.h#L159), however, we frequently set/unset event loop in posix socket make things complicated (https://github.com/awslabs/aws-c-io/blob/get_eventloop_group/source/posix/socket.c#L449). I will leave the socket as it is for now.


sbSteveK (approved):
Optional name change for clarity. Otherwise looks solid.

bretambrose (dismissed):
Let's talk about this

source/darwin/nw_socket.c
```diff
@@ -412,19 +412,28 @@ static bool s_validate_event_loop(struct aws_event_loop *event_loop) {
     return event_loop && event_loop->vtable && event_loop->impl_data;
 }
 
-static void s_set_event_loop(struct aws_socket *aws_socket, struct aws_event_loop *event_loop) {
+static bool s_set_event_loop(struct aws_socket *aws_socket, struct aws_event_loop *event_loop) {
```
TingDaoK, line 415: style: don't use bool for error report.

use int, AWS_OP_SUCCESS and AWS_OP_ERR for consistence. 

Otherwise, when we check for error, some place need to check `if false` and the other places need to check `if not zero`, which is easy to mess up.

waahm7 (commented):
Similar PR: https://github.com/awslabs/aws-c-io/pull/536. I don't know why that was not merged.

PR #727, opened 2025-04-30
AWS_IO_TLS_CIPHER_PREF_PQ_TLSv1_0_2021_05 is not supported now.

TingDaoK:
*Issue #, if available:*

- https://github.com/awslabs/aws-c-io/blob/9392c7b06673db4751940a695a60dad8796d67bf/include/aws/io/tls_channel_handler.h#L37 marks the `AWS_IO_TLS_CIPHER_PREF_PQ_TLSv1_0_2021_05` deprecated
- There was a plan for S2N to keep this policies around with graceful fallback, but S2N decided to fully deprecated the policy, [here](https://github.com/aws/s2n-tls/pull/5194/commits/e8a0710b3fcd02145c9cda7402d437259b0d3473#diff-13800ddfa8b6bc9cd94fd97a635ade9381addd21a700125683901b74a4c7471e), so, it's not supported now.
- Remove the cipher from the supported list.

*Description of changes:*

### DIFF
--- a/source/s2n/s2n_tls_channel_handler.c
+++ b/source/s2n/s2n_tls_channel_handler.c
@@ -1539,9 +1539,6 @@ static struct aws_tls_ctx *s_tls_ctx_new(
             /* The specific PQ policy used here may change over time. */
             security_policy = "AWS-CRT-SDK-TLSv1.2-2023-PQ";
             break;
-        case AWS_IO_TLS_CIPHER_PREF_PQ_TLSv1_0_2021_05:
-            security_policy = "PQ-TLS-1-0-2021-05-26";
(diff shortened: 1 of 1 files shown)


waahm7 (approved):
Any ideas why we are maintaining a separate list, and can we just let S2N manage it?

TingDaoK:
> Any ideas why we are maintaining a separate list, and can we just let S2N manage it?

in my opinion, we have a much shorter list we maintain, it's not a bad idea to just keep it in track.
And s2n don't have a public API to check the version is supported or not. https://github.com/aws/s2n-tls/blob/main/api/s2n.h

PR #729, opened 2025-05-06
Fix dispatch queue to use wall time

xiazhvera:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/source/darwin/dispatch_queue_event_loop.c
+++ b/source/darwin/dispatch_queue_event_loop.c
@@ -648,7 +648,7 @@ static void s_try_schedule_new_iteration(struct aws_dispatch_loop *dispatch_loop
          * requested time. Any blocks scheduled using `dispatch_async_f()` or `dispatch_after_f()` with a closer
          * dispatch time will be placed on the dispatch queue and execute in order.
          */
-        dispatch_time_t when = dispatch_time(DISPATCH_TIME_NOW, delta);
+        dispatch_time_t when = dispatch_walltime(DISPATCH_WALLTIME_NOW, delta);
         dispatch_after_f(when, dispatch_loop->dispatch_queue, entry, s_run_iteration);
         AWS_LOGF_TRACE(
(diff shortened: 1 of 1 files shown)


bretambrose (commented):
Can we get an explanation why?

Using wall clock for scheduling opens a potential can of worms.

source/darwin/dispatch_queue_event_loop.c
```diff
@@ -648,7 +648,10 @@ static void s_try_schedule_new_iteration(struct aws_dispatch_loop *dispatch_loop
          * requested time. Any blocks scheduled using `dispatch_async_f()` or `dispatch_after_f()` with a closer
          * dispatch time will be placed on the dispatch queue and execute in order.
          */
-        dispatch_time_t when = dispatch_time(DISPATCH_TIME_NOW, delta);
+
+        // `dispatch_walltime()` will return an absolute time according to the wall clock. As NULL is passed, this
+        // function uses the result of gettimeofday() as the base time, and add delta to it.
+        dispatch_time_t when = dispatch_walltime(NULL, delta);
```
sfod, line 654: Can `dispatch_walltime` be affected by jumps in system time?

According to [docs](https://developer.apple.com/documentation/dispatch/1420517-dispatch_walltime), `dispatch_walltime` is based on `gettimeofday`, which is affected by system time jumps.

xiazhvera: You are right. I would need revisited the clock options (CLOCK_BOOTTIME, CLOCK_MONOTONIC_RAW and CLOCK_MONOTONIC) we had in our code base. Though, I dont think apple supports all 3, I need figure out how to use the proper time in this case here...

xiazhvera:
dispatch_time will return the timestamp matches "clock_gettime_nsec_np(CLOCK_UPTIME_RAW)", which does not match our high res clock. I didnt see if there is a better option to match the clock... leave it for now...

PR #735, opened 2025-06-11
Fix pem validation

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*
Pem validation logic had a potential to read past the bound to check for header when input terminated prematurely with a dash


source/pem.c
```diff
@@ -29,63 +29,56 @@ int aws_sanitize_pem(struct aws_byte_buf *pem, struct aws_allocator *allocator)
         return AWS_OP_ERR;
     }
     struct aws_byte_cursor pem_cursor = aws_byte_cursor_from_buf(pem);
+    /* Note: this is a bit hacky, but begin is basically used as a state that represents everything non-data */
```
TingDaoK, line 32: trivial: we can remove `END` from `aws_pem_parse_state` now.

TingDaoK (approved):
fix&ship

source/pem.c
DmitriyMusatkin: yeah. removed. 
i also turned this line into just a bool to avoid hackiness

PR #736, opened 2025-06-12
Fix casing on Windows header files

giordano:
*Description of changes:*

This change corrects Windows-specific header names to use canonical casing, which helps with cross-compilation on non-Windows hosts (description just copied from #730, since it's the same type of change).  This was broken by @sfod in #732.


sfod:
Thanks for the fix

PR #752, opened 2025-08-20
Scheduled Iteration Mem-release Order

sbSteveK:
Just to be overly safe in memory release timing, release the memory for a run iteration before releasing the dispatch loop.


xiazhvera (approved):
The PR looks good to me. Let's do a multiple CI runs to make sure the race condition is fixed.

PR #754, opened 2025-09-05
Serialized scheduling

bretambrose:
* Adds a separate API to event loop scheduling that does not do the short-circuit optimization when called from the same event loop.  This guarantees task order-of-submission is order-of-execution and is crucial when using tasks as io write events.

### DIFF
--- a/include/aws/io/event_loop.h
+++ b/include/aws/io/event_loop.h
@@ -35,6 +35,7 @@ struct aws_event_loop_vtable {
     int (*stop)(struct aws_event_loop *event_loop);
     int (*wait_for_stop_completion)(struct aws_event_loop *event_loop);
     void (*schedule_task_now)(struct aws_event_loop *event_loop, struct aws_task *task);
+    void (*schedule_task_now_serialized)(struct aws_event_loop *event_loop, struct aws_task *task);
     void (*schedule_task_future)(struct aws_event_loop *event_loop, struct aws_task *task, uint64_t run_at_nanos);
     void (*cancel_task)(struct aws_event_loop *event_loop, struct aws_task *task);
     int (*connect_to_io_completion_port)(struct aws_event_loop *event_loop, struct aws_io_handle *handle);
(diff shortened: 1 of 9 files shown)


source/darwin/dispatch_queue_event_loop.c
```diff
@@ -690,6 +692,11 @@ static void s_schedule_task_now(struct aws_event_loop *event_loop, struct aws_ta
     s_schedule_task_common(event_loop, task, 0 /* zero denotes "now" task */);
 }
 
+/* dispatch queue event loop impl does not have any short-circuiting, so just use the base scheduling logic */
+static void s_schedule_task_now_serialized(struct aws_event_loop *event_loop, struct aws_task *task) {
```
TingDaoK, line 696: trivial: maybe just put this to the vtable, and the comments as well.
```
/* dispatch queue event loop impl does not have any short-circuiting, so just use the base scheduling logic */
.schedule_task_now_serialized = s_schedule_task_now,
```

include/aws/io/event_loop.h
```diff
@@ -116,6 +117,15 @@ AWS_EXTERN_C_BEGIN
...
 void aws_event_loop_schedule_task_now(struct aws_event_loop *event_loop, struct aws_task *task);
 
+/**
+ * Variant of aws_event_loop_schedule_task_now that forces all tasks to go through the cross thread task queue,
+ * guaranteeing that order-of-submission is order-of-execution.  If you need this guarantee, you must use this
+ * function; the base function contains short-circuiting logic that breaks ordering invariants.  Beyond that, all
+ * properties of aws_event_loop_schedule_task_now apply to this function as well.
+ */
+AWS_IO_API
+void aws_event_loop_schedule_task_now_serialized(struct aws_event_loop *event_loop, struct aws_task *task);
```
TingDaoK, line 127: > guaranteeing that order-of-submission is order-of-execution

But if a task is cancelled, that task will not be following this order. Let's point this out in case we forget it in the future.

include/aws/io/event_loop.h
```diff
@@ -35,6 +35,7 @@ struct aws_event_loop_vtable {
     int (*stop)(struct aws_event_loop *event_loop);
     int (*wait_for_stop_completion)(struct aws_event_loop *event_loop);
     void (*schedule_task_now)(struct aws_event_loop *event_loop, struct aws_task *task);
+    void (*schedule_task_now_serialized)(struct aws_event_loop *event_loop, struct aws_task *task);
```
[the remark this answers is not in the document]
bretambrose: Well, it's a pattern that already exists with channel task scheduling as well.

include/aws/io/event_loop.h
bretambrose: Using cancel not only breaks order, it also can happen on any thread.  There's not much we can do there.

PR #758, opened 2025-10-02
macOS dispatch queue and secitem

sbSteveK:
- Use SecItem and dispatch queue on macOS.
- Remove bsd and Kqueue from macOS builds
- Add tests for importing supported certificates
- Add tests for TLS 1.3 connections
- Split out slow CI jobs into its own "ci-slow" workflow to allow retrying flaky tests in faster jobs without needing to cancel slow builds.


.github/workflows/ci.yml
```diff
@@ -245,7 +245,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        eventloop: ["kqueue", "dispatch_queue"]
```
sbSteveK: Let's remove matrix that has a singular option and just use the option in the build command.

tests/CMakeLists.txt
```diff
@@ -278,41 +280,48 @@ if(NOT BYO_CRYPTO)
     add_net_test_case(tls_client_channel_negotiation_success_mozilla_modern)
 
     # Misc non-badssl tls tests
-    add_net_test_case(test_duplicate_cert_import)
+    if(NOT APPLE)
```
sbSteveK: Should we allow this test here since file-based shouldn't require entitlement?

.github/workflows/ci.yml
sfod: I restored the matrix to match the current CI jobs. Currently, we have `macos-debug (dispatch_queue, ,thread)` specified as required, so if remove the matrix, CI will hang up on them.
I could remove the dispath_queue CI jobs from the required, but considering that the development is far from completion, it feels safer to keep the matrix.

tests/CMakeLists.txt
```diff
@@ -278,41 +280,48 @@ if(NOT BYO_CRYPTO)
...
-    # has succeeded prior to the TLS negotiation. As such, this test will not work for Secitem.
-    add_net_test_case(tls_client_channel_negotiation_error_socket_closed)
-
-    # TLS statistics tracks and handles the timeout of TLS. Using SecItem, the TLS handshake takes place within
-    # the socket establishment and does not need a separate timeout task for TLS.
-    add_net_test_case(tls_channel_statistics_test)
-endif()
+
+    if(NOT APPLE)
+        # These tests require the test binary to be codesigned with an Apple Developer account with entitlements.
```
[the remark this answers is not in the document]
sfod: It should. I'll enable this test and we'll see when file-based keychain is merged.

sfod (approved):

PR #766, opened 2025-11-04
macOS Final Pass

sbSteveK:
Final Pass to get macOS branch ready for merge.
Primarily consists of removing any iOS, Apple Network Framework, or Secitem specific branching and using __APPLE__ as everything on apple will use both network framework and secitem now. Old Security Framework used by macOS before the migration has also been cleaned up.

### DIFF
--- a/source/darwin/darwin_pki_utils.c
+++ b/source/darwin/darwin_pki_utils.c
@@ -45,7 +45,7 @@ static int s_import_key_into_keychain_with_seckeychain(
     struct aws_allocator *alloc,
     CFAllocatorRef cf_alloc,
     const struct aws_byte_cursor *private_key,
-    SecKeychainRef import_keychain) {
+    AwsSecKeychainRef import_keychain) {
 
     (void)alloc;
     (void)cf_alloc;
@@ -130,7 +130,7 @@ static int s_aws_secitem_add_certificate_to_keychain(
     SecCertificateRef cert_ref,
     CFDataRef serial_data,
     CFStringRef label,
-    SecKeychainRef import_keychain) {
+    AwsSecKeychainRef import_keychain) {
 
     (void)import_keychain;
 
@@ -245,7 +245,7 @@ static int s_aws_secitem_add_private_key_to_keychain(
     SecKeyRef key_ref,
     CFStringRef label,
     CFDataRef application_label,
-    SecKeychainRef import_keychain) {
+    AwsSecKeychainRef import_keychain) {
 
     (void)import_keychain;
 
@@ -359,7 +359,7 @@ static int s_aws_secitem_get_identity(
     CFDataRef serial_data,
     SecCertificateRef cert_ref,
     sec_identity_t *out_identity,
-    SecKeychainRef import_keychain) {
+    AwsSecKeychainRef import_keychain) {
 
     (void)cert_ref;
     (void)import_keychain;
@@ -520,7 +520,7 @@ int s_import_private_key_into_keychain(
     CFAllocatorRef cf_alloc,
     const struct aws_byte_cursor *private_key,
     const struct aws_secitem_options *secitem_options,
-    SecKeychainRef import_keychain) {
+    AwsSecKeychainRef import_keychain) {
 
     int result = AWS_OP_ERR;
 


tests/tls_handler_test.c
```diff
@@ -1444,6 +1444,7 @@ static int s_verify_good_host(
     return AWS_OP_SUCCESS;
 }
 
+// TODO Rename, no mqtt.
```
sbSteveK: Let's do this while we're here.

tests/tls_handler_test.c
```diff
@@ -2641,18 +2642,14 @@ static int s_tls_destroy_null_context(struct aws_allocator *allocator, void *ctx
 }
 AWS_TEST_CASE(tls_destroy_null_context, s_tls_destroy_null_context);
 
-static int s_test_ecc_cert_import(struct aws_allocator *allocator, void *ctx) {
-    (void)ctx;
-    (void)allocator;
-
-#    ifndef AWS_OS_APPLE
+static int s_test_ecc_import(struct aws_allocator *allocator, const char *key_path, const char *cert_path) {
```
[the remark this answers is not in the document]
sbSteveK: A bit out of scope for this but maybe before we do a final merge we can actually use this func across a handful of other tests that repeatedly run through the same steps.

tests/tls_handler_test.c
```diff
@@ -2672,12 +2669,23 @@ static int s_test_ecc_cert_import(struct aws_allocator *allocator, void *ctx) {
...
     aws_io_library_clean_up();
-#    endif /* AWS_OS_APPLE */
 
     return AWS_OP_SUCCESS;
 }
 
-AWS_TEST_CASE(test_ecc_cert_import, s_test_ecc_cert_import)
+static int s_test_ecc_p256_cert_import(struct aws_allocator *allocator, void *ctx) {
+    (void)ctx;
+    return s_test_ecc_import(allocator, "ecc-key.pem", "ecc-cert.pem");
```
sbSteveK: Trivial: Let's rename the cert/key pair to include 256 the way we differentiate the 384 cert/key for clarity in resources.

source/darwin/darwin_pki_utils.c
```diff
@@ -517,39 +579,214 @@ int aws_secitem_import_cert_and_key(
...
+     */
     key_ref = SecKeyCreateWithData(key_data, key_attributes, &error);
     if (error) {
         char description_buffer[256];
         aws_get_core_foundation_error_description(error, description_buffer, sizeof(description_buffer));
-        AWS_LOGF_ERROR(AWS_LS_IO_PKI, "Failed creating SecKeyRef from key_data: %s", description_buffer);
+        AWS_LOGF_WARN(AWS_LS_IO_PKI, "Failed importing private key using SecItem: %s", description_buffer);
+
+        /*
+         * If parsing with SecItem fails, we fall back to trying to add the private key via SecKeychain API.
```
sbSteveK: Either we should rename `s_import_ecc_key_into_keychain` to not be "ecc" specific or we should call out that we are attempting to import an ecc private key.

source/darwin/darwin_pki_utils.c
```diff
@@ -26,11 +26,86 @@ void aws_cf_release(CFTypeRef obj) {
...
+    // A PEM file could contains multiple PEM data section. Try importing each PEM section until find the first
+    // succeed key.
+    for (size_t index = 0; index < aws_array_list_length(&decoded_key_buffer_list); index++) {
+        struct aws_pem_object *pem_object_ptr = NULL;
+        /* We only check the first pem section. Currently, we dont support key with multiple pem section. */
+        aws_array_list_get_at_ptr(&decoded_key_buffer_list, (void **)&pem_object_ptr, index);
+        AWS_ASSERT(pem_object_ptr);
+        CFDataRef key_data = CFDataCreate(cf_alloc, pem_object_ptr->data.buffer, pem_object_ptr->data.len);
+        if (!key_data) {
+            AWS_LOGF_WARN(AWS_LS_IO_PKI, "static: error in creating ECC key data system call.");
```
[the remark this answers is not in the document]
sfod: Actually, this one should be error as it's not supposed to fail even for non-private-key PEM sections. Additional info sounds good, will add.

PR #769, opened 2025-11-10
CI Race Condition Fix

sbSteveK:
A test on event loop serialized scheduling was encountering a race condition in which a conditional variable was being used after it was cleaned up. Reordering the release of the event loop group before cleaning up the context should resolve this issue.


tests/event_loop_test.c
```diff
@@ -1568,9 +1568,11 @@ static int s_test_event_loop_serialized_scheduling(struct aws_allocator *allocat
     aws_thread_join(&external_thread);
     aws_thread_clean_up(&external_thread);
 
-    s_serialized_scheduling_context_clean_up(&context);
+    /* Ensure event loop group is released before cleaning up context to avoid race */
     aws_event_loop_group_release(event_loop_group);
 
+    s_serialized_scheduling_context_clean_up(&context);
```
sfod: Why does reordering function help? AFAICS `aws_event_loop_group_release` just fires up a new thread that will release the event loop group. There is no synchronization there.

PR #773, opened 2025-11-26
Final Pass Changes

sbSteveK:
Debatable but I think we might want to remove the reliance on CMakeLists defines for separation between macOS and iOS/tvOS. One reason being that it adds another point of potential failure for anyone using the CRT that's may be using different build tools which feels slightly more likely for building for iOS/tvOS :man-shrugging::skin-tone-2: TargetConditionals.h (available on all apple OS) exposes TARGET_OS_IPHONE which allows us to drop the use of both aws_is_using_secitem() (since EVERYTHING is now using SecItem) and s_is_filebased_keychain. 


sfod:
LGTM, though I'm not sure about replacing a runtime function with ifdefs. Code without ifdefs is easier to read.

PR #776, opened 2025-12-03
use issuer as well for certificate import and identity retrieval

sbSteveK:
Add the Issuer key for certificate ingestion as well as identity retrieval.

### DIFF
--- a/source/darwin/darwin_pki_utils.c
+++ b/source/darwin/darwin_pki_utils.c
@@ -366,7 +366,7 @@ static int s_aws_secitem_get_identity(
     int result = AWS_OP_ERR;
     OSStatus status;
     CFMutableDictionaryRef search_query = NULL;
-    CFArrayRef sec_identity_array = NULL;
+    SecIdentityRef sec_identity_ref = NULL;
 
     CFArrayRef cert_filter = NULL;
     CFArrayRef keychain_filter = NULL;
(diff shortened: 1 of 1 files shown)


source/darwin/darwin_pki_utils.c
```diff
@@ -410,57 +415,20 @@ static int s_aws_secitem_get_identity(
...
-         * during establishing a TLS connection, user verification will be done when Network framework will try to
-         * retrieve private key data from the provided identity. That part remains the same. */
-        const SecIdentityRef sec_identity_ref = (const SecIdentityRef)CFArrayGetValueAtIndex(sec_identity_array, i);
-
-        SecCertificateRef found_cert = NULL;
-        OSStatus copy_cert_status = SecIdentityCopyCertificate(sec_identity_ref, &found_cert);
-        if (copy_cert_status != errSecSuccess) {
-            AWS_LOGF_ERROR(AWS_LS_IO_PKI, "SecIdentityCopyCertificate failed with OSStatus %d", (int)copy_cert_status);
+
+    if (identity_num == 1) {
```
sfod: If a user keeps their cert/key in multiple keychains (just because they can), `identity_num` will be >1. The use case sounds kinda ridiculous, but maybe we still should handle this?

sfod: Setting keychain to the default one (around line 395) will address this use case.
Or, we can return only one result, as we did before.

PR #788, opened 2026-01-08
Return error on using tls13 on macOS

sfod:
Return error on trying to use tls13 on macOS

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -297,8 +297,8 @@ if(NOT BYO_CRYPTO)
     add_net_test_case(alpn_successfully_negotiates)
     add_net_test_case(alpn_no_protocol_message)
     add_net_test_case(test_ecc_cert_import)
+    add_net_test_case(test_tls_cipher_preference)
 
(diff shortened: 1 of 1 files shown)


source/darwin/secure_transport_tls_channel_handler.c
```diff
@@ -926,17 +929,18 @@ static struct aws_channel_handler *s_tls_handler_new(
 #if TLS13_AVAILABLE
             SSLSetProtocolVersionMin(secure_transport_handler->ctx, kTLSProtocol13);
 #else
-            AWS_LOGF_FATAL(
+            AWS_LOGF_ERROR(
                 AWS_LS_IO_TLS,
                 "static: TLS 1.3 is not supported on this device. You may just want to specify "
```
sbSteveK: Trivial: I'm assuming this will be relatively clear to anyone that sees it but could we add clarity for users in this log? Something along the lines of "Minimum TLS version has been set to TLS 1.3 which is not supported on this device. If minimum TLS version is not set or AWS_IO_TLS_VER_SYS_DEFAULTS is used the latest supported version of TLS will be used automatically."

PR #798, opened 2026-03-26
Fix undefined behavior from incompatible function pointer casts in `ref_count` destroy callbacks

bjosv:
This PR fixes UBSan failures caused by calling functions through incompatible pointer types.
`aws_ref_count_init` expects a callback with signature `void (*)(void *)` but we were casting typed destroy functions (e.g., `void (*)(struct aws_async_input_stream *))` to `aws_simple_completion_callback *`.
This is undefined behavior per the C standard, and UBSan rightfully flags it.

**Fixes**
- async_stream.c: Add a wrapper function that casts and forwards to the vtable destroy, instead of casting the vtable destroy directly.
- pkcs11_tls_op_handler.c: Change s_aws_custom_key_op_handler_destroy to accept void * and cast internally, removing the incompatible function pointer cast at the call site.

**How to reproduce**
Build as described in https://github.com/awslabs/aws-c-s3?tab=readme-ov-file#building but enable UBSan via
`-DCMAKE_C_FLAGS="-fsanitize=undefined -fno-sanitize-recover=all"`, then run S3 tests:
`cd build && ctest --output-on-failure`

UBSan will report errors like:
`runtime error: call to function s_aws_custom_key_op_handler_destroy through pointer to incorrect function type 'void (*)(void *)'`


sfod:
Thank you for your contribution. Similarly to aws-c-sdkutils, I'll add a few additional fixes and bring in your changes in https://github.com/awslabs/aws-c-io/pull/844

PR #799, opened 2026-04-24
Support s2n-tls on macOS

sfod:
Add support for s2n-tls as a TLS backend on macOS, enabling TLS 1.3 support.
Apple Secure Transport remains the default for FIPS compliance.


source/s2n/s2n_apple_keychain.c
```diff
@@ -0,0 +1,171 @@
...
+
+done:
+    CFRelease(values);
+    return is_ca;
+}
+
+/* This function loads trusted root CA certificates from the macOS searchable keychains into the s2n config.
+ * NOTE: Certificates from SystemRootCertificates.keychain are NOT processed here since s2n-tls will get them from the
+ * bundle system roots instead. */
+void aws_tls_s2n_load_macos_keychain_root_cas(struct s2n_config *config, struct aws_allocator *alloc) {
```
xiazhvera, line 90: It would be nice to have more comments explain the details in the function

source/s2n/s2n_apple_keychain.c
```diff
@@ -0,0 +1,171 @@
...
+                    if (buf) {
+                        CFStringGetCString(summary, buf, len, kCFStringEncodingUTF8);
+                        AWS_LOGF_DEBUG(AWS_LS_IO_TLS, "Failed to add certificate '%s' to trust store", buf);
+                        aws_mem_release(alloc, buf);
+                    }
+                    CFRelease(summary);
+                } else {
+                    AWS_LOGF_DEBUG(AWS_LS_IO_TLS, "Failed to add certificate to trust store");
+                }
+            }
```
xiazhvera, line 158: Maybe log the success as well?

xiazhvera (approved):
LGTM. Asked for more documentation on the loading keychain CA workflow. 
Not directly related to this PR,while do we have a test in downstream to test against tls1.3 with httprpoxy on macOS?

tests/resources/generateCerts.sh
TingDaoK, line 1: Add a brief readme about how to use it.

source/tls_channel_handler.c
```diff
@@ -783,6 +788,106 @@ void aws_tls_init_static_state(struct aws_allocator *alloc) {
...
+#    elif defined(_WIN32)
+    secure_channel_init_tls_vtable(&s_tls_channel_handler_vtable);
+
+#    elif defined(USE_S2N)
+    s2n_init_tls_vtable(&s_tls_channel_handler_vtable);
+
+#    else
+    AWS_FATAL_ASSERT(false && "No TLS implementation available.");
+#    endif
+}
```
TingDaoK, line 835: document this behavior in README. Especially the `AWS_CRT_USE_NON_FIPS_TLS_13`.

source/tls_channel_handler.c
```diff
@@ -783,6 +788,106 @@ void aws_tls_init_static_state(struct aws_allocator *alloc) {
...
+    secure_transport_init_tls_vtable(&s_tls_channel_handler_vtable);
+
+#    elif defined(__APPLE__) && defined(USE_S2N)
+    struct aws_string *use_non_fips_13 = aws_get_env_nonempty(allocator, "AWS_CRT_USE_NON_FIPS_TLS_13");
+    if (use_non_fips_13) {
+        s2n_init_tls_vtable(&s_tls_channel_handler_vtable);
+        aws_string_destroy(use_non_fips_13);
+    } else {
+        secure_transport_init_tls_vtable(&s_tls_channel_handler_vtable);
+    }
```
TingDaoK, line 821: This based on a runtime environment variable. Should we have some protection to avoid going through this logic after first initialization?

source/tls_channel_handler.c
TingDaoK: I'd like to have a readme section about the build time config and the runtime environment variables to be easier to to find.

source/s2n/s2n_tls_channel_handler.c
```diff
@@ -107,7 +111,59 @@ AWS_STATIC_STRING_FROM_LITERAL(s_android_path, "/system/etc/security/cacerts");
 AWS_STATIC_STRING_FROM_LITERAL(s_free_bsd_path, "/usr/local/share/certs");
 AWS_STATIC_STRING_FROM_LITERAL(s_net_bsd_path, "/etc/openssl/certs");
 
-AWS_IO_API const char *aws_determine_default_pki_dir(void) {
+void aws_tls_s2n_load_macos_keychain_root_cas(struct s2n_config *config, struct aws_allocator *alloc);
```
TingDaoK: maybe add a header file for this. Using a forward declaring and hope the linker can find the implementation from another C file seems a bit hacky to me. Or just put it into one file

include/aws/io/pem.h
```diff
@@ -95,5 +95,14 @@ AWS_IO_API int aws_pem_objects_init_from_file_path(
     struct aws_allocator *allocator,
     const char *filename);
 
+/**
+ * Encodes DER-encoded certificate bytes into a PEM-formatted, null-terminated string
+ * stored in out_pem_buf. Caller is responsible for calling aws_byte_buf_clean_up() on out_pem_buf.
+ */
+AWS_IO_API int aws_der_cert_to_pem(
```
TingDaoK: it is probably simpler and easier to follow if we return a aws_string* instead of taking a pointer in and return int. We usually use aws_string* for null-terminated string and buffer/cursor will rely on the length.

source/tls_channel_handler.c
sfod: I've added info at the bottom of the main README.

source/s2n/s2n_tls_channel_handler.c
sfod: Forward declarations used everywhere in this codebase, so it should work. But private header is probably a cleaner choice, so added it.

source/tls_channel_handler.c
sfod: This function is protected by `s_io_library_initialized`, which is set to false only in `aws_io_library_clean_up`, i.e. when CRT is cleaned up by a downstream project. I'm not sure multiple init/cleanup of crt libraries is allowed anywhere (e.g. java is certainly does not support this), maybe in crt-cpp.
Nevertheless, I've added a designated init flag for tls vtable, so it's easier to reason about.

PR #805, opened 2026-06-10
Updated `s_do_read` to handle NULL right slot

xiazhvera:
*Issue #, if available:*

*Description of changes:*
Updated the s_do_read() to handle the case where right slot is NULL. There are two cases where the right slot could be NULL. 

1. Race condition: a readable event fires after the setup callback calls `aws_channel_shutdown` but before the shutdown task runs. In this window, no downstream handler was added yet. We will directly return and wait for the shutdown task process.

2. When using Apple SecItem for TLS: the channel has only a socket slot with no downstream application handler (adj_right is NULL). We should still read from the socket so that the OS network stack can process incoming data (e.g. connection close, errors). The read message is released immediately since there are no downstream slots to pass it to.

The issue was first reported in https://github.com/aws/aws-iot-device-sdk-swift/issues/61. 


source/socket_channel_handler.c
```diff
@@ -126,6 +126,18 @@ static void s_do_read(struct socket_handler *socket_handler) {
         return;
     }
 
+    /* Guard against the race condition where a readable event fires after the setup callback
+     * calls aws_channel_shutdown but before the shutdown task runs.
+     * In this window, adj_right is NULL (no downstream handler was added yet),
+     * and calling aws_channel_slot_downstream_read_window() would assert-crash. */
+    if (socket_handler->slot->adj_right == NULL) {
```
bretambrose: I guess this is ok but I kinda don't like it.  It solves the problem but is not really checking the right thing.  The right thing is "has shutdown been called" not "no downstream handler exists"

There is currently no situation where a channel that is just a socket handler makes sense (bytes always need to be interpreted by an additional processing layer), but it's uncomfortable.

bretambrose (approved):

source/socket_channel_handler.c
```diff
@@ -163,6 +187,16 @@ static void s_do_read(struct socket_handler *socket_handler) {
...
 
+        /* When using Apple SECITEM with only TLS, the channel may have a socket handler slot
+         * without a downstream application handler installed. For the case, we still read
+         * from the socket, but never pass the message to downstream slots. */
+#if defined(AWS_USE_SECITEM)
+        if (socket_handler->slot->adj_right == NULL) {
+            aws_mem_release(message->allocator, message);
+            continue;
+        }
+#endif
```
sfod, line 198: What are the reasons to read data and immediately dispose it for NW socket but skip the reading for POSIX socket?

xiazhvera: For Apple Network Framework (SECITEM), TLS negotiation is handled internally by the `nw_connection` at the socket level. Calling `aws_socket_read` drives the underlying `nw_connection` to process incoming data, including TLS handshake messages. We must perform the read to keep the TLS negotiation progressing. However, in the case with no downstream channel handler (`adj_right`), we don't need to forward the data (if there is any), so we immediately release it.

sfod: Since TLS negotiation is handled internally in network framework, wouldn't it be a part of establishing a socket connection? From my understanding, this function (`s_do_read`) will be called after the TLS connection is established.

xiazhvera: You are right. I revisited the case, actually we don't need the read to handle TLS negotiation, but need it to detect socket errors and socket closures in the case. With Apple Network Framework, socket read is the mechanism that surfaces these events, so the socket handler must keep reading even when there's no downstream handler.

For POSIX sockets, an empty `adj_right` is not a valid channel configuration. The channel always has a TLS handler and/or application handler downstream.

sfod (approved):

PR #806, opened 2026-06-11
Option to disable revocation check

sbSteveK:
Revocation checks done on Windows uses an outbound network call that operates on an OS timeout that doesn't respect the tls timeout that we set. This is due to the TLS handling thread being blocked while Windows attempts to complete the network call to check revocation via CRL/OCSP endpoints. When a network is unavailable, this causes a huge delay during the TLS handshake.

Adding `disable_certificate_revocation_check` option to `aws_tls_ctx_options` which allows revocation checks to be turned off. This bypasses the issue outlined above for Windows.

In Linux, we only check the "Server-stapled" OCSP and this is also bypassed if the `disable_certificate_revocation_check` is set to true.

In Apple, this is a no-op as revocation checks were never done.

This commit also adds two tests against badssl.com. One that confirms that the revocation check fails, and one that confirms that if the revocation check is disabled, we connect even with a bad revocation check.

### DIFF
--- a/include/aws/io/tls_channel_handler.h
+++ b/include/aws/io/tls_channel_handler.h
@@ -305,9 +305,10 @@ struct aws_tls_ctx_options {
      *
      * On Apple platforms, this is a no-op as revocation checking is not enabled by default.
      *
-     * Default is false (revocation checking enabled).
+     * Default is false (Windows makes a best-effort certificate revocation check to CRL/OCSP and Linux (s2n)
+     * checks against stapled responses).
      */
-    bool disable_certificate_revocation_check;
+    bool no_certificate_revocation;
 };
 
 struct aws_tls_negotiated_protocol_message {
@@ -695,15 +696,17 @@ AWS_IO_API void aws_tls_ctx_options_set_verify_peer(struct aws_tls_ctx_options *
 /**
  * Enables or disables certificate revocation checking during TLS negotiation.
  *
- * On Windows (SChannel), when enabled (the default), the TLS handshake will make outbound network calls
- * to CRL/OCSP revocation endpoints. In environments without internet access, this can cause the handshake
- * to block for minutes while waiting for network timeouts.
+ * On Windows (SChannel), the TLS handshake will make outbound network calls to CRL/OCSP revocation endpoints and makes
+ * a best-effort check. In environments without internet access, this can cause the handshake to block for minutes while
+ * waiting for network timeouts.
+ *
+ * On Linux (s2n) checks against stapled responses
  *
  * Set this to true to skip revocation checking entirely.
  */
-AWS_IO_API void aws_tls_ctx_options_set_certificate_revocation_check_disabled(
+AWS_IO_API void aws_tls_ctx_options_set_no_certificate_revocation(
     struct aws_tls_ctx_options *options,
-    bool disabled);
+    bool no_revocation);
 
 /**
  * Sets preferred TLS Cipher List
--- a/include/aws/io/tls_channel_handler.h
+++ b/include/aws/io/tls_channel_handler.h
@@ -293,6 +293,21 @@ struct aws_tls_ctx_options {
      * cipher preferences that allow TLS 1.3. If this is set, we will always use non TLS 1.3 preferences.
      */
     struct aws_custom_key_op_handler *custom_key_op_handler;
+
+    /**
+     * Set to true to disable certificate revocation checking during TLS negotiation.
+     *
+     * On Windows (SChannel), this prevents the TLS handshake from making outbound network calls
+     * to CRL/OCSP revocation endpoints, which can block for minutes when the endpoints are unreachable
+     * (e.g., in private subnets without internet access).
+     *
+     * On Linux (s2n), this disables validation of OCSP stapled responses provided by the server.
+     *
+     * On Apple platforms, this is a no-op as revocation checking is not enabled by default.
+     *
+     * Default is false (revocation checking enabled).
+     */
+    bool disable_certificate_revocation_check;
 };
 
 struct aws_tls_negotiated_protocol_message {
@@ -676,6 +691,19 @@ AWS_IO_API int aws_tls_ctx_options_set_alpn_list(struct aws_tls_ctx_options *opt
  * set verify_peer to true.
  */
 AWS_IO_API void aws_tls_ctx_options_set_verify_peer(struct aws_tls_ctx_options *options, bool verify_peer);
+
+/**
+ * Enables or disables certificate revocation checking during TLS negotiation.
+ *
+ * On Windows (SChannel), when enabled (the default), the TLS handshake will make outbound network calls
+ * to CRL/OCSP revocation endpoints. In environments without internet access, this can cause the handshake
+ * to block for minutes while waiting for network timeouts.
+ *
+ * Set this to true to skip revocation checking entirely.
+ */
+AWS_IO_API void aws_tls_ctx_options_set_certificate_revocation_check_disabled(
+    struct aws_tls_ctx_options *options,
+    bool disabled);
 
 /**
  * Sets preferred TLS Cipher List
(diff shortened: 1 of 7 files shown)


include/aws/io/tls_channel_handler.h
```diff
@@ -293,6 +293,21 @@ struct aws_tls_ctx_options {
...
+     * to CRL/OCSP revocation endpoints, which can block for minutes when the endpoints are unreachable
+     * (e.g., in private subnets without internet access).
+     *
+     * On Linux (s2n), this disables validation of OCSP stapled responses provided by the server.
+     *
+     * On Apple platforms, this is a no-op as revocation checking is not enabled by default.
+     *
+     * Default is false (revocation checking enabled).
+     */
+    bool disable_certificate_revocation_check;
```
azkrishpy: do we need to expose it to all platforms if the issue is specific to windows?

sbSteveK: The option is set in tls options and there's no strong reason to have it only available for Windows. Implementing its use in Linux was trivial and it's a no-op in Apple. Feels cleaner to just have it available everywhere

azkrishpy (approved):

include/aws/io/tls_channel_handler.h
```diff
@@ -293,6 +293,21 @@ struct aws_tls_ctx_options {
...
     struct aws_custom_key_op_handler *custom_key_op_handler;
+
+    /**
+     * Set to true to disable certificate revocation checking during TLS negotiation.
+     *
+     * On Windows (SChannel), this prevents the TLS handshake from making outbound network calls
+     * to CRL/OCSP revocation endpoints, which can block for minutes when the endpoints are unreachable
+     * (e.g., in private subnets without internet access).
+     *
+     * On Linux (s2n), this disables validation of OCSP stapled responses provided by the server.
```
DmitriyMusatkin, line 304: hmm if im reading it correctly it disables asking server for ocsp. arent we supposed to manually validate stapled ocsp response, but i dont see calls to s2n_connection_get_ocsp_response

include/aws/io/tls_channel_handler.h
```diff
@@ -293,6 +293,21 @@ struct aws_tls_ctx_options {
...
+     *
+     * On Apple platforms, this is a no-op as revocation checking is not enabled by default.
```
DmitriyMusatkin, line 306: isnt revocation check enabled by default in security framework?

source/s2n/s2n_tls_channel_handler.c
```diff
@@ -1698,17 +1698,20 @@ static struct aws_tls_ctx *s_tls_ctx_new(
...
-            if (s2n_config_set_status_request_type(s2n_ctx->s2n_config, S2N_STATUS_REQUEST_OCSP) != S2N_SUCCESS) {
-                s_log_and_raise_s2n_errno("ctx: ocsp status request cannot be set");
-                goto cleanup_s2n_config;
-            }
-        } else {
-            if (s2n_error_get_type(s2n_errno) == S2N_ERR_T_USAGE) {
-                AWS_LOGF_INFO(AWS_LS_IO_TLS, "ctx: cannot enable ocsp stapling: %s", s2n_strerror(s2n_errno, "EN"));
+
+        if (!options->disable_certificate_revocation_check) {
+            if (s2n_config_set_check_stapled_ocsp_response(s2n_ctx->s2n_config, 1) == S2N_SUCCESS) {
```
azkrishpy, line 1703: @DmitriyMusatkin this is where we manual verify the response right? the existing path?

source/windows/secure_channel_tls_handler.c
```diff
@@ -2424,7 +2424,9 @@ struct aws_tls_ctx *s_ctx_new(
         dw_flags |= SCH_CRED_IGNORE_NO_REVOCATION_CHECK | SCH_CRED_IGNORE_REVOCATION_OFFLINE |
                     SCH_CRED_NO_SERVERNAME_CHECK | SCH_CRED_MANUAL_CRED_VALIDATION;
     } else if (is_client_mode) {
-        dw_flags |= SCH_CRED_REVOCATION_CHECK_CHAIN | SCH_CRED_IGNORE_REVOCATION_OFFLINE;
+        if (!options->disable_certificate_revocation_check) {
+            dw_flags |= SCH_CRED_REVOCATION_CHECK_CHAIN | SCH_CRED_IGNORE_REVOCATION_OFFLINE;
```
DmitriyMusatkin, line 2428: im a bit confused on this change. isnt SCH_CRED_IGNORE_REVOCATION_OFFLINE supposed to mean that it should work correctly when server is unreacheable

source/s2n/s2n_tls_channel_handler.c
DmitriyMusatkin: yeah ignore me. i think my llm was halucinating and insisting s2n need to have different validation

include/aws/io/tls_channel_handler.h
sbSteveK: It appears as though Apple's revocation checks are opt-in. We never set and use `SecPolicyCreateRevocation` (https://developer.apple.com/documentation/security/secpolicycreaterevocation(_:)?language=objc) and definitely don't use `kSecRevocationOCSPMethod` (https://developer.apple.com/documentation/security/ksecrevocationocspmethod?language=objc) to check against anything online.

source/windows/secure_channel_tls_handler.c
sbSteveK: These are the two flags that are currently set. So our documentation (at least in the tests CMakeLists.txt) claim we don't enable any revocation checks: https://github.com/awslabs/aws-c-io/blob/970516ff2339b3b4f525621a75242577fb63114d/tests/CMakeLists.txt#L221

but in windows we were checking but in the event of being offline (what the ticket was doing) we allow a fail to pass. The problem seems to be that we are bypassing our TLS timeout and other timeouts in favor of the OS level timeout when it comes to attempting to connect to the revocation server. I think the connections would eventually be allowed despite being offline but most likely we were destroying the channel due to all other timeouts being hit by the time the OS unblocked and let eventloop proceed.

tests/tls_handler_test.c
```diff
@@ -1781,6 +1781,30 @@ AWS_TEST_CASE(
     tls_client_channel_negotiation_no_verify_untrusted_root,
     s_tls_client_channel_negotiation_no_verify_untrusted_root_fn)
 
+#    ifdef _WIN32
```
TingDaoK: nit: can we have the test cross platform? 

at least after we disable the revocation, it can success on all platforms, right?

TingDaoK (approved):

source/windows/secure_channel_tls_handler.c
DmitriyMusatkin: Ok, so i did a bit more research on those flags. 
SCH_CRED_REVOCATION_CHECK_CHAIN - enables all revocation checking
SCH_CRED_IGNORE_REVOCATION_OFFLINE - do not fail req if revocation check cannot be done, i.e. error from revocation is ignored and it proceeds with the connection

I looks like revocation check does not use whatever settings we configured on the client and instead goes with defaults (can be changed on os level i guess). What this means in practice is with revocation check on, it will try to connect to revocation server, which is not reachable, wait 15s (default time out) and then proceed to next step because of  SCH_CRED_IGNORE_REVOCATION_OFFLINE.

So effectively the current behavior is best effort revocation check. We should also probably use SCH_CRED_IGNORE_NO_REVOCATION_CHECK by default. 

Should option be something explicit like enum Revocation_Checking_mode = {best_effort, no_check}?

tests/tls_handler_test.c
sbSteveK: I will remove the ifdef for disabled. enabled will probably only appropriately fail on windows and maybe linux.

PR #808, opened 2026-06-23
badssl.com starts to close sockets now.

TingDaoK:
*Issue #, if available:*
- lot of badssl.com test start to fail with socket closed without finishing TLS

*Description of changes:*
- allow to socket closed as flacky as well
- I have try a local docker image of [badssl](https://github.com/chromium/badssl.com) so that we don't need to hit their endpoint, but problems are:
  - https://github.com/chromium/badssl.com/blob/master/certs/Makefile#L11 they don't generate real expired cert
  - they don't provide all endpoints we are using eg: `rc4.` and `rc4-md5.`
  - the host has openssl 3.0 that doesn't support tls10 anymore. 
  - Our CI setup is running from another docker that doesn't have everything needed `openssl` or `docker`.

There are lot of workaround for the issues I listed above, but overall, I lost my interest. https://github.com/awslabs/aws-c-io/tree/local-badssl here is the branch


tests/tls_handler_test.c
```diff
@@ -37,7 +37,8 @@
 
 bool s_is_badssl_being_flaky(const struct aws_string *host_name, int error_code) {
     if (strstr(aws_string_c_str(host_name), "badssl.com") != NULL) {
-        if (error_code == AWS_IO_SOCKET_TIMEOUT || error_code == AWS_IO_TLS_NEGOTIATION_TIMEOUT) {
+        if (error_code == AWS_IO_SOCKET_TIMEOUT || error_code == AWS_IO_TLS_NEGOTIATION_TIMEOUT ||
+            error_code == AWS_IO_SOCKET_CLOSED) {
             fprintf(
                 AWS_TESTING_REPORT_FD,
                 "Warning: badssl.com is timing out right now. Maybe run the test again later?\n");
```
sbSteveK: Trivial but could we update this log to output the specific error we are encountering since it can be three different ones including a new one that isn't a timeout.

azkrishpy (approved):

PR #811, opened 2026-07-07
Unsetting USE_S2N disables s2n on macOS

sfod:
`USE_S2N` CMake option controls whether to enable s2n-tls on macOS.

On Linux, BSD, and Android, s2n-tls is still always forced as only available TLS implementation.


TingDaoK (approved):
side note:

https://github.com/awslabs/aws-c-io#build-time-options
can we double check those are correct?

PR #814, opened 2026-07-10
Handle optional s2n-tls in the install CMake config

sfod:
*Issue #, if available:*
The install config always run `find_dependency(s2n)` on Unix, forcing consumers to have s2n even when aws-c-io was built without it.

*Description of changes:*
- gate `find_dependency(s2n)` on the actual build decision via `@USE_S2N@` substitution, so s2n is only required when it was really linked in.
- Force `USE_S2N=OFF` when `AWS_USE_SECITEM` is set, to avoid linking an unused s2n dependency.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -121,6 +121,12 @@ elseif (APPLE)
         option(USE_S2N "Use s2n-tls as the TLS implementation." OFF)
     endif()
 
+    # Force USE_S2N OFF to avoid pulling in an unused s2n dependency.
+    if (DEFINED AWS_USE_SECITEM AND USE_S2N)
+        message(WARNING "AWS_USE_SECITEM takes precedence over s2n-tls; forcing USE_S2N=OFF to avoid linking unused s2n.")
+        set(USE_S2N OFF CACHE BOOL "Use s2n-tls as the TLS implementation." FORCE)
+    endif()
+
 elseif (CMAKE_SYSTEM_NAME STREQUAL "FreeBSD" OR CMAKE_SYSTEM_NAME STREQUAL "NetBSD" OR CMAKE_SYSTEM_NAME STREQUAL "OpenBSD")
     file(GLOB AWS_IO_OS_HEADERS
             )
--- a/tests/tls_handler_test.c
+++ b/tests/tls_handler_test.c
@@ -48,14 +48,17 @@ bool s_is_badssl_being_flaky(const struct aws_string *host_name, int error_code)
 
 bool s_is_apple_with_secure_transport(struct aws_allocator *allocator) {
     (void)allocator;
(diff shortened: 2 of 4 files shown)


tests/tls_handler_test.c
```diff
@@ -48,14 +48,17 @@ bool s_is_badssl_being_flaky(const struct aws_string *host_name, int error_code)
 
 bool s_is_apple_with_secure_transport(struct aws_allocator *allocator) {
     (void)allocator;
-#    ifdef __APPLE__
+#    if defined(__APPLE__) && defined(USE_S2N)
```
sbSteveK, line 51: See below comment regarding use of `USE_S2N` here

CMakeLists.txt
sbSteveK, line 208: If this is defined as `PRIVATE`, are we actually using it above in the `tls_handler_test.c` ?

sbSteveK (approved):
Looks good but please investigate the `PRIVATE` issue just to confirm we are doing what we thing we are doing.

CMakeLists.txt
sfod: `tls_handler_test.c` handles s2n on its own:
```cmake
if(USE_S2N)
    target_compile_definitions(${PROJECT_NAME}-tests PRIVATE "-DUSE_S2N")
endif()
```
I checked just in case: the test, which is gated behind USE_S2N, was succesfully executed for the macos-s2n CI jobs.

---

As for the `PRIVATE` vs `PUBLIC`. `PRIVATE` means that `USE_S2N` won't be propagated to the customers of this lib, which I believe is the correct behavior. For instance, we add the event loop defines as PUBLIC:
```cmake
target_compile_definitions(${PROJECT_NAME} PUBLIC "-DAWS_ENABLE_${EVENT_LOOP_DEFINE}")
```
, which leads to the following lines in the install cmake config:
```cmake
set_target_properties(AWS::aws-c-io PROPERTIES
  INTERFACE_COMPILE_DEFINITIONS "AWS_ENABLE_DISPATCH_QUEUE;AWS_ENABLE_KQUEUE"
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
  INTERFACE_LINK_LIBRARIES "AWS::s2n;AWS::aws-c-common;AWS::aws-c-cal;-framework Security -framework Network"
)
```
This means these definitions then are used in the downstream compilation. For example, aws-crt-cpp on Linux currently has this:
```
/usr/bin/clang++-9
-DAWS_ENABLE_EPOLL     <=== HERE
-DCJSON_HIDE_SYMBOLS
...
-c /root/aws-crt-cpp/source/auth/Sigv4Signing.cpp
```

I'd change all or almost all defines to `PRIVATE`, but it's out of scope here.

PR #818, opened 2026-08-05
Fix read-window stall by gating window updates on the scarcest slot

TingDaoK:
## Issue

`aws_channel_slot_increment_read_window()` gated the window-update task on the
**calling slot's** `window_size`. That's the wrong quantity — slots drain at
different rates, so the requesting slot's window says nothing about whether some
other slot is about to starve.

In a `socket → tls → h1` channel the TLS slot drains ciphertext while the HTTP
slot drains plaintext. TLS sizes its own window by estimating record overhead:

```c
likely_records_count = ceil(downstream_size / MAX_RECORD_SIZE);
total_desired_size   = downstream_size + likely_records_count * EST_TLS_RECORD_OVERHEAD;
```

That assumes the peer packs plaintext into full 16 KB records. A peer returning
many small responses on one connection does not, so real per-byte overhead is far
higher than budgeted and the TLS slot drains much faster than the HTTP slot.

HTTP is the only slot requesting window in steady state, so it's the only window
the gate ever tests. It stays far above the threshold while the TLS slot hits
zero — at which point the socket stops reading, no data reaches HTTP, no window
is requested, the gate is never evaluated again, and the connection is wedged
with a full accumulated batch that is never emitted.

## Changes

- Gate on the **scarcest** slot window, not the caller's. 

The left-most slot is excluded from both quantities. 

## Threshold

Raised from `2 * g_aws_channel_max_fragment_size` to `4 *`. One fragment is the
hard floor — below it `s_do_read()` truncates, since it reads
`min(downstream_window, max_rw_size)`. The rest is head-room for a handler that
absorbs a whole read without emitting downstream, and for a caller-supplied
`max_rw_size` larger than one fragment. The exact multiple is not load-bearing:
erring high only makes updates chattier, erring low stalls the read. At `2 *` the
window could reach 1 byte before the task fired, costing a truncated read and a
wasted syscall.

## Follow-up

The per-slot split is the underlying problem, not the gate. Each hop measures
bytes in a different unit, so N slots require N−1 conversions, and each is a
*prediction* of that handler's own expansion factor — unbounded as
plaintext-per-record approaches one byte. This PR removes the stall. A `TODO` in `s_schedule_window_update_task_if_scarce()` records the
direction: manage the window for the channel as a whole so slots cannot disagree.

By submitting this pull request, I confirm that my contribution is made under the
terms of the Apache 2.0 license.


source/channel.c
```diff
@@ -869,8 +909,16 @@ int aws_channel_slot_increment_read_window(struct aws_channel_slot *slot, size_t
         slot->current_window_update_batch_size =
             aws_add_size_saturating(slot->current_window_update_batch_size, window);
 
+        /* Take the minimum of the channel-wide tracked value and this slot's own window.
+         * `min_slot_window_size` is only refreshed when a read message is passed rightward or
+         * when s_window_update_task walks the slots, so a slot that was just created (window_size
+         * of 0, no data through it yet) is not represented there. Without this a slot's initial
+         * window grant would never be emitted. Taking the minimum can only make the task fire
+         * sooner, never later. */
+        size_t effective_window_size = aws_min_size(slot->channel->min_slot_window_size, slot->window_size);
```
azkrishpy: might be a nit but doing it because window update is a scheduled task: if the effective window size can be lower here, we should probably record that in the channel->min_slot_window_size.

source/channel.c
```diff
@@ -694,6 +722,9 @@ int aws_channel_slot_remove(struct aws_channel_slot *slot) {
     }
 
     s_update_channel_slot_message_overheads(slot->channel);
+    /* The removed slot may have been holding the minimum. Reset to SIZE_MAX; the next
+     * s_window_update_task run recomputes it from the surviving slots. */
+    slot->channel->min_slot_window_size = SIZE_MAX;
```
azkrishpy: when do slots get removed? is this okay to do since we might end up disrespecting the user defined window here although only for a shorter period of time.

source/channel.c
```diff
@@ -787,6 +818,8 @@ int aws_channel_slot_send_message(
                 (void *)slot->adj_right,
                 (void *)slot->adj_right->handler);
             slot->adj_right->window_size -= message->message_data.len;
+            slot->channel->min_slot_window_size =
+                aws_min_size(slot->channel->min_slot_window_size, slot->adj_right->window_size);
```
azkrishpy: question: do we do it this way because the first slot is supposed to not have a window?

source/channel.c
TingDaoK: the `min_slot_window_size` only used to check if we need to kick off the batched window update. it doesn't affect the read window size.

source/channel.c
TingDaoK: yeah, the read window on the left most slot is meaningless. the read window is for the upper stream to not provide you more data then expected. 

the left most slot has no upper stream (it's the other side of the network).

source/channel.c
```diff
@@ -239,11 +254,24 @@ struct aws_channel *aws_channel_new(struct aws_allocator *alloc, const struct aw
...
     if (creation_args->enable_read_back_pressure) {
         channel->read_back_pressure_enabled = true;
-        /* we probably only need room for one fragment, but let's avoid potential deadlocks
-         * on things like tls that need extra head-room. */
-        channel->window_update_batch_emit_threshold = g_aws_channel_max_fragment_size * 2;
+        /* Sized so the socket handler can always issue a full-fragment read while a window update
+         * is pending, rather than a truncated one:
+         *
+         *   1 fragment  - below this s_do_read() truncates, since it reads
+         *                 min(downstream_window, max_rw_size) and max_rw_size is one fragment.
```
azkrishpy: This might be just me but I am finding it a little hard to understand the increase of threshold. Not a blocker for the PR since the 2*fragment size was cutting it too close anyway. 
But the comment here says `max_rw_size` is one fragment. but the comment below on line 272 says `max_rw_size` is caller-supplied and need not equal to the fragment_size.

source/channel.c
azkrishpy: would this delay the batched_window_update_task then? also the comment above suggests the window_update_task would compute the minimum? aren't we computing this window_size for scheduling the update task? (Guessing send_message is what computes the min_window of the channel as messages go from left to right)

azkrishpy (approved):
Nits only

source/channel.c
TingDaoK: the `channel->min_slot_window_size` should be a thread local variable. this function can be invoked from out side the channel thread.

TingDaoK: i don't know. my mental model for this value is a helper, that is not necessary to be 100% correct. Since all it does it to batch the updates, doesn't affect the real windowing logic. 

Maybe, we should just iterate the slots here to get the number instead.

DmitriyMusatkin (approved):

PR #821, opened 2026-08-10
Clean up docs

azkrishpy:
*Issue #, if available:*

*Description of changes:*


include/aws/io/private/pki_utils.h
```diff
@@ -63,7 +63,7 @@ int aws_import_pkcs12_to_identity(
     CFArrayRef *identity);
 
 /**
- * Loads PRM armored PKCS#7 certificates into certs
```
azkrishpy, line 66: this was correct

azkrishpy: Reverted — restored 'PRM' as it was in main.

TingDaoK (approved):

source/darwin/darwin_pki_utils.c
```diff
@@ -931,7 +931,7 @@ int aws_import_trusted_certificates(
             CFRelease(certificate_ref);
             CFRelease(cert_blob);
         } else {
-            err = AWS_OP_SUCCESS;
+            err = aws_raise_error(AWS_ERROR_OOM);
```
sfod, line 934: This is unfortunately a sliiightly breaking change.
Secure transport [passes a CA file](https://github.com/awslabs/aws-c-io/blob/v0.27.6/source/darwin/secure_transport_tls_channel_handler.c#L1228-L1237), which can contain multiple CAs. Currently, some users **might** importing CA file with a few invalid records and still get success. This fix will break them.

I think we should make this fix (or maybe something more sophisticated). But I suggest we implement it in a separate PR describing all required details, calling out risks, etc. Not in a PR with a bunch of comment fixes.

sfod: Wait, actually it's not a breaking change - validation happens in `aws_pem_objects_init_from_file_contents`, which returns `AWS_OP_ERR` even if a single CA is invalid. The only scenario when we can reach this line is indeed OOM. So, I guess it's fine having it in this PR.

sfod, line 934: This should be followed by `break;` since this error is final and will be returned.

sfod (approved):

Issue #822, opened 2026-08-11
Add support for keychain-less mode to avoid storing the certificate and private key in the macOS keychain.


vm-cyber:
### Describe the feature

Currently the library adds the certificate and private key to the macOS keychain to obtain a SecIdentityRef.

### Use Case

This is incompatible with our product's security requirements, which do not allow storing the certificate and private key in the macOS keychain.

### Proposed Solution

Use SecIdentityCreate function to create a new identity object from the provided certificate and its associated private key.

https://developer.apple.com/documentation/security/secidentitycreate(_:_:_:)?language=objc

Proposed Solution:
[avoid-using-the-keychain.patch](https://github.com/user-attachments/files/30941252/avoid-using-the-keychain.patch)

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

azkrishpy:
Hi, could you highlight how aws-c-io is being consumed within your product? This would help us determine the feasibility of this request better. Thanks!

vm-cyber:
Hi, In our product, we use the AWS IoT Device SDK for C++ v2 to establish a secure MQTT connection using mutual TLS (mTLS).Due to security requirements, we cannot store the certificate and private key in the macOS System keychain.

rakshil14-2:
Hi,
We've added an opt-in `AWS_SECITEM_NO_KEYCHAIN` build option for Apple platforms, which lets you use SecItem/Secure Transport without storing items in the system Keychain. It will be available in the next release.
Thanks!

vm-cyber:
Thank you!
One note, SecIdentityCreate isn't a private (SPI) function. It is a public API. https://developer.apple.com/documentation/security/secidentitycreate(_:_:_:)?language=objc

rakshil14-2:
Thanks. 
You're right, it's public now. Apple made it a public documented API in the macOS 26 SDK (October 2025). In earlier SDKs it isn't declared in the Security.framework headers.

PR #826, opened 2026-08-18
feat: keychain-less macOS identity mode

azkrishpy:
*Issue: * #822

*Description of changes:*
Adds an opt-in `AWS_SECITEM_NO_KEYCHAIN` build option for Apple platforms. When
enabled, the mTLS SecIdentityRef is built in memory directly from the provided
certificate and private key via SecIdentityCreate (wrapped with
sec_identity_create), instead of adding them to the shared macOS Keychain and
reading the identity back out. This suits products whose security requirements
forbid persisting the certificate and private key in the system Keychain.

The existing keychain-based path is preserved under #else, so the option is
off by default and existing consumers are unaffected.

CMakeLists.txt: define -DAWS_SECITEM_NO_KEYCHAINwhen the option is set.
source/darwin/darwin_pki_utils.c: #ifdef AWS_SECITEM_NO_KEYCHAIN branch in
aws_secitem_import_cert_and_key.


.github/workflows/ci.yml
```diff
@@ -345,6 +345,23 @@ jobs:
...
+    steps:
+    - uses: aws-actions/configure-aws-credentials@v4
+      with:
+        role-to-assume: ${{ env.CRT_CI_ROLE }}
+        aws-region: ${{ env.AWS_DEFAULT_REGION }}
+    - name: Build ${{ env.PACKAGE_NAME }} + consumers
+      run: |
+        python3 -c "from urllib.request import urlretrieve; urlretrieve('${{ env.BUILDER_HOST }}/${{ env.BUILDER_SOURCE }}/${{ env.BUILDER_VERSION }}/builder.pyz?run=${{ env.RUN }}', 'builder')"
+        chmod a+x builder
+        ./builder build -p ${{ env.PACKAGE_NAME }} --cmake-extra=-DAWS_USE_SECITEM=ON --cmake-extra=-DAWS_DONOT_USE_KEYCHAIN=ON --cmake-extra=-DAWS_USE_APPLE_NETWORK_FRAMEWORK=ON --cmake-extra=-DENABLE_SANITIZERS=ON --cmake-extra=-DSANITIZERS="${{ matrix.sanitizers }}"
```
sfod: From what I see, mTLS tests aren't executed for SecItem, so these changes are not actually covered by tests.
I believe we added at least one suitable test when enabled Network Framework for macOS (see [this PR](https://github.com/awslabs/aws-c-io/pull/756/changes#diff-4461c617ceae0c7e0206622aacdf555aedd62eb129c2efb74c84fa1567bcbe0d)) but then they were reverted along with the implementation. You can take a look at that PR, though probably it's easier to generate a new test.

CMakeLists.txt
```diff
@@ -201,6 +201,10 @@ if (AWS_USE_SECITEM)
     target_compile_definitions(${PROJECT_NAME} PUBLIC "-DAWS_USE_SECITEM")
 endif()
 
+if (AWS_DONOT_USE_KEYCHAIN)
```
sfod: This flag works only when SecItem is used. Probably it should be verified here to avoid confusion in the future.

sfod: And for macOS only.

source/darwin/darwin_pki_utils.c
```diff
@@ -363,6 +363,10 @@ void aws_cf_release(CFTypeRef obj) {
     }
 }
 
+#ifdef AWS_DONOT_USE_KEYCHAIN
```
sfod: We should describe this new flag in README: https://github.com/awslabs/aws-c-io#build-time-options

sfod (approved):

PR #827, opened 2026-08-19
Handle invalid PEM objects

sfod:
Improve robustness against invalid (i.e. empty or containing garbage) PEM objects. It mostly touched certificates except Apple platforms, which improves handling of empty private keys.


source/darwin/darwin_pki_utils.c
bretambrose: Really hate that we're setting an output variable and escaping (leaking?) a resource allocation (temp_cert_array) even on failure.  Anyway we could change that?  Might need to change callers too if they currently clean up on failure.

source/s2n/s2n_tls_channel_handler.c
```diff
@@ -1787,7 +1787,7 @@ static struct aws_tls_ctx *s_tls_ctx_new(
 
                 if (set_ca_result) {
                     s_log_and_raise_s2n_errno("ctx: configuration error");
-                    AWS_LOGF_ERROR(AWS_LS_IO_TLS, "Failed to set ca_file %s\n", (const char *)options->ca_file.buffer);
+                    AWS_LOGF_ERROR(AWS_LS_IO_TLS, "Failed to set ca_file " PRInSTR, AWS_BYTE_BUF_PRI(options->ca_file));
```
bretambrose: Do we know that this is printable at this point?

sfod: Good callout. This field is **supposed** to contain PEM certs, i.e. printable symbols only. However, this is raw user-supplied data, with no sanitization, so in practice there can be anything.
We could sanitize the content. But if a failure is caused by non-printable symbols, this line will print a valid PEM saying it's invalid, which will be confusing. So, it's safer to remove CA content from logs.

bretambrose: Maybe we could add some kind of "printable" helper function and ternary switch between the raw data and "INVALID" or similar in the log invocation?

source/darwin/darwin_pki_utils.c
sfod: It's indeed leaking on an seemingly-valid-but-not-really cert. I've added a test for that. This wouldn't happen if we had only one path for cleaning `secure_transport_ctx`. I'll try to implement it without breaking anything.

source/s2n/s2n_tls_channel_handler.c
sfod: It can be implemented, though is it worth it? Since `ca_file` is passed from user side and we don't transform it, they can examine data on their side. Without us specifying where exactly parsing failed, it'll simply look too verbose without bringing much value imo:
```
[DEBUG] [2026-08-25T19:14:14Z] [00000001f603e180] [tls-handler] - Set security policy to AWS-CRT-SDK-TLSv1.0-2025-PQ (minimum_tls_version: 128; cipher_pref: 0)
[ERROR] [2026-08-25T19:14:14Z] [00000001f603e180] [tls-handler] - Failed to set ca_file -----BEGIN CERTIFICATE-----
MIICeDCCAeGgAwIBAgIJAObttnPKQhVlMA0GCSqGSIb3DQEBDgUAMF8xCzAJBgNV
MIICeDCCAeGgAwIBAgIJAObttnPKQhVlMA0GCSqGSIb3DQEBDgUAMF8xCzAJBgNV
MIICeDCCAeGgAwIBAgIJAObttnPKQhVlMA0GCSqGSIb3DQEBDgUAMF8xCzAJBgNV
MIICeDCCAeGgAwIBAgIJAObttnPKQhVlMA0GCSqGSIb3DQEBDgUAMF8xCzAJBgNV
MIICeDCCAeGgAwIBAgIJAObttnPKQhVlMA0GCSqGSIb3DQEBDgUAMF8xCzAJBgNV
MIICeDCCAeGgAwIBAgIJAObttnPKQhVlMA0GCSqGSIb3DQEBDgUAMF8xCzAJBgNV
MIICeDCCAeGgAwIBAgIJAObttnPKQhVlMA0GCSqGSIb3DQEBDgUAMF8xCzAJBgNV
MIICeDCCAeGgAwIBAgIJAObttnPKQhVlMA0GCSqGSIb3DQEBDgUAMF8xCzAJBgNV
MIICeDCCAeGgAwIBAgIJAObttnPKQhVlMA0GCSqGSIb3DQEBDgUAMF8xCzAJBgNV
-----END CERTIFICATE-----
-----BEGIN CERTIFICATE-----
BAYTAlVTMQswCQYDVQQIDAJXQTEQMA4GA1UEBwwHU2VhdHRsZTEPMA0GA1UECgwG
BAYTAlVTMQswCQYDVQQIDAJXQTEQMA4GA1UEBwwHU2VhdHRsZTEPMA0GA1UECgwG
BAYTAlVTMQswCQYDVQQIDAJXQTEQMA4GA1UEBwwHU2VhdHRsZTEPMA0GA1UECgwG
BAYTAlVTMQswCQYDVQQIDAJXQTEQMA4GA1UEBwwHU2VhdHRsZTEPMA0GA1UECgwG
BAYTAlVTMQswCQYDVQQIDAJXQTEQMA4GA1UEBwwHU2VhdHRsZTEPMA0GA1UECgwG
BAYTAlVTMQswCQYDVQQIDAJXQTEQMA4GA1UEBwwHU2VhdHRsZTEPMA0GA1UECgwG
-----END CERTIFICATE-----
... etc

[DEBUG] [2026-08-25T19:14:14Z] [00000001f603e180] [mqtt] - ... 
```

source/darwin/darwin_pki_utils.c
sfod: User creds also leaking when CA is invalid. I routed the failure path to `s_aws_secure_transport_ctx_destroy`, so I believe these leaks and maybe other undiscovered should be fixed.

PR #829, opened 2026-08-26
chore: test for Keychainless macOS identity

rakshil14-2:
*Issue #, if available:*

*Description of changes:*

  - Added an mTLS test case for the `AWS_DONOT_USE_KEYCHAIN` .
  - Uses `mtls_device_pkcs1.key` (PKCS1 re-encoding of the same key) because SecItem cannot import PKCS8 private keys; the certificate is unchanged.
  - Targets the `local TLS 1.2` server because Apple Secure Transport does not support TLS 1.3.
  - Registered in CMake under Darwin + AWS_DONOT_USE_KEYCHAIN so it only runs on Apple with SecItem.
  - `mtls_device_pkcs1.key` is generated by tests/resources/generateCerts.sh.


tests/tls_handler_test.c
```diff
@@ -1723,6 +1731,39 @@ AWS_TEST_CASE(
     tls_client_channel_negotiation_failure_mtls_untrusted_server_from_keychain,
     s_tls_client_channel_negotiation_failure_mtls_untrusted_server_from_keychain_fn)
 
+/* Keychain-less SecItem mTLS test. When aws-c-io is built with AWS_DONOT_USE_KEYCHAIN.*/
+#    if defined(AWS_DONOT_USE_KEYCHAIN)
+static void s_mtls_disable_verify_peer(struct aws_tls_ctx_options *options) {
```
sfod: There is a `s_disable_verify_peer` function defined below doing the same. So no need in adding this one.

tests/CMakeLists.txt
```diff
@@ -128,6 +128,10 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
     add_net_test_case(tls_client_channel_negotiation_success_mtls_from_keychain)
     add_net_test_case(tls_client_channel_negotiation_failure_mtls_untrusted_server_from_keychain)
     add_net_test_case(tls_client_channel_negotiation_failure_tls13_to_tls12_server)
+
+    if(AWS_DONOT_USE_KEYCHAIN)
+        add_net_test_case(tls_client_channel_negotiation_success_mtls_secitem_no_keychain)
```
sfod, line 136: Add a short note here and in `s_tls_client_channel_negotiation_success_mtls_secitem_no_keychain_fn` explaining why we're not running this test with SecItem and keychain.
You can refer to an [explanation here](https://github.com/awslabs/aws-c-io/blob/v1.0.0/tests/CMakeLists.txt#L327-L329).

sfod (approved):
fix&ship

PR #834, opened 2026-08-28
Socks5 Implementation Part Two - Protocol

bretambrose:
Protocol and channel handler implementation. Tests only give partial coverage; broader coverage comes in part 3 when we integrate with channel construction and add end-to-end tests with mock servers.


source/socks5/socks5.c
```diff
@@ -54,3 +56,528 @@ int aws_socks5_proxy_negotiation_strategy_instance_get_auth_methods(
...
+static struct aws_l4_proxy_channel_handler *s_aws_l4_proxy_channel_handler_new_socks5(
+    struct aws_l4_proxy_config *config,
+    struct aws_l4_proxy_channel_handler_options *options) {
+    struct aws_allocator *allocator = config->allocator;
+    struct aws_socks5_channel_handler *socks5_handler =
+        aws_mem_calloc(allocator, 1, sizeof(struct aws_socks5_channel_handler));
+    struct aws_socks5_proxy_config *socks5_config = config->impl;
+
+    socks5_handler->allocator = allocator;
+    socks5_handler->negotiation_impl = aws_socks5_proxy_impl_new(allocator, socks5_config, options);
```
sfod: I don't see any checks for `socks5_handler->negotiation_impl` not being NULL. If `aws_socks5_proxy_impl_new` fails, the error may stay unnoticed until `socks5_handler->negotiation_impl` is accessed also without checks later.

source/l4_proxy.c
```diff
@@ -0,0 +1,385 @@
...
+
+    // the handler tells us how many bytes they used, use that for window updates
+    size_t pre_consumed_bytes = fragment_cursor.len;
+    int negotiation_result = s_drive_negotiation_l4_proxy(handler, &context);
+    size_t consumed_bytes = pre_consumed_bytes - fragment_cursor.len;
+    aws_channel_slot_increment_read_window(handler->channel_handler.slot, consumed_bytes);
+
+    if (message) {
+        aws_mem_release(message->allocator, message);
+    }
```
sfod, line 141: For the server-speak-first protocols, `message` can contain server data, which is freed here.
The intended use (HTTP, TLS) shouldn't be affected, but still worth revisiting this part.

include/aws/io/l4_proxy.h
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Creates a new channel handler that will negotiate a tunnel based on the proxy config
+ *
+ * @param config proxy config to use for tunnel negotiation
+ * @param options additional channel handle options
+ * @return a new channel handler for proxy negotiation
+ */
+AWS_IO_API struct aws_l4_proxy_channel_handler *aws_l4_proxy_config_new_channel_handler(
+    struct aws_l4_proxy_config *config,
+    struct aws_l4_proxy_channel_handler_options *options);
```
sfod: `aws_l4_proxy_config_new_channel_handler` is public but it accepts `struct aws_l4_proxy_channel_handler_options`, which is private. I don't see any public functions that could create an instance of `struct aws_l4_proxy_channel_handler_options`, so `aws_l4_proxy_config_new_channel_handler` is effectively private. Can it be moved to a private header?

And the same applies to `aws_l4_proxy_config_get_proxy_address` below.

source/l4_proxy.c
bretambrose: Supporting that greatly complicates windowing.  The implementation originally did support it by queuing data in the handler but I ended up ripping it out because of the inability to test and the fact that our other handlers don't support it either.

include/aws/io/l4_proxy.h
bretambrose: They're used in tests; unsure if I can move and not break the .so build.  Will try and see what happens.

source/l4_proxy.c
sfod: Maybe worth adding a comment outlining limitations for future us? Though probably not here but in a more visible place. I'll leave it up to you.

bretambrose: Could detect when there's additional data and just shut the channel down with logging maybe

include/aws/io/l4_proxy.h
bretambrose: Seems to be fine

source/socks5/socks5.c
```diff
@@ -54,3 +56,534 @@ int aws_socks5_proxy_negotiation_strategy_instance_get_auth_methods(
...
+
+static void s_build_connect_request(struct aws_socks5_proxy_impl *impl) {
+    aws_byte_buf_reset(&impl->write_buffer, false);
+
+    // version (1 byte) + command (1 byte) + reserved (1 byte) + address type (1 byte) +
+    //   address length (1 byte) + address (variable length) + port (2 bytes)
+    size_t requiredBytes = 7 + impl->remote_host.len;
+    aws_byte_buf_reserve(&impl->write_buffer, requiredBytes);
+
+    aws_byte_buf_write_u8(&impl->write_buffer, SOCKS5_VERSION_VALUE);                // version byte
```
sfod, line 334: Is `SOCKS5_VERSION_BYTE` the same as `SOCKS5_VERSION_VALUE`?

sfod (approved):

source/socks5/socks5.c
bretambrose: Yes

PR #836, opened 2026-09-01
feat: Opt in flags for apple network framework

rakshil14-2:
*Issue #, if available:*

*Description of changes:*
Adds three opt-in socket options for the Apple Network Framework backend.
- **Connect-by-name:** (_Default False_) hands the hostname to Network Framework via `nw_endpoint_create_host` instead of pre-resolving, enabling VPN On Demand, content filters, and HTTP CONNECT proxies (Apple TN3151). It is valid for for tCP over IPv4/IPv6 only; other type/domain combos rejected with `AWS_IO_SOCKET_INVALID_OPTIONS`.
    - New `aws_socket_endpoint.host_name` field:  carries the full FQDN unbounded by `AWS_ADDRESS_MAX_LEN`.
- **prefer_no_proxy:** ignores system-configured proxies via `nw_parameters_set_prefer_no_proxy`
- **privacy_context:** opaque void* passthrough to `nw_parameters_set_privacy_context`


source/darwin/nw_socket.c
```diff
@@ -1841,74 +1852,113 @@ static int s_socket_connect_fn(struct aws_socket *socket, struct aws_socket_conn
...
-            (int)remote_endpoint->port);
-        aws_raise_error(AWS_IO_SOCKET_INVALID_ADDRESS);
-        goto error;
-    }
+    nw_endpoint_t endpoint = NULL;
 
-    struct socket_address address;
-    AWS_ZERO_STRUCT(address);
-    int pton_err = 1;
+    if (socket->options.connect_by_name) {
```
bretambrose, line 1857: Why isn't this caught and resolved earlier?

source/channel_bootstrap.c
```diff
@@ -560,6 +560,20 @@ static bool s_aws_socket_domain_uses_dns(enum aws_socket_domain domain) {
     return domain == AWS_SOCKET_IPV4 || domain == AWS_SOCKET_IPV6;
 }
 
+/* connect_by_name is only used on the Apple Network Framework backend (nw_endpoint_create_host). On every
+ * other backend there is no OS-level connect-by-name path, so treat the flag as false to keep it a true no-op
+ * rather than routing a hostname into a resolver-bypass that later fails address parsing. */
+static bool s_connect_by_name_is_supported(const struct aws_socket_options *socket_options) {
```
bretambrose: is_enabled

bretambrose (changes requested):
As a quick reaction, I'm really uncomfortable with the direction of this PR and the precedent it would be setting.  Let's try and figure out if there are some alternative approaches.

General comments:
1. I don't like that we're putting platform-specific fields in a common, copy-by-value structure that is bound out to every CRT.  We're making things worse (complexity, usability) and establishing a precedent that will also only get worse over time.
2. I don't like that there's an opaque pointer to an OS handle.  How would this be bound at the CRT layer?
3. I don't like that there's a redundant name field that's also just a pointer with uncertain lifetime (client connection args gets destroyed after connection resolution but a socket endpoint might persist longer if its copied, for example)
4. It's not at all clear that the pointers are safe and it creates a nasty surprise potential when writing bindings that might be forced to keep things alive for an arbitrary amount of time.

Maybe there's some kind of "extension"/"platform extension" framework that could be developed.  I hated what happened with tls options and platform-specific fields for similar reasons.

Issue #838, opened 2026-09-02
epoll event loop does not run due tasks after device suspend because the `epoll_wait` timeout is measured on `CLOCK_MONOTONIC`


nrutherford:
### Describe the bug

The Linux epoll event loop schedules tasks using `aws_high_res_clock_get_ticks`, which since aws-c-common#1189 is `CLOCK_BOOTTIME` and advances through suspend. `epoll_event_loop.c `then turns the next due time into a relative millisecond timeout for `epoll_wait`, and the kernel measures that timeout on `CLOCK_MONOTONIC`, which does not advance while the device is suspended. 

While suspended the wait makes no progress; each wake advances it only by the time the CPU stays awake. A task scheduled 40 s ahead therefore runs after 40 s of accumulated awake time, or when a file descriptor in the set becomes readable, whichever comes first. On a device that suspends between short wakes the MQTT 3.1.1 keep-alive PINGREQ is never sent in time; the broker's own TCP close is what finally wakes the loop.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

A task due at time T on the loop's clock runs at the first opportunity the CPU is awake after T, regardless of how long the device was suspended, so that with a 40 s keep-alive a PINGREQ leaves within the broker's 1.5x window whenever the device wakes at least once between 40 s and 60 s after the last packet.

### Current Behavior

Android 11 device (arm64), mobile data, screen off, on battery, app holding a persistent MQTT 3.1.1 connection to AWS IoT Core through aws-crt-android, no other traffic:

- keep-alive 40 s: broker disconnects with `MQTT_KEEP_ALIVE_TIMEOUT` 60 s after the last packet the client sent; the client reconnects 2 to 4 s later; repeats every 62 to 64 s.
- keep-alive 300 s: same, every 454 to 455 s.
- In both runs the device's battery-stats history shows the CPU waking every 10 to 40 s for 1 to 3 s at a time (modem, RTC, alarms, jobs). No PINGREQ was sent on any of those wakes.
- On the same device on Wi-Fi, where the firmware did not suspend the CPU, the connection stayed up indefinitely with a PINGREQ at every keep-alive interval.
- The broker's `GetConnection` confirmed the negotiated `keepAliveDuration` in each run.

No client-side error is logged before the disconnect; the first event is the connection-interrupted callback when the broker closes the socket.

### Reproduction Steps

[...]

PR #840, opened 2026-09-04
Fix epoll event loop delays tasks after device suspend 

sfod:
*Issue #, if available:* https://github.com/awslabs/aws-c-io/issues/838

`epoll_wait`'s timeout is measured on `CLOCK_MONOTONIC`, which stops advancing while the device is suspended. A task scheduled for a real-time deadline therefore isn't run until that many awake seconds pass (i.e. the task fires roughly the suspend duration late).

*Description of changes:*

Add a timerfd with `CLOCK_BOOTTIME` (which does count suspend time) for the next deadline and register it with epoll, so `epoll_wait` returns on the first wake at/after the deadline.

I verified these changes manually on a bare-metal Linux machine using `rtcwake` to trigger S3 suspend.  An unfixed build delivered a 60s task ~30s late after a 20s suspend; the fixed build fired on time. Since it depends on actual suspend on physical hardware, implementing tests is impractical.


README.md
```diff
@@ -793,6 +793,7 @@ Option | Platform | Description
...
 `BYO_CRYPTO` | Linux/Non-Apple Unix | Disables the built-in TLS implementation and crypto linkage. Your application must provide its own `aws_tls_ctx` and `aws_channel_handler` implementations.                                                                                                                                                        | OFF
+`BUILD_RELOCATABLE_BINARIES` | Linux | Builds a binary that runs on target kernels older than the build host, for prebuild pipelines that compile on an old toolchain (defines `COMPAT_MODE`). The Linux event loop then avoids newer facilities those old kernels lack: cross-thread notification falls back from `eventfd` to a pipe, and scheduled-task wakeups fall back from a `CLOCK_BOOTTIME` timerfd (which keeps tasks on time across device suspend) to the `CLOCK_MONOTONIC` `epoll_wait` timeout. Enable only when the build host's kernel is newer than the oldest kernel you must support. | OFF
```
bretambrose: Is there CI for this?

source/linux/epoll_event_loop.c
```diff
@@ -278,6 +321,11 @@ static void s_complete_destroy(struct aws_event_loop *event_loop) {
     close(epoll_loop->write_task_handle.data.fd);
 #endif
 
+    if (epoll_loop->timer_handle.data.fd >= 0) {
```
bretambrose: Feels like we need an `aws_io_handle_is_valid_resource` API.

source/linux/epoll_event_loop.c
```diff
@@ -18,8 +18,30 @@
 
 #include <errno.h>
 #include <limits.h>
+#include <time.h>
 #include <unistd.h>
 
+#if defined(COMPAT_MODE) || !defined(CLOCK_BOOTTIME)
```
bretambrose: Anyway we could move these to cmake-level?  

If not, I feel like we want some feature header that has stuff like this so the values are accessible outside here.

source/linux/epoll_event_loop.c
```diff
@@ -541,6 +589,48 @@ static void s_on_tasks_to_schedule(
...
+#if AWS_USE_BOOTTIME_TIMERFD
+    if (epoll_loop->timer_handle.data.fd < 0) {
+        return;
+    }
+
+    /* NOTE: it_value of 0 disarms the timer. */
+    struct itimerspec timer_value = {
+        .it_value =
+            {
+                .tv_sec = (time_t)(timeout_ns / 1000000000ULL),
```
bretambrose: clearer to use aws_timestamp_convert

